Issue #13, opened 2020-02-12
Inverted port assignments


nickperry:
In https://github.com/plunder-app/kube-vip/blob/master/kubernetes-control-plane.md the kube-vip and kube-apiserver port assignments are inverted between the kube-vip config and the kube init params. I.e. you've swapped who is 6443 and who is 6444.

nickperry:
And in the kubeadm join

thebsdbox:
This is now fixed

PR #14, opened 2020-02-12
This fixes the init command

thebsdbox:
The `init` example uses the wrong arguments for the VIP and the API-servers, when compared to the configuration of `kube-vip`. This commit fixes this issue.

/cc @nickperry


nickperry:
Sprr, needs it on the join too

thebsdbox:
Fixes #13

PR #17, opened 2020-02-25
Kube-vip now works as an app LB

thebsdbox:
Huge changes:

`kube-vip` can now run in the cluster on worker nodes and be used to load balancer services to the outside world.

This is used through the `kube-vip service` sub command, additionally:

- `--configMap` points to the configMap used by the cloud-provider
- `--arp` enables the use of arp broadcasts

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@ TARGET := kube-vip
 .DEFAULT_GOAL: $(TARGET)
 
 # These will be provided to the target
-VERSION := 0.1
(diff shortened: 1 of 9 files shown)


aleccunningham:
Any example documentation of this in practice?

thebsdbox:
@aleccunningham not yet, but soon. It’s still a bit unstable, but it generally works.

thebsdbox:
@aleccunningham https://plndr.io/kube-vip

PR #24, opened 2020-03-04
InCluster VIP now uses leader election

thebsdbox:
All auto-detect per namespace...

`sips whiskey`

This allows multiple `Kube-vip` in a namespace, with a leader holding the VIP and the load-balancing. In the event of maintenance or failure another will take leadership and host the VIP and load-balancing.

**TODO**: docs

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@ TARGET := kube-vip
 .DEFAULT_GOAL: $(TARGET)
 
 # These will be provided to the target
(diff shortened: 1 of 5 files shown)


thebsdbox:
https://github.com/kubernetes/kubernetes/pull/88192

^ Currently failover is broken in `leadership` election.. so dropping the leadership election checks yields decent enough failover for now.

PR #37, opened 2020-04-22
Update kubernetes control plane README for version 0.1.5

megian:
The binary /kube-vip must not be defined in version 0.1.5.


thebsdbox:
Ah thanks for this PR, I'm terrible with keeping documentation up to date !

PR #39, opened 2020-04-22
Update sample manifest

megian:
Args instead of command, because the kube-vip binary is already executed
by the the image.

Use docker.io in image path, because docker.io is not always the
default registry.

Removing Env as it is currently empty and causes the sample not to work.

Add volume mount for the config file.

### DIFF
--- a/cmd/kube-vip-config.go
+++ b/cmd/kube-vip-config.go
@@ -104,7 +104,7 @@ var kubeVipSampleManifest = &cobra.Command{
 				Containers: []appv1.Container{
 					{
 						Name:  "kube-vip",
-						Image: fmt.Sprintf("plndr/kube-vip:%s", Release.Version),
+						Image: fmt.Sprintf("docker.io/plndr/kube-vip:%s", Release.Version),
 						SecurityContext: &appv1.SecurityContext{
 							Capabilities: &appv1.Capabilities{
 								Add: []appv1.Capability{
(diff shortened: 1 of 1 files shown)


thebsdbox:
Hrm.. I'll need to think about this.. I was working on something undocumented to make it simpler to deploy by making a separate configuration file un-required.

Use `<kube-vip> kubeadm init`:

```
sudo docker run --network host --rm plndr/kube-vip:0.1.4 kubeadm init --interface ens192 --vip 192.168.0.81 | sudo tee /etc/kubernetes/manifests/vip.yaml
```

Environment variables in manifest:

```
$ cat /etc/kubernetes/manifests/vip.yaml
apiVersion: v1
kind: Pod
metadata:
  creationTimestamp: null
  name: kube-vip
  namespace: kube-system
spec:
  containers:
  - args:
    - start
    env:
    - name: vip_arp
      value: "true"
    - name: vip_interface
      value: ens192
    - name: vip_address
      value: 192.168.0.81
    - name: vip_startleader
      value: "true"
    - name: vip_addpeerstolb
      value: "true"
    - name: vip_localpeer
      value: controlPlane01:192.168.0.70:10000
    - name: lb_backendport
      value: "6443"
    - name: lb_name
      value: Kubeadm Load Balancer
    - name: lb_type
      value: tcp
    - name: lb_bindtovip
      value: "true"
    image: plndr/kube-vip:0.1.4
    imagePullPolicy: Always
    name: kube-vip
    resources: {}
    securityContext:
      capabilities:
        add:
        - NET_ADMIN
        - SYS_TIME
  hostNetwork: true
status: {}
```

megian:
For simple configurations no doubt, but the config file might be a valid solution, if multiple ports shall be supported for i.e. Cluster API related things like cloud-init, images required for the bootstrap before the Kubernetes API is up. But agree, it's questionable, if such special configurations needs have a built-in sample.

Anyway the PR could be a quick fix to match the current documentation and increase the user experience until this has been reworked.

thebsdbox:
Makes sense, I'll raise an issue to track the addition of `ENV` in the manifest separately. The plan is all around making it more simple for Cluster-API. Any input you have is totally appreciated :D

PR #46, opened 2020-06-19
Finalise join command

thebsdbox:
This PR adds all of the functionality to use the kube-vip `kubeadm init` and `kubeadm join` commands. It also adds in client-go, to find all controlplane nodes.

### DIFF
--- a/cmd/kube-vip-kubeadm.go
+++ b/cmd/kube-vip-kubeadm.go
@@ -20,6 +20,9 @@ import (
 var initConfig kubevip.Config
 var initLoadBalancer kubevip.LoadBalancer
 
+// Points to a kubernetes configuration file
+var kubeConfigPath string
+
 func init() {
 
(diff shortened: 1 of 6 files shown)


cmd/kube-vip-kubeadm.go
```diff
@@ -84,6 +88,68 @@ var kubeKubeadmJoin = &cobra.Command{
...
+			cmd.Help()
+			log.Fatalln("No address is specified for kube-vip to expose services on")
+		}
+
+		//if  {
+
+		// We will use kubeconfig in order to find all the master nodes
+
+		// use the current context in kubeconfig
+		config, err := clientcmd.BuildConfigFromFlags("", "/etc/kubernetes/admin.conf")
```
thebsdbox, line 110: Check file exists before continuing (Panic is a little confusing in the UX)

PR #61, opened 2020-07-21
update arp config to be inline with help

### LINKED ISSUE #60: fix Gratuitous ARP config default
Binary help file defaults to true for `--arp` flag
```
$ sudo ctr run --rm --net-host docker.io/plndr/kube-vip:0.1.5 init /kube-vip kubeadm init -h
The "init" subcommand will generate the Kubernetes manifest that will be started by kubeadm through the kubeadm init process

Usage:
  kube-vip kubeadm init [flags]

Flags:
  -h, --help   help for init

Global Flags:
      --addPeersToLB        The Virtual IP addres (default true)
      --arp                 Enable Arp for Vip changes (default true)
      --interface string    Name of the interface to bind to
      --lbBackEndPort int   A port that all backends may be using (optional) (default 6444)
      --lbBindToVip         Bind example load balancer to VIP (default true)
      --lbName string       The name of a load balancer instance (default "Kubeadm Load Balancer")
      --lbPort int          Port that load balander will expose on (default 6443)
      --lbType string       Type of load balancer instance (tcp/http) (default "tcp")
      --log uint32          Set the level of logging (default 4)
      --startAsLeader       Start this instance as the cluster leader
      --vip string          The Virtual IP addres
```

However, code config implementation defaults to false.

pickledrick:
fixes #60

### DIFF
--- a/cmd/kube-vip-config.go
+++ b/cmd/kube-vip-config.go
@@ -22,7 +22,7 @@ func init() {
 	kubeVipSampleConfig.Flags().StringVar(&cliConfig.VIP, "vip", "192.168.0.1", "The Virtual IP addres")
 	kubeVipSampleConfig.Flags().BoolVar(&cliConfig.SingleNode, "singleNode", false, "Start this instance as a single node")
 	kubeVipSampleConfig.Flags().BoolVar(&cliConfig.StartAsLeader, "startAsLeader", false, "Start this instance as the cluster leader")
-	kubeVipSampleConfig.Flags().BoolVar(&cliConfig.GratuitousARP, "arp", false, "Use ARP broadcasts to improve VIP re-allocations")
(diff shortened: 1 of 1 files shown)


pickledrick:
Not sure what the intended default is. Happy to change if it's the other way around. The architecture documentation site refers to gratuitous arp as being (optional) this wording makes me feel the intended default is `false`. Let me know and I will update!

thebsdbox:
I think ideally, nearly everything should `default` to `false` and leave the end-user with no assumptions to what the expected behaviour is. i.e. if you want a feature then YOU turn it on, kind of thing. 👍

thebsdbox:
For the `kubeadm` sub-command we will default to enabling arp as it's the mechanism for the control plane VIP fail-over. The changes to to `./kube-vip sample config` above may be deprecated at some point as that config now lives within a Kubernetes static pod manifest.

PR #65, opened 2020-07-24
standardize default interface

### LINKED ISSUE #64: make default interface the same across all components
I found in `kube-vip.go` a default interface of `ens192` as opposed to `eth0` in other configs. Not sure if this is left over from different environment testing.

pickledrick:
fixes #64


thebsdbox:
In hindsight, not sure if we should even specify an example, as every Linux distro seems to have slightly different ways of naming devices.. but `eth0` is the defacto standard so this makes sense. Thanks again 💯

Issue #45, opened 2020-06-18
Advertise VIP through BGP


thebsdbox:
Currently being investigated... shouldn't be too much work (probably)

plunder-app:
This is now added into `0.1.8`

PR #67, opened 2020-08-24
This adds BGP functionality to kube-vip

thebsdbox:
This work is the beginning of #45 

This PR contains the functionality to create a BGP server, connect to a peer and advertise the VIP. When used with `--packet` it will use the token/project to find the peer information through the Packet API in order to create the correct BGP information before advertising the VIP.


pkg/kubevip/config_generator.go
```diff
@@ -53,6 +53,17 @@ const (
 	//vipPacket defines which project within Packet to use
 	vipPacketProject = "vip_packetproject"
 
+	//bgpEnable defines if BGP should be enabled
+	bgpEnable = "bgp_enable"
+	//bgpRouterID defines the routerID for the BGP server
+	bgpRouterID = "bgp_routerID"
```
displague, line 59: nitpic: This doesn't seem to fit the naming pattern (`lower_loweralowerb`)

pkg/cluster/clusterLeader.go
```diff
@@ -181,48 +193,70 @@ func (cluster *Cluster) StartLeaderCluster(c *kubevip.Config, sm *Manager) error
...
-								if ip.Address == c.VIP {
-									log.Infof("Found EIP ->%s ID -> %s\n", ip.Address, ip.ID)
+					// We're using Packet with BGP
+					if c.EnableBGP {
+						log.Debugf("Looking up the BGP configuration from packet")
+						err = packet.BGPLookup(packetClient, c)
+						if err != nil {
+							log.Error(err)
+						}
+					} else {
```
displague, line 205: `// We're using Packet with ...` would compliment the previous comment

displague: This has reached a depth worth extracting it to a new function:
`err := packet.AssignEIP(id, c.PacketProject, c.VIP)`

thebsdbox:
> If BGP is enabled without Packet support, should it return an error?

Nope, packet isn't a requirement. As long as a BGP configuration exists then it will attempt to use it to advertise the vip address. With packet enabled then the API will give us the BGP configuration.

displague:
I see that now, I previously only noticed the BGP block with the Packet block.
https://github.com/plunder-app/kube-vip/pull/67/files#diff-1d55220df04a3715db916fe0b8d788e0R249-R258

Unrelated, would the Packet EIP configuration benefit from a OnStoppedLeading action? Is there ever a time where Unassign (EIP) should be applied by the current leader rather the leader in waiting?

When looking through this PR, I also wondered why do the Packet Project and Secret fields not have yaml tags? Are they not imported/exported with the other config values?

displague:
I'm not in a position to approve this PR (especially for lack of context and a build/run), but ... nothing else jumped out at me, fwiw 👍 .

pkg/kubevip/config_generator.go
thebsdbox: Fixed this, in the lates commit. 👍

pkg/cluster/clusterLeader.go
thebsdbox: Moved the EIP function out to its own function, and simplified a lot of the code with helper functions. 👍

thebsdbox:
> Unrelated, would the Packet EIP configuration benefit from a OnStoppedLeading action? Is there ever a time where Unassign (EIP) should be applied by the current leader rather the leader in waiting?

If anyone else becomes a leader, and especially in the event that an existing leader goes away unexpectedly then the OnLeader action will use the API to remove that existing connection. So regardless of the current state, when Kubernetes elects a new leader we will forcefully assume the EIP and start handling traffic.

plunder-app:
🔥

PR #70, opened 2020-08-28
Leader election fork

thebsdbox:
This does the following:
 - Bumps client-go a version
 - Adds the missing `contexts`
 - Takes leaderElection from Kubernetes upstream and adds a fix


thebsdbox:
Tested with both layer3 and layer2

PR #81, opened 2020-09-29
send arprequest and arpreply alternatively for gratuitous arp

### LINKED ISSUE #80: gratuitous ARP message is not received by other nodes
In one of the CAPV setup with kube-vip, we saw that the gARP message was not received by other nodes. During upgrade when VIP migrated to the new control plane node, 'arp -n' shows other nodes still have the old MAC address for the VIP. 

After digging more it turns out that, both ARPRequest and ARPReply can be used to send gratuitous ARP, and network devices may only support either one of them. 

Also, for the ARPReply that kube-vip is sending, the receiver MAC/IP seems have a mis configuration which does not match what the protocol defines.  

Ref:  
https://tools.ietf.org/html/rfc5944#section-4.6
https://en.wikipedia.org/wiki/Address_Resolution_Protocol. 

Will have a PR to send Request/Reply alternatively, also change the Reply message.

jzhoucliqr:
Both ARPRequest and ARPReply can be used to send gratuitous ARP, and network devices may only support either one of them.  Change to send request and reply alternatively.  

Fix #80

### DIFF
--- a/pkg/vip/arp.go
+++ b/pkg/vip/arp.go
@@ -23,6 +23,8 @@ const (
 
 var (
 	ethernetBroadcast = net.HardwareAddr{0xff, 0xff, 0xff, 0xff, 0xff, 0xff}
+	// arpRequest is used to flip between garp request or garp reply
+	arpRequest = true
 )
 
 func htons(p uint16) uint16 {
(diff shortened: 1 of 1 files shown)


thebsdbox:
Looking at the RFC we probably could just add a tiny toggle at the top of the gratuitous ARP function and keep the payload as below?

```
	m := &arpMessage{}
	m.arpHeader = arpHeader{
		1,           // Ethernet
		0x0800,      // IPv4
		hwLen,       // 48-bit MAC Address
		net.IPv4len, // 32-bit IPv4 Address
		opARPReply,  // ARP Reply
	}

	// https://tools.ietf.org/html/rfc5944#section-4.6
	m.senderHardwareAddress = mac

	// When using an ARP Reply packet, the Target Hardware
	// Address is also set to the link-layer address to which this cache
	// entry should be updated (this field is not used in an ARP Request
	// packet).
	m.targetHardwareAddress = ethernetBroadcast

	// In either case, the ARP Sender Protocol Address and
	// ARP Target Protocol Address are both set to the IP address of the
	// cache entry to be updated,
	m.senderProtocolAddress = ip.To4()
	m.targetProtocolAddress = ip.To4()
```

jzhoucliqr:
Good point. This is cleaner.

ncdc:
@jzhoucliqr are you still working on this? Just wondering how critical it is & if we need to get it in sooner rather than later? Thanks!

PR #82, opened 2020-09-29
fix garp reply msg 

### LINKED ISSUE #80: gratuitous ARP message is not received by other nodes

jzhoucliqr:
Ref:  https://tools.ietf.org/html/rfc5944#section-4.6

A Gratuitous ARP [45] is an ARP packet sent by a node in order to
      spontaneously cause other nodes to update an entry in their ARP
      cache.  A gratuitous ARP MAY use either an ARP Request or an ARP
      Reply packet.  In either case, the ARP Sender Protocol Address and
      ARP Target Protocol Address are both set to the IP address of the
      cache entry to be updated, and the ARP Sender Hardware Address is
      set to the link-layer address to which this cache entry should be
      updated.  When using an ARP Reply packet, the Target Hardware
      Address is also set to the link-layer address to which this cache
      entry should be updated (this field is not used in an ARP Request
      packet).

Fix #80

### DIFF
--- a/pkg/vip/arp.go
+++ b/pkg/vip/arp.go
@@ -84,8 +84,8 @@ func gratuitousARPReply(ip net.IP, mac net.HardwareAddr) (*arpMessage, error) {
 		},
 		mac,
 		ip.To4(),
-		ethernetBroadcast,
-		net.IPv4bcast,
+		mac,
+		ip.To4(),
 	}
 
 	return m, nil
--- a/pkg/vip/arp.go
+++ b/pkg/vip/arp.go
@@ -111,9 +111,6 @@ func sendARP(iface *net.Interface, m *arpMessage) error {
 		Halen:    m.hardwareAddressLength,
 	}
 	target := ethernetBroadcast
-	if m.opcode == opARPReply {
-		target = m.targetHardwareAddress
-	}
 	for i := 0; i < len(target); i++ {
 		ll.Addr[i] = target[i]
 	}


thebsdbox:
I’ll take this for a test drive tomorrow. 😀

thebsdbox:
This change results in no arp updates in my network.. 

1. Node01 is leader and will assign the VIP to `ens160` it will also begin arp updates
2. reboot Node01
3. Node02 will assign VIP to `ens160` and begin arp broadcasts .. 
4. remote hosts should immediately see the new MAC <--> VIP however the arp broadcast doesn't work.

jzhoucliqr:
surprise to me that the change cause issue in your network. So the other nodes don't receive the broadcast? How about a direct arping, which is using ARPReply with the same format that this change use, does that go through to other nodes?

thebsdbox:
I think we can drop this in favour of the other PR, along with the cleaner ARP request/reply format? does that sound OK?

yastij:
sounds good to me, let's discuss this in the CAPV office hours

jzhoucliqr:
@yastij do you see the issue @thebsdbox mentioned in your testing with this PR?

yastij:
@jzhoucliqr - sorry my last comment wasn't clear, yes I'm seeing the same issue, the VIP isn't failing over

jzhoucliqr:
close in favor of #81

jzhoucliqr:
close in favor of #81

PR #95, opened 2020-11-04
Fixes to int parsing

thebsdbox:
I've no clue why the incorrect int parsing ended up in the code.

This fixes `i, err := strconv.ParseInt(env, 8, 0)` => **`i, err := strconv.ParseInt(env, 10, 32)`**

The incorrect `parseInt` would break larger numbers i.e. "15" becomes `13`


dhawal55:
Yes, I saw this when adding port and was wondering about it

thebsdbox:
Yeah ... some tests are really needed. These have always been used with small numbers so the rounding error hasn't caused issues.

PR #96, opened 2020-11-05
This adds upnp functionality to kube-vip

### LINKED ISSUE #58: Expose Load-balancer port through uPNP
This is technically possible to expose the port mapping to a commercial router so we could:

- Create VIP
- Create Load-balancer
- Expose VIP:Port to router:port through uPNP

`kubectl expose deployment nginx-deployment --port=18080 --type=LoadBalancer --name=nginx-loadbalancer`

The `--port=xxx` will need to be considered for the uPNP endpoint (router)

Errors will be hard to handle as it's not something that Kubernetes load-balancers are designed to handle, errors would exist the `kube-vip` pod in the namespace the load-balancer is created in. This is where an end-user would find that the exposed port hasn't made it to the router (such as having an already existing port-mapping created).

cc / @displague

thebsdbox:
With upnp functionality enabled any loadbalancer will be exposed through a upnp gateway! Fixes #58 

Finally worked it out!

**Enabled UPNP in `kube-vip` manifest:**
```
        env:
          - name: enableUPNP
            value: "true"
```

**Expose a service:**

```
dan@k8s01:~$ k expose deployment nginx-deployment --port=32380 --target-port=80 --type=LoadBalancer --name=nginx-loadbalance
service/nginx-loadbalance exposed
dan@k8s01:~$ k get svc
NAME                TYPE           CLUSTER-IP       EXTERNAL-IP     PORT(S)           AGE
kubernetes          ClusterIP      10.96.0.1        <none>          443/TCP           21h
nginx-loadbalance   LoadBalancer   10.110.170.214   192.168.0.201   32380:32479/TCP   3s
```

**Check upnp logs on router:**
```
$ cat /var/log/upnp.leases | grep nginx
TCP:32380:192.168.0.201:32380:0:nginx-loadbalance
```

**Test from outside world:**
```
$ curl <external_ip>:32380
<!DOCTYPE html>
<html>
<head>
<title>Welcome to nginx!</title>
```

cc/ @displague

### DIFF
--- a/pkg/service/manager.go
+++ b/pkg/service/manager.go
@@ -8,11 +8,14 @@ import (
 	"os"
 	"os/signal"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"syscall"
 	"time"
 
 	"github.com/davecgh/go-spew/spew"
+	"github.com/kamhlos/upnp"
+
 	dhclient "github.com/digineo/go-dhclient"
 	"github.com/plunder-app/kube-vip/pkg/cluster"
 	"github.com/plunder-app/kube-vip/pkg/kubevip"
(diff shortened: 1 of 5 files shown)


yastij:
queueing this for a review later today

pkg/service/manager.go
```diff
@@ -170,6 +175,21 @@ func (sm *Manager) Start() error {
...
+	upnpEnabled, _ := strconv.ParseBool(os.Getenv("enableUPNP"))
+
+	if upnpEnabled {
+		sm.upnp = new(upnp.Upnp)
+		err := sm.upnp.ExternalIPAddr()
+		if err != nil {
+			log.Errorf("Error Enabling UPNP %s", err.Error())
+			// Set the struct to nil so nothing should use it in future
+			sm.upnp = nil
+		} else {
```
displague, line 188: > Errors will be hard to handle as it's not something that Kubernetes load-balancers are designed to handle, errors would exist the kube-vip pod in the namespace the load-balancer is created in. This is where an end-user would find that the exposed port hasn't made it to the router (such as having an already existing port-mapping created).

Why are we not returning the `err`?

thebsdbox: If the environment variable does't exist OR it isn't a string form of "yes/YES/true/TRUE" then the bool is set to `false` .. the error wouldn't be any use in this context.

PR #98, opened 2020-11-06
LoadBalancer Service changes

thebsdbox:
This removes the TCP LoadBalancer from the path and leaves it up to `kube-proxy` to handle the requests.

### DIFF
--- a/pkg/cluster/singleNode.go
+++ b/pkg/cluster/singleNode.go
@@ -102,7 +102,7 @@ func (cluster *Cluster) StartSingleNode(c *kubevip.Config, disableVIP bool) erro
 	return nil
 }
 
-// StartLoadBalancerService will start a VIP/LoadBalancer instance
+// StartLoadBalancerService will start a VIP instance and leave it for kube-proxy to handle
 func (cluster *Cluster) StartLoadBalancerService(c *kubevip.Config, disableVIP bool) error {
 	// Start a kube-vip loadbalancer service
 	log.Infoln("Starting kube-vip as a LoadBalancer Service")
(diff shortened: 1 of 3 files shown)


pkg/cluster/singleNode.go
```diff
@@ -101,3 +101,48 @@ func (cluster *Cluster) StartSingleNode(c *kubevip.Config, disableVIP bool) erro
...
+	}
+
+	err = cluster.Network.AddIP()
+	if err != nil {
+		log.Warnf("%v", err)
+	}
+
+	if c.GratuitousARP == true {
+		// Gratuitous ARP, will broadcast to new MAC <-> IP
+		err := vip.ARPSendGratuitous(cluster.Network.IP(), c.Interface)
```
yastij, line 125: Are we relying on the user to create the `Service` object ? shouldn't kube-vip be doing this ?

thebsdbox: This change is transparent to an end-user.. 

Before the traffic flowed like this:
`<USER> --> <VIP> --> <Kube-vip TCP port> --> One of the pod endpoints in the service`

Now the traffic flows:
`<USER> --> <VIP> --> <IPTables (managed by kube-proxy)>--> One of the pod endpoints in the service`

thebsdbox: We had to run a daemonset called starboard that would handle IPTables rules so that kube-vip would get the traffic, we no longer need this with this approach.

dhawal55: I'm unclear which part of the code here forwards the VIP traffic to k8s service IP managed by kube-proxy

thebsdbox: haha, yeah I found it confusing too at first.. So the answer is that *nothing* is forwarding the traffic to kube-proxy.. in actual fact traffic is sent to the VIP where `kube-proxy` has a bunch of iptables rules (set up by Kubernetes) to capture and distribute the service traffic to endpoints. 

All this time I was largely fighting against Kubernetes by injecting my own iptables rules and having kube-vip open a port to listen on etc..

thebsdbox:
Been successfully testing this all weekend with additional changes coming into the cloud controller to manage IP addresses.

PR #117, opened 2020-11-30
0.2.2

thebsdbox:
Docs and fixes release

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ The idea behind `kube-vip` is a small self-contained Highly-Available option for
 
 ## Features
 
-Kube-Vip was originally created to provide a HA solutions for the Kubernetes control plane, over time it has evolved to incorporate that same functionality into Kubernetes service type [load-balancers](https://kubernetes.io/docs/concepts/services-networking/service/#loadbalancer).
+Kube-Vip was originally created to provide a HA solution for the Kubernetes control plane, over time it has evolved to incorporate that same functionality into Kubernetes service type [load-balancers](https://kubernetes.io/docs/concepts/services-networking/service/#loadbalancer).
 
 - Control Plane with ARP (Layer 2) or BGP (Layer 3)
 - Control Plane using either [leader election](https://godoc.org/k8s.io/client-go/tools/leaderelection) or [raft](https://en.wikipedia.org/wiki/Raft_(computer_science))
--- a/README.md
+++ b/README.md
@@ -1,8 +1,8 @@
 # kube-vip
 
+High Availability and Load-Balancing 
 
-**NOTE** All documentation of both usage and architecture are now available at [https://kube-vip.io](https://kube-vip.io)
-
+![](https://kube-vip.io/kube-vip.png)
 
 ## Overview
 Kubernetes Virtual IP and Load-Balancer for both control plane and Kubernetes services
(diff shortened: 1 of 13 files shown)


README.md
```diff
@@ -11,15 +11,25 @@ The idea behind `kube-vip` is a small self-contained Highly-Available option for
...
 - Pretty much anywhere else :)
 
-![](/overview.png)
+**NOTE** All documentation of both usage and architecture are now available at [https://kube-vip.io](https://kube-vip.io)
 
-The `kube-vip` application builds a multi-node or multi-pod cluster to provide High-Availability. When a leader is elected, this node will inherit the Virtual IP and become the leader of the load-balancing within the cluster. 
+## Features
 
-When running **out of cluster** it will use [raft](https://en.wikipedia.org/wiki/Raft_(computer_science)) clustering technology 
+Kube-Vip was originally created to provide a HA solutions for the Kubernetes control plane, over time it has evolved to incorporate that same functionality into Kubernetes service type [load-balancers](https://kubernetes.io/docs/concepts/services-networking/service/#loadbalancer).
```
deitch, line 21: "a HA solutions"? Grammatical.

docs/architecture/index.md
```diff
@@ -17,9 +17,9 @@ There are a number of technologies or functional design choices that provide hig
 
 The `kube-vip` service builds a multi-node or multi-pod cluster to provide High-Availability. When a leader is elected, this node will inherit the Virtual IP and become the leader of the load-balancing within the cluster. 
 
-When running **out of cluster** it will use [raft](https://en.wikipedia.org/wiki/Raft_(computer_science) clustering technology 
+When using ARP or layer2 it will use [leader election](https://godoc.org/k8s.io/client-go/tools/leaderelection)
 
-When running **in cluster** it will use [leader election](https://godoc.org/k8s.io/client-go/tools/leaderelection)
+It is also possible to use [raft](https://en.wikipedia.org/wiki/Raft_(computer_science) clustering technology, but this approach has largely been superseded by leader election.
```
deitch, line 22: This implies that it is deprecated. Is anyone using it? Do you have a need to maintain it? If not, maybe just deprecate it explicitly?

docs/architecture/index.md
```diff
@@ -17,9 +17,9 @@ There are a number of technologies or functional design choices that provide hig
 
 The `kube-vip` service builds a multi-node or multi-pod cluster to provide High-Availability. When a leader is elected, this node will inherit the Virtual IP and become the leader of the load-balancing within the cluster.
```
deitch, line 18: I thought that is true only in arp mode. In bgp, all nodes get the VIP?

docs/architecture/index.md
```diff
@@ -58,6 +58,10 @@ Request timeout for icmp_seq 150
 
 Within a Kubernetes cluster, the load-balancing is managed by the `plndr-cloud-provider` which watches all service that are created, and for those of `type: LoadBalancer` will create the configuration for `kube-vip` to consume.
 
+#### Load Balancing (Inside a cluster)
+
+When using `type:LoadBalancer` within a Kubernetes cluster `kube-vip` will assign the VIP to the leader (when using ARP) all to all running Pods (when using BGP). When traffic is directed to a node with the VIP then the rules configured by `kube-proxy` will redirect the traffic to one of the pods running in the service.
```
deitch, line 63: I think this is usually written is `type=LoadBalancer`.

In any case, it probably should be, "when creating a Service of type=LoadBalancer" etc.

deitch: "all to all"?

deitch: It might help to have an introductory paragraph, explaining that Kubernetes itself takes care of getting traffic directed at a Service IP to the correct Pods, so kube-vip deals solely with getting the traffic onto a cluster node, etc.

docs/architecture/index.md
```diff
@@ -131,10 +134,6 @@ data:
   cidr-testing: 192.168.0.220/29
 ``` 
 
-### Starboard Daemonset
-
-The `starboard` pod is a simple service that implements a [client-go](https://github.com/kubernetes/client-go) [watch](https://github.com/plunder-app/starboard/blob/master/main.go#L72) on a Kubernetes `configMap` that contains the network range/cidr that load-balancer addresses will use. It then ensures that this cidr is present in the ipTables of all machines in the cluster, this will ensure that network traffic is directed to `kube-vip` to be load-balancer to the underlying application pods.
-
 ### Kube-vip
 
  The `kube-vip` pod should exist in every namespace that requires load-balancing services, it also implements a "client-go watch" over a `configMap` in its current namespace. These configurations and created and managed by the cloud-provider, when `kube-vip` sees a change happen (i.e. a new service is defined) then it will implement the virtual IP and start the load-balancer. Furthermore `kube-vip` implements a second "watch" over the endpoints of the service and as that service changes (pods die, scaling etc..) so will the the endpoints that `kube-vip` will load-balancer over.
```
deitch, line 139: Is this still correct?

docs/architecture/index.md
thebsdbox: It's still a requirement to run a Load-Balancing cluster without Kubernetes.

docs/architecture/index.md
thebsdbox: gone

PR #124, opened 2020-12-07
0.2.3 release

thebsdbox:
This is a **whopper**

This moves to a `hybrid` manager, allowing for everything to run within a single pod. This is especially useful for bare-metal.


yastij (approved):
/lgtm

Makefile
```diff
@@ -6,7 +6,7 @@ TARGET := kube-vip
 .DEFAULT_GOAL: $(TARGET)
 
 # These will be provided to the target
-VERSION := 0.2.2
+VERSION := 0.2.3
```
displague, line 9: Is this a patch level change or more significant?

pkg/service/manager_bgp.go
```diff
@@ -30,7 +30,7 @@ func (sm *Manager) startBGP() error {
 		// We're using Packet with BGP, popuplate the Peer information from the API
 		if sm.config.EnableBGP {
 			log.Infoln("Looking up the BGP configuration from packet")
-			err = packet.BGPLookup(packetClient, sm.config)
+			err = packet.BGPLookup(packetClient, sm.config, "") // TODO: This will need looking at in the future
```
displague, line 33: Why will this need to be looked at in the future might be a better TODO note.

pkg/manager/manager_bgp.go
```diff
@@ -0,0 +1,118 @@
...
+			if err != nil {
+				log.Error(err)
+			} else {
+				// Set the environment variable with the key for the project
+				os.Setenv("PACKET_AUTH_TOKEN", key)
+				// Update the configuration with the project key
+				projectID = project
+			}
+		}
+		packetClient, err = packngo.NewClient()
```
displague, line 36: It would be nice to include the UserAgent and a ConsumerID here. (ConsumerID can be string, but it doesn't vary between versions like UserAgent)

https://github.com/packethost/packet-cli/blob/2759d28b097c0f735e7324ee40a86a175d3ce71b/cmd/root.go#L60-L64

displague:
I'm only pointing out things that interested me - not offering a thorough review by any means.

thebsdbox:
> I'm only pointing out things that interested me - not offering a thorough review by any means.

That's fine @displague this release is a big bucket of hacks and one big change I've been wanted to do for a while. A lot will be reverted and removed in later versions once I can have an idea (or standards) for annotations to parse.

Makefile
thebsdbox: I'm making versioning up as I go along

PR #137, opened 2020-12-16
Fix permission access in GitHub Action

gianarb:
I use `act` locally to run GitHub Action and apparently it works
differently and on GitHub you have to run apt as sudo.


thebsdbox:
Awesome sauce !

PR #143, opened 2020-12-18
Fix documentation, kube-vip alias for containerd

### LINKED ISSUE #142: containerd kube-vip alias is not working
I tried to set 

`alias kube-vip="ctr run --rm --net-host docker.io/plndr/kube-vip:0.2.3 vip`
and then
```
kube-vip manifest pod \
    --interface $INTERFACE \
    --vip $VIP \
    --controlplane \
    --services \
    --arp \
    --leaderElection
```
as described in
https://github.com/plunder-app/kube-vip/blob/master/docs/hybrid/static/index.md .

With containerd 1.4.3 I get the error
`ctr: OCI runtime create failed: container_linux.go:370: starting container process caused: exec: "manifest": executable file not found in $PATH: unknown`
because the "manifest" is seen as the [CMD] argument to ctr.

**Solution:**
I fixed this by setting
`alias kube-vip="ctr run --rm --net-host docker.io/plndr/kube-vip:0.2.3 vip /kube-vip"` 
instead, this should probably be updated in the docs. 

 - OS/Distro: Ubuntu 20.04
 - Kubernetes Version: v1.19
 - Kube-vip Version: 0.2.3
 - containerd 1.4.3

mhuttner:
Fixes #142

### DIFF
--- a/docs/hybrid/static/index.md
+++ b/docs/hybrid/static/index.md
@@ -19,7 +19,7 @@ This section details creating a number of manifests for various use cases
 The easiest method to generate a manifest is using the container itself, below will create an alias for different container runtimes.
 
 #### containerd
-`alias kube-vip="ctr run --rm --net-host docker.io/plndr/kube-vip:0.2.3 vip"`
+`alias kube-vip="ctr run --rm --net-host docker.io/plndr/kube-vip:0.2.3 vip /kube-vip"`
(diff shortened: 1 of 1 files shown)


thebsdbox (approved):
Awesome !!

Thanks for the fix :-D

PR #148, opened 2020-12-24
Fixes to capabilities and CI

thebsdbox:


thebsdbox:
Hey @gianarb if you're every bored.. I presumed updating the `go.mod` to look in the current directory for the imports would fix things :'( Any ideaS?

gianarb (approved):

Issue #153, opened 2020-12-31
Parse annotations


thebsdbox:
**Is your feature request related to a problem? Please describe.**
Not currently, I'd like to have the daemonset parse annotations if they exist

**Describe the solution you'd like**

```
Annotations:        kubeadm.alpha.kubernetes.io/cri-socket: /var/run/dockershim.sock
                    node.alpha.kubernetes.io/ttl: 0
                    packet.com/node-asn: 65000
                    packet.com/peer-asn: 65530
                    packet.com/peer-ip: 10.80.96.24
                    packet.com/src-ip: 10.80.96.25
```

Read this configuration from the node.

**Describe alternatives you've considered**

**Additional context**

thebsdbox:
@gianarb is this something you'd be interested to look at? we'd probably need to keep `kube-vip` in a waiting state until it can read these annotations for BGP. Or have kube-vip exit.. at least then people will investigate why it isn't starting.

gianarb:
Sure thing @thebsdbox ! Let's assume I don't know anything about this issue and I don't have any context about what you are doing.

Who sets those annotations in the cloud control manager? Are those annotations something kube-vip should wait for before doing any work? If so why?

Thanks

thebsdbox:
Yeah. I guess my idea is a flag that is --annotations "vendor" and there is something that looks for `fmt.Sprintf("%s/peerIP", vendor)` or something like that.

gianarb:
I have to admit I am not sure what you want and why but if you look at this:

> Yeah. I guess my idea is a flag that is --annotations "vendor" and there is something that looks for fmt.Sprintf("%s/peerIP", vendor) or something like that.

You are forcing cloud providers probably in the CCM layer to implement a set of annotations `peerIP`, `asn`, `src` and so on, each of them with their own `vendor` identifier, even if the annotation are all the same.

Why not having `kube-vip/node-asn` and so on? In this way, kube-vip stays vendor-agnostic (looking at your example I can't tell why vendor matters other than for branding)

thebsdbox:
I was following the notation definition as set per the packet CCM? The idea of having a `--annotations "packet.com"` was just incase another CCM opted to advertise BGP configuration and could do it in the same way as the packet CCM.

gianarb:
Ok great

thebsdbox:
Initial use of annotations merged in #157 !!

PR #157, opened 2021-01-20
Annotations watcher

thebsdbox:
This now offers the `--annotations <provider>` flag, this is used to watch for BGP annotations being applied and will watch for: 

`<provider>/src-ip`
`<provider>/node-asn`
`<provider>/peer-ip`
`<provider>/peer-asn`

Once all these have been populated the watch ends and starts the BGP server

### DIFF
--- a/pkg/manager/watcher.go
+++ b/pkg/manager/watcher.go
@@ -3,6 +3,8 @@ package manager
 import (
 	"context"
 	"fmt"
+	"os"
+	"strconv"
 
 	log "github.com/sirupsen/logrus"
 
@@ -10,6 +12,7 @@ import (
 	v1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	watchtools "k8s.io/client-go/tools/watch"
 
 	"k8s.io/apimachinery/pkg/watch"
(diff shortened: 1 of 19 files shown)


deitch:
A few questions:

* Are the annotations configurable? 
* Are the docs updated to reflect this option?
* Will the whole "packet" option go away, as this can just use those?

gianarb (approved):
As we discussed in issue #153 I understand the feature and I know why it is coded in this way (with the possibility to override the prefix). I can't buy that flexibility. It looks a bit confusing for somebody that does not have the same context we have (and the opportunity to ping Dan on a daily bases).

But code is code and it implemented a needed feature!

thebsdbox:
> As we discussed in issue #153 I understand the feature and I know why it is coded in this way (with the possibility to override the prefix). I can't buy that flexibility. It looks a bit confusing for somebody that does not have the same context we have (and the opportunity to ping Dan on a daily bases).
> 
> But code is code and it implemented a needed feature!

If there is a better alternative i'd be happy to implement it? I just need some standard for getting BGP configuration from annotations?

gianarb:
> If there is a better alternative i'd be happy to implement it? I just need some standard for getting BGP configuration from annotations?

I think you should not require a provider as a prefix, you can make your own one. It can be `kube-vip` or `bgp-configure`. It will make kube-vip easier but it will probably require CCM to adapt to your standard (but I think it is something CCM already has to do because only Packet CCM will work out of the box).
If I am wrong all the CCM already set BGP in the right way what I am saying does not have any sense!

detiber:
> > If there is a better alternative i'd be happy to implement it? I just need some standard for getting BGP configuration from annotations?
> 
> I think you should not require a provider as a prefix, you can make your own one. It can be `kube-vip` or `bgp-configure`. It will make kube-vip easier but it will probably require CCM to adapt to your standard (but I think it is something CCM already has to do because only Packet CCM will work out of the box).

+1 on a kube-vip specific the prefix here...  I would say that something like:
`bgp.kube-vip.io/<attribute>` would probably be a good convention to follow for bgp-related annotations.

> If I am wrong and all the CCM implementations already set BGP in the right way what I am saying does not have any sense!

I'm not aware of any other CCM implementations that are interacting with any BGP related config outside of ours.

That said, to ease migrations it might make sense to have some backwards compat support on the kube-vip side at least temporarily.

pkg/manager/watcher.go
```diff
@@ -89,3 +92,120 @@ func (sm *Manager) servicesWatcher(ctx context.Context) error {
...
+	}
+
+	rw, err := watchtools.NewRetryWatcher("1", &cache.ListWatch{
+		WatchFunc: func(options metav1.ListOptions) (watch.Interface, error) {
+			return sm.clientSet.CoreV1().Nodes().Watch(context.Background(), listOptions)
+		},
+	})
+	if err != nil {
+		return fmt.Errorf("error creating annotations watcher: %s", err.Error())
+	}
```
detiber, line 121: This is likely a good candidate to use a metadata only client/informer: https://godoc.org/k8s.io/client-go/metadata

thebsdbox: I'll have to investigate that 🧐

PR #162, opened 2021-01-26
fix: handle multiple peer-ips from ccm

rawkode:


thebsdbox (approved):
Awesome stuff! 

Thanks so much @rawkode

PR #164, opened 2021-02-01
Add prometheus support

gianarb:
We want metrics exposed via Promtheus exposing format.
By default the HTTP server runs on port 2112.


thebsdbox (approved):
Awesome, so by default it will advertise metrics from the Go Runtime.. and I can add in my own from here?

gianarb:
Let's pair on a couple of them tomorrow 👍

PR #170, opened 2021-02-18
leaderelection exit when the election is lost

yastij:
Signed-off-by: Yassine TIJANI <ytijani@vmware.com>

This PR adds code to exit when the cleanup is done. This is especially useful when running on a single node cluster, where there's no other kube-vip pod to takeover the lease


thebsdbox (approved):
`/LGTM`

PR #193, opened 2021-04-15
fix: unassign IPs correctly for Equinix Metal

smira:
IP assignment href returned from the API looks like:

```
/metal/v1/ips/97972920-f00b-453f-be1e-31eb53c19e38
```

It should be cleaned up to contain ID only, removing `/ips/` leads to
broken requests like:

```
2021/04/15 12:03:41 [DEBUG] DELETE https://api.equinix.com/metal/v1/ips//metal/v197972920-f00b-453f-be1e-31eb53c19e38
```

Signed-off-by: Andrey Smirnov <smirnov.andrey@gmail.com>

### DIFF
--- a/pkg/packet/eip.go
+++ b/pkg/packet/eip.go
@@ -2,7 +2,7 @@ package packet
 
 import (
 	"fmt"
-	"strings"
+	"path"
 
 	"github.com/packethost/packngo"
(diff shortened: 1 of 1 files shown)


thebsdbox (approved):
`/lgtm`

PR #196, opened 2021-04-19
Bgp password support from annotations.

thebsdbox:
This now adds the capability for the Password to be pulled from the annotations.


deitch (approved):
LGTM

Issue #202, opened 2021-05-04
digineo/go-dhclient has Limited GPL License


TerryHowe:
**Is your feature request related to a problem? Please describe.**
digineo/go-dhclient has a GPL license which limits the usability of of kube-vip

**Describe the solution you'd like**
There are plenty of other dhcp clients out there and most likely another has a more permissive license.

**Describe alternatives you've considered**
Switch to another client with a permissive license such as https://github.com/insomniacslk/dhcp

thebsdbox:
Makes sense, i'll investigate it tomorrow !

thebsdbox:
Taking a little longer than expected.. we use DHCP in two places in slightly different ways and there isn't a single replacement that offers the same functionality. So i'm having to write my own at this point :-|

thebsdbox:
Fixed in #203

PR #203, opened 2021-05-07
Swap DHCP to a pkg with better licenses

thebsdbox:
This replaces one DHCP package with one that has better licenses.. this incurred a huge amount of changes unfortunately. For what was expecting to be a quick rip and replace.

The DDNS is still to be finished.


thebsdbox:
This will fix #202 cc @terryhowe

pkg/service/services.txt
```diff
@@ -0,0 +1,159 @@
+package service
```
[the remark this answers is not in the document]
thebsdbox: Oh.. It's because I re-wrote the entire thing and kept the old source code round incase I needed it.. I'll remove that file.

thebsdbox:
> I can't tell what the code does but it looks like there is a good amount of code that looks the same in `manager` and `service` package. I am wondering if it sounds useful for you to create a `DHCP package that offers a dedicated API that you can use for both packages. It should improve maintainability by reducing the amount of code. But as I said, maybe it is the only way to have this working :) I can't really tell!

It's technical debt already ! :-( I re-wrote the `manager` of superseding the services part but people are still using it the old way for now.

gianarb:
> It's technical debt already 

Let's call it retro compatibility 👍

pkg/service/services.txt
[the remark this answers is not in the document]
yastij: +1 to remove this, it's confusing

pkg/service/services.go
```diff
@@ -61,209 +59,112 @@ func (sm *Manager) deleteService(uid string) error {
...
+		sm.upnpMap(newService)
 
-			newService.cluster = *c
+		newService.cluster = *c
 
-			// Begin watching this service
-			go sm.newWatcher(&newService)
+		// Begin watching this service
+		// TODO - we may need this
+		// go sm.serviceWatcher(&newService, sm.config.Namespace)
```
yastij, line 145: dead code ?

pkg/service/watcher.go
```diff
@@ -15,102 +15,112 @@ import (
...
+		rw.Stop()
+	}()
 	ch := rw.ResultChan()
-	defer rw.Stop()
-	log.Infof("Beginning watching Kubernetes Endpoints for service [%s]", s.service.ServiceName)
+	//defer rw.Stop()
+	log.Infoln("Beginning watching services for type: LoadBalancer in all namespaces")
 
 	for event := range ch {
+		//sm.countServiceWatchEvent.With(prometheus.Labels{"type": string(event.Type)}).Add(1)
```
yastij, line 72: same ?

pkg/vip/ddns.go
```diff
@@ -51,6 +53,7 @@ func (ddns *ddnsManager) Start() (string, error) {
 			ipCh <- lease.FixedAddress.String()
 		},
 	}
+	//options[DHCPOptHostname] = []byte(client.Hostname)
```
yastij, line 56: same ?

yastij (commented):
lgtm, pending comments

PR #237, opened 2021-07-09
Add example manifest for kube-vip:v0.3.5

abhinavmpandey08:
Add example manifest for kube-vip:v0.3.5 for deploying it as a daemonset for service `type: LoadBalancer`


thebsdbox:
Can we remove the `serviceAccount` bits and leave that to `rbac.yaml`?

abhinavmpandey08:
> Can we remove the `serviceAccount` bits and leave that to `rbac.yaml`?

Done

PR #249, opened 2021-07-26
fix: ensure node annotations are checked before watch

rawkode:
`kube-vip` gets "stuck" waiting for the Metal annotations when they already exist on the nodes.

This change allows for the annotations on the node to be checked prior to enter the watch.

I've got a new environment spinning up now to test and will report back shortly.


rawkode:
@thebsdbox I've added some tests around this and it's help identify some weird behaviour that may or may not be intended.

The `peer-ip` is parsed, comma separated, and stored within the `BGPConfig` struct. However, only a single peer is stored in the `BGPPeerConfig`.

Is this intended?

thebsdbox (approved):
/LATM (Looks Awesome to me)

PR #263, opened 2021-08-31
[metal] Use provided provider id in config if defined

detiber:
Do not perform project lookup by name when projectID is defined in the config.


thebsdbox:
Ah cool, 

Thanks for cleaning this up!

PR #267, opened 2021-09-02
Update DaemonSet generation to improve scheduling

detiber:
- Tolerate more than just the node-role.kubernetes.io/master taint, this allows
  the daemonset to schedule prior to CNI being deployed.

- Switch from using a nodeSelector to using nodeAffinity
  - Allows scheduling even if the node label does not match the value (for the node-role label, the value is generally ignored).
  - Add support for the control-plane node role rather than just master.


thebsdbox (approved):
awesome, much cleaner!

Thanks

PR #285, opened 2021-10-12
Use localhost to connect to manager

thebsdbox:
Instead of `os.Hostname()` just use the `localhost` as this is always present in the certificate SANs.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,7 @@ TARGET := kube-vip
 .DEFAULT_GOAL: $(TARGET)
 
 # These will be provided to the target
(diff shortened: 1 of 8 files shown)


thebsdbox:
Strange why that worked in kind but not in real life ... Anyway

Change api-server to `kubernetes` in kube-vip instead of `localhost` and add this to the hostAliases (which is what coreDNS would do anyway)

```
spec:
  hostAliases:
  - ip: "127.0.0.1"
    hostnames:
    - "kubernetes"
 ```

PR #289, opened 2021-10-19
Auto detect default interface for VIP

thebsdbox:
This adds a `--autoInterface` cmdline that is used to generate manifests and to start kube-vip, when this is used it will look at the default Gateway and use that interface for VIPs moving forward.


yaocw2020:
LGTM

PR #294, opened 2021-10-20
Removes local leaderElection fix

thebsdbox:
A long long long time ago, `leaderElection` was broken 😢  and noone was fixing it. So we grabbed a local copy of the code and fixed it, changed the import paths and VIOLA..

The fixes is now upstream, so we can remove our local copy! 🎉


thebsdbox:
Ah @yastij requires bumping go to 1.16.

PR #317, opened 2021-11-22
Various KIND documentation fixes

lewisdiamond:
KVVERSION needs to be preceeded with `$` for shell expansion of the variable.
`kubectl` is unlikely to be a binary in the current directory and is likely to be in `$PATH`


thebsdbox:
Awesome catch! Thanks

thebsdbox:
Ok doing the DCO thing?

thebsdbox (commented):
:-)

PR #326, opened 2021-12-10
fix ddns support

sadysnaat:
Fixes: #328

### DIFF
--- a/pkg/vip/ddns.go
+++ b/pkg/vip/ddns.go
@@ -46,6 +46,10 @@ func (ddns *ddnsManager) Start() (string, error) {
 		ipCh <- lease.ACK.YourIPAddr.String()
 	})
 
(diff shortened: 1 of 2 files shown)


thebsdbox (changes requested):
Looks like it fixes an introduced bug, can we get a little bit of information?

thebsdbox:
I'm probably going to merge this sooner rather than later as I'd like to cut a new release.. but a note would be helpful @sadysnaat :-D

sadysnaat:
> I'm probably going to merge this sooner rather than later as I'd like to cut a new release.. but a note would be helpful @sadysnaat :-D

@thebsdbox Issue logged at #328

thebsdbox (approved):
AWESOME Thanks !

PR #329, opened 2021-12-16
Check interface

yaocw2020:
In some cases, kube-vip will start to serve before the interface is not
existed or not up. We should make sure the interface is valid.

We met the problem in the Harvester. Here is the issue https://github.com/harvester/harvester/issues/1681


thebsdbox:
Going to re-run the failed test, it has been "flaky" of late.

thebsdbox (approved):
`/LGTM`

PR #333, opened 2021-12-21
provide support for backend traffic forwarding

jbguerraz:
This brings decoupling between vip's address/port and backend's (api server) address/port.
Coupling is setup at start using either iptables or IPVS (depends if load balancer is enabled or not) if the backend address isn't `0.0.0.0` (which is the default value and is backward compatible).

It also adds/removes backends as they become ready/not-ready (in addition to already implemented in / out)

`lbBackEndPort` looks unused and conflict/confuse with new `backendPort` that is used to define the api server port. Maybe better to remove ?

```
rg BackendPort
pkg/kubevip/config_types.go
147:	//BackendPort, is a port that all backends are listening on (To be used to simplify building a list of backends)
148:	BackendPort int `yaml:"backendPort"`

cmd/kube-vip-start.go
39:	kubeVipStart.Flags().IntVar(&startConfigLB.BackendPort, "lbBackEndPort", 6443, "A port that all backends may be using (optional)")
```

Deprecates `lbPort` since `port` is used for the loadbalancer port (and `backendPort` is used for real (api)servers port)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,9 +11,11 @@ RUN --mount=type=cache,sharing=locked,id=gomod,target=/go/pkg/mod/cache \
     --mount=type=cache,sharing=locked,id=goroot,target=/root/.cache/go-build \
     CGO_ENABLED=0 GOOS=linux make build
 
-FROM scratch
-# Add Certificates into the image, for anything that does API calls
-COPY --from=dev /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
+FROM alpine:3.15
+RUN apk add --no-cache iproute2 net-tools ca-certificates iptables && update-ca-certificates
+COPY dist/iptables-wrapper-installer.sh /
+RUN /iptables-wrapper-installer.sh
+
 # Add kube-vip binary
 COPY --from=dev /src/kube-vip /
 ENTRYPOINT ["/kube-vip"]
--- a/go.mod
+++ b/go.mod
@@ -4,6 +4,7 @@ go 1.14
 
 require (
 	github.com/cloudflare/ipvs v0.8.0
+	github.com/coreos/go-iptables v0.6.0
 	github.com/davecgh/go-spew v1.1.1
 	github.com/ghodss/yaml v1.0.0
 	github.com/golang/protobuf v1.5.2
--- a/pkg/forwarder/forwarder.go
+++ b/pkg/forwarder/forwarder.go
@@ -1,0 +1,65 @@
+package forwarder
+
+import (
+	"context"
+
+	log "github.com/sirupsen/logrus"
+)
+
+type Forwarder interface {
+	Start(context.Context) error
+}
+
+type forwarder struct {
+	srcAddress           string
+	srcPort              int
+	dstAddress           string
+	dstPort              int
+	ipvs                 bool
+	ipvsForwardingMethod string
+}
+
+func NewForwarder(srcAddress string, srcPort int, dstAddress string, dstPort int, ipvs bool, ipvsForwardingMethod string) Forwarder {
+	return &forwarder{
+		srcAddress:           srcAddress,
+		srcPort:              srcPort,
+		dstAddress:           dstAddress,
+		dstPort:              dstPort,
+		ipvs:                 ipvs,
+		ipvsForwardingMethod: ipvsForwardingMethod,
+	}
+}
+
+func (f *forwarder) Start(ctx context.Context) error {
+	log.Infoln("starting forwarder")
+	// If backend address is quad-zero route, early return
+	// (destination address is unknown, can't setup forwarding).
+	// If at some point we want to prepare forwarding none the less we would have to pick up any interface
+	// address and continue with it. If the user want to profit from different vip port and backend port
+	// it's probably easier the user define the backend address instead
+	if f.dstAddress == "0.0.0.0" {
+		return nil
+	}
+	go func(ctx context.Context, stop func() error) {
+		<-ctx.Done()
+		if err := stop(); err != nil {
+			log.Errorf("forwarder stop error: %v", err)
+		}
+	}(ctx, f.Stop)
+	if f.ipvs {
+		return startIPVS(f.srcAddress, f.srcPort, f.dstAddress, f.dstPort, f.ipvsForwardingMethod)
+	}
+	startIPTables(ctx, f.srcAddress, f.srcPort, f.dstAddress, f.dstPort)
+	return nil
+}
+
+func (f *forwarder) Stop() error {
+	log.Infof("stopping forwarder")
+	if f.dstAddress == "0.0.0.0" {
+		return nil
+	}
+	if f.ipvs {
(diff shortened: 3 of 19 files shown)


thebsdbox:
A few `golint` checks that have failed.

chipzoller:
If there are implications for docs that need to be covered, can you either submit a PR in kube-vip/website or raise an issue with the details, please?

jbguerraz:
> If there are implications for docs that need to be covered, can you either submit a PR in kube-vip/website or raise an issue with the details, please?

There are few implications for docs which I updated in this PR but sure, I'll move that to kube-vip/website once this is ready (this is yet a WIP).

Dockerfile
```diff
@@ -1,6 +1,6 @@
 # syntax=docker/dockerfile:experimental
 
-FROM golang:1.16.12-alpine3.15 as dev
+FROM golang:1.16.5-alpine3.13 as dev
```
mrueg, line 3: Probably should not change back to an older version of golang

Dockerfile
```diff
@@ -11,9 +11,11 @@ RUN --mount=type=cache,sharing=locked,id=gomod,target=/go/pkg/mod/cache \
     --mount=type=cache,sharing=locked,id=goroot,target=/root/.cache/go-build \
     CGO_ENABLED=0 GOOS=linux make build
 
-FROM scratch
-# Add Certificates into the image, for anything that does API calls
-COPY --from=dev /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
+FROM alpine:3.13
```
mrueg, line 14: Can you use alpine:3.15 here?

go.mod
```diff
@@ -4,6 +4,7 @@ go 1.14
 
 require (
 	github.com/cloudflare/ipvs v0.8.0
+	github.com/coreos/go-iptables v0.4.5
```
mrueg, line 7: why not use https://github.com/coreos/go-iptables/releases/tag/v0.6.0 ?

pkg/forwarder/forwarder.go
```diff
@@ -0,0 +1,65 @@
...
+}
+
+func (f *forwarder) Start(ctx context.Context) error {
+	log.Infoln("starting forwarder")
+	// If backend address is quad-zero route, early return
+	// (destination address is unknown, can't setup forwarding).
+	// If at some point we want to prepare forwarding none the less we would have to pick up any interface
+	// address and continue with it. If the user want to profit from different vip port and backend port
+	// it's probably easier the user define the backend address instead
+	if f.dstAddress == "0.0.0.0" {
```
mrueg, line 40: Do we need to check for ::/0 on ipv6 as well or will this only work on ipv4?

go.mod
jbguerraz: thx; updated

Dockerfile
jbguerraz: Yes, fixed thank you.
FTR: I faced https://github.com/opencontainers/runc/issues/2151 locally; had to update docker to be able to work with alpine > 3.14 as stated here https://wiki.alpinelinux.org/wiki/Release_Notes_for_Alpine_3.14.0#faccessat2

Dockerfile
jbguerraz: Yes, fixed thank you.
FTR: I faced opencontainers/runc#2151 locally; had to update docker to be able to work with alpine > 3.14 as stated here https://wiki.alpinelinux.org/wiki/Release_Notes_for_Alpine_3.14.0#faccessat2

pkg/forwarder/forwarder.go
jbguerraz: As it is it'll only works with ipv4 for iptables variant and probably it'll be fine with ipvs (at least, as much as the current load balancer used by kube-vip).
Could we address the ipv6 support in another contribution ? maybe we could here detect ipv6+iptables forwarding and then trigger a fatal with a message to inform the user the combo issn't yet supported ?

mrueg: Sure probably worth adding a TODO and make users aware of potential limitations

jbguerraz: Done: https://github.com/kube-vip/kube-vip/pull/333/files#diff-c42cd718bf633c200838c1ba185053d11240419f8646f960f7c8ab7d6d8ccdb4R14

thebsdbox:
This all looks great to me, I have but one concern and I might be completely wrong here 😄. Originally Kube-vip came in a `scratch` image, meaning that the only thing that lived within the image was the binary itself. This PR moves Kube-vip to use an alpine base meaning that there are now a number of user land tools and "assuming" shell binaries etc. Given that the kube-vip container can mount things like `admin.conf` is this a security concern 🤔 

cc / @neolit123 @yastij @yaocw2020 @detiber (for comment, if they have any)

neolit123 (commented):
K8s distributed images have moved in the direction of being distroless. Preferably there should be no shell and exploitable bits in the image and only the required tools.

Alpine is considered insecure by some providers.

jbguerraz:
Hello @thebsdbox 

Thank you for the review!

I've just added a commit. Sometime CSI init can fail (related: https://bugzilla.redhat.com/show_bug.cgi?id=1817382) at node add time which cause the node status to be not ready (so the node is removed from the IPVS LB pool and k8s API is no more reachable), I had to separate the add and update logic so to always add a new node to the pool (nevermind its status) in case of initial addition to the cluster.

About base image, since this need iptables binary there is not much alternative AFAIK, at least, if eventually there is a security issue, it must be the true for flannel (https://github.com/flannel-io/flannel/blob/master/Dockerfile.amd64) and kube-proxy (https://github.com/kubernetes/kubernetes/blob/master/build/common.sh#L93) as well: we eventually may ask them how they cope with that.

jbguerraz:
> K8s distributed images have moved in the direction of being distroless. Preferably there should be no shell and exploitable bits in the image and only the required tools.
> 
> Alpine is considered insecure by some providers.

flannel rely on alpine and kube-proxy on debian. we may move to debian as kube-proxy if preferable.

neolit123:
I think the debian for kube proxy only contains iptables and other
dependencies but no shell or package managers.

jbguerraz:
> I think the debian for kube proxy only contains iptables and other dependencies but no shell or package managers.

```
 $ docker exec -it k8s_kube-proxy_kube-proxy-qvqc8_kube-system_0af83979-d489-4979-9ce3-0028b644fe32_0 /bin/sh
# apt -v
apt 2.2.4 (amd64)
# 
```

We can go with the same base image than kube-proxy, I don't mind until we have iptables it's fine.

jbguerraz:
FTR, another option is to move to nftables (https://github.com/google/nftables) which has a netlink interface so we can stay with `scratch` then.

thebsdbox:
Due to security concerns `re: userland` tools in the image we're looking to move this code to nftables.

linsite:
You need to `git commit` with a `--signoff` to sign off the commit to avoid the DCO check. @jbguerraz

thebsdbox:
This has all been re-implemented with health checks in `ipvs`, it will likely move to eBPF when we have time.

PR #336, opened 2021-12-22
removes an incorrect CAP

### LINKED ISSUE #335: Re-evaluate and document required Linux capabilities
Currently, kube-vip requires the following capabilities:

```
NET_ADMIN
NET_RAW
SYS_TIME
```
As a user I'm interested in reducing capabilities to a minimum to increase the security of my systems.

Have required capabilities documented, so users can see why and for which reason they are needed. 

None

**Additional context**
It seems like SYS_TIME was introduced in the beginning of this project. I might miss something here and currently don't understand why it is needed for kube-vip to function.
```
       CAP_SYS_TIME
              Set system clock (settimeofday(2), stime(2), adjtimex(2));
              set real-time (hardware) clock.
```
Introduced in: https://github.com/kube-vip/kube-vip/commit/9258e3e20565615aff2f4c14e4a08648ee647c11

thebsdbox:
Fixes #335 as CAP_SYS_TIME is no longer required..


mrueg:
Thanks for the quick turnaround on the issue. Should it also be removed from the following files? (in these one NET_RAW might be missing)
https://github.com/kube-vip/kube-vip/blob/main/cmd/kube-vip-config.go#L113 
https://github.com/kube-vip/kube-vip/blob/main/testing/e2e/kube-vip.yaml.tmpl#L33

thebsdbox:
TBH the `kube-vip-config.go` we can look at removing.. it is replaced by `kube-vip-manifests.go` the E2E stuff will need modifying though.

PR #339, opened 2021-12-22
Monitor default VIP interface

yaocw2020:
The kube-vip chooses the default interface based on the default route. Once the default route changes, the default interface is very likely invalid. In this case, it would be better to crash the process. The pod will restart immediately if the kube-vip is deployed as a daemonset.

### DIFF
--- a/pkg/vip/util.go
+++ b/pkg/vip/util.go
@@ -1,12 +1,14 @@
 package vip
 
 import (
+	"context"
 	"fmt"
 	"net"
 	"strings"
 	"syscall"
 
 	"github.com/pkg/errors"
+	log "github.com/sirupsen/logrus"
 	"github.com/vishvananda/netlink"
 )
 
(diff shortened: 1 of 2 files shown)


thebsdbox (approved):

pkg/vip/util.go
```diff
@@ -85,3 +87,23 @@ func GetDefaultGatewayInterface() (*net.Interface, error) {
...
+func MonitorDefaultInterface(ctx context.Context, defaultIF *net.Interface) error {
+	routeCh := make(chan netlink.RouteUpdate)
+	if err := netlink.RouteSubscribe(routeCh, ctx.Done()); err != nil {
+		return fmt.Errorf("subscribe route failed, error: %w", err)
+	}
+
+	for {
+		select {
+		case r := <-routeCh:
+			log.Infof("route event: %+v", r)
```
yaocw2020, line 101: prefer to log it as a debug log.

PR #341, opened 2021-12-22
set route event log as debug level

yaocw2020:
Fix comment issue: https://github.com/kube-vip/kube-vip/pull/339#discussion_r773918190


yaocw2020:
@thebsdbox Sorry for that. I will check more carefully before submitting pr.

thebsdbox:
Not a problem, thanks for the quick turn around :D

Issue #342, opened 2021-12-22
0.4.1 has issues bringing up interface on lo


cprivite:
**Describe the bug**
As of 0.4.1, it no longer detects the localhost interface correctly on my test servers (c2.medium.x86 on Equinix Metal). 

Error message:
time="2021-12-22T19:22:26Z" level=fatal msg="lo is not valid interface, reason: lo is not up"

```
ip a output:
root@k8s-cluster1-pool1-cp1:~# ip a
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host
       valid_lft forever preferred_lft forever
2: enp65s0f0: <BROADCAST,MULTICAST,SLAVE,UP,LOWER_UP> mtu 1500 qdisc mq master bond0 state UP group default qlen 1000
    link/ether 40:a6:b7:74:7a:a0 brd ff:ff:ff:ff:ff:ff
3: enp65s0f1: <BROADCAST,MULTICAST,SLAVE,UP,LOWER_UP> mtu 1500 qdisc mq master bond0 state UP group default qlen 1000
    link/ether 40:a6:b7:74:7a:a0 brd ff:ff:ff:ff:ff:ff
4: bond0: <BROADCAST,MULTICAST,MASTER,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP group default qlen 1000
    link/ether 40:a6:b7:74:7a:a0 brd ff:ff:ff:ff:ff:ff
    inet 145.XX.XX.X/31 brd 255.255.255.255 scope global bond0
       valid_lft forever preferred_lft forever
    inet 10.XX.XX.X/31 brd 255.255.255.255 scope global bond0:0
       valid_lft forever preferred_lft forever
    inet6 2604::1/127 scope global
       valid_lft forever preferred_lft forever
    inet6 fe80::0/64 scope link
       valid_lft forever preferred_lft forever
```

**Environment (please complete the following information):**
 - OS/Distro: Ubuntu 20.04
 - Kubernetes Version: 1.23.1
 - Kube-vip Version: 0.4.1

**`Kube-vip.yaml`:**
```
apiVersion: v1
kind: Pod
metadata:
  creationTimestamp: null
[...]

PR #348, opened 2022-01-10
Don't check interface is up if it's loopback.

cprivite:
Signed-off-by: Chris Privitere <cprivite@users.noreply.github.com>

### DIFF
--- a/pkg/kubevip/config_manager.go
+++ b/pkg/kubevip/config_manager.go
@@ -3,6 +3,7 @@ package kubevip
 import (
 	"fmt"
 	"io/ioutil"
+	"net"
 	"os"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


cprivite:
Fixes #342

thebsdbox (approved):
`YAY` also `/LGTM`

pkg/kubevip/config_manager.go
```diff
@@ -205,6 +206,10 @@ func isValidInterface(iface string) error {
 		return fmt.Errorf("get %s failed, error: %w", iface, err)
 	}
 
+	if l.Attrs().Flags&net.FlagLoopback == net.FlagLoopback {
```
displague, line 209: I think a linter would add some space here.

cprivite: golangci-lint, go vet, and go fmt aren't suggesting anything. What are you thinking would be better?

cprivite: golangci-lint, go vet, and go fmt aren't suggesting anything. What are you thinking would be better?

thebsdbox: https://github.com/kube-vip/kube-vip/runs/4764985605?check_suite_focus=true#step:6:116

apparently not :D

displague: I stand corrected - I tried `gofumpt -l -w .` and `golangci-lint run ./... --fix` (with the enabled settings for this project, also with the tinkerbell/tink settings). No change.

PR #350, opened 2022-01-11
Merge pull request #348 from cprivite/cprivite/issue342

cprivite:
Fix the nightly main image build so it works (provide github credentials).
Simplify the github action to use contexts instead of generating environment variables.
Unify the release and main workflows to use the same content other than what they're triggered from.


thebsdbox (approved):
This is perfection 💪🏼

PR #351, opened 2022-01-18
Fix DHCP release bug

yaocw2020:
- The macvlan interface should not be deleted until the DHCP client has
released the IP.

- If the network manager in the operating system uses DHCP to obtain an IP
for the network card, it will occupy the DHCP client port. In this case,
the release function will fail because the unicast connection with the UDP
socket needs the same port. The solution depends on the pr https://github.com/insomniacslk/dhcp/pull/458

Related issue: https://github.com/harvester/harvester/issues/1835

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,6 +2,8 @@ module github.com/kube-vip/kube-vip
 
 go 1.18
 
+replace github.com/insomniacslk/dhcp => github.com/harvester/dhcp v0.0.0-20220421024905-28b38eafefe3
+
 require (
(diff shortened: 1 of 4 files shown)


thebsdbox:
Awesome, waiting for the dependant PR at this time.

thebsdbox (approved):

yaocw2020:
We replace the `github.com/insomniacslk/dhcp` with `github.com/harvester/dhcp`. Once the DHCP Pr is merged with the upstream repo, we'll take the origin package back.

thebsdbox (approved):

PR #358, opened 2022-02-03
add --cidr example and explaination

viknet365:
Using the default 32 never worked (Ubuntu bare metal on a to-be control plane)


thebsdbox:
Thanks for this! Unfortunately due to a number of requests I've had to refactor the subnet code so this change is unfortunately out of date.

PR #367, opened 2022-02-26
Retry on service status update in case of conflicts

xprt64:
When a service is deleted then recreated it is possible that the `updateStatus` operation fails due to a conflict error.
In this case, a retry can be made. In order for the second update to work we need to fetch the service resource again and then update the status.

Signed-off-by: Constantin Galbenu <xprt64@gmail.com>


thebsdbox (approved):
Awesome, this looks good to me.

PR #370, opened 2022-02-28
Add s390x arch support

### LINKED ISSUE #369: Add support for s390x arch
No. Increasing usage of kubernetes on s390x require additional tools such as kube-vip on s390x

Go is cross compilable. So adding s390x arch should not be difficult.

rohitsakala:
Fixes #369


thebsdbox (approved):

thebsdbox:
I do worry that the build time is going to be very long with all these architectures 😂

PR #388, opened 2022-04-08
Don't check status on Point-To-Point interfaces

mrueg:
If I remember correctly, point-to-point interfaces will stay in UNKNOWN as well and never go to UP.
(Currently trying to use kube-vip to expose a Kubernetes API server on a wireguard network)


thebsdbox (approved):
Awesome, thanks

PR #389, opened 2022-04-11
Treat interfaces with unknown state as valid

### LINKED ISSUE #385: Error Interface is not up in v0.4.2
**Describe the bug**
Kube-vip refuses to start arguing that the interface is not up, but it is. Furthermore, doing exactly the same steps but using v0.4.0 instead of v0.4.2 works flawlessly.

**To Reproduce**
1. Create a manifest using `ctr run --rm --net-host ghcr.io/kube-vip/kube-vip:v0.4.2 vip /kube-vip manifest pod --interface ens33 --address 192.168.140.200 --controlplane --arp --leaderElection` and place it /etc/kubernetes/manifests/kube-vip.yaml
2. Init the cluster with `kubeadm init --config kubeadm-init.yml --upload-certs` (config file provided at the end of the issue)
3. When the installation fails check the container logs with `crictl --runtime-endpoint /run/containerd/containerd.sock logs CONTAINERID`

The log will show something like this
```
time="2022-03-29T07:59:55Z" level=fatal msg="ens33 is not valid interface, reason: ens33 is not up"
```
**Expected behavior**
I expect it to work normally, as it does with v0.4.0

 - OS/Distro: Flatcar 3033.2.4
 - Kubernetes Version: v1.23.5
 - Kube-vip Version: v0.4.2

**`Kube-vip.yaml`:**
```yaml
apiVersion: v1
kind: Pod
metadata:
  creationTimestamp: null
  name: kube-vip
  namespace: kube-system
spec:
  containers:
  - args:
    - manager
    env:
    - name: vip_arp
      value: "true"
    - name: port
      value: "6443"
    - name: vip_interface
      value: ens33
    - name: vip_cidr
      value: "32"
    - name: cp_enable
      value: "true"
    - name: cp_namespace
      value: kube-system
    - name: vip_ddns
      value: "false"
    - name: vip_leaderelection
      value: "true"
    - name: vip_leaseduration
      value: "5"
    - name: vip_renewdeadline
      value: "3"
    - name: vip_retryperiod
      value: "1"
    - name: address
      value: 192.168.140.200
    image: ghcr.io/kube-vip/kube-vip:v0.4.2
    imagePullPolicy: Always
    name: kube-vip
    resources: {}
    securityContext:
      capabilities:
        add:
        - NET_ADMIN
        - NET_RAW
    volumeMounts:
      name: kubeconfig
  hostAliases:
  - hostnames:
    - kubernetes
    ip: 127.0.0.1
  hostNetwork: true
  volumes:
  - hostPath:
      path: /etc/kubernetes/admin.conf
    name: kubeconfig
status: {}
```
**` kubeadm-init.yml`:***
```yaml
apiVersion: kubeadm.k8s.io/v1beta3
kind: InitConfiguration
nodeRegistration:
  criSocket: /run/containerd/containerd.sock
  kubeletExtraArgs:
    volume-plugin-dir: "/opt/libexec/kubernetes/kubelet-plugins/volume/exec/"
---
apiVersion: kubeadm.k8s.io/v1beta3
kind: ClusterConfiguration
controlPlaneEndpoint: "192.168.140.200:6443"
controllerManager:
  extraArgs:
    flex-volume-plugin-dir: "/opt/libexec/kubernetes/kubelet-plugins/volume/exec/"
---
kind: KubeletConfiguration
apiVersion: kubelet.config.k8s.io/v1beta1
cgroupDriver: systemd
```
**Extra context**
This looks similar to #360.

Thank you for this great application.

claudio4:
Previous commits addressed the issue of interface being incorrectly marked as
down for loopback (1a4465e) and point-to-point (051bb1f) interfaces by directly
passing the test when the interface is identified as any of those types. Both
commits failed to address the core issue, those interfaces (and many others) do
not publish their operational state resulting in the state UNKOWN. With this
commit, now interfaces with unknown status are accepted but the user is warned as
he needs to manually ensure that the interface is ready to handle traffic.
Fix #385.

### DIFF
--- a/pkg/kubevip/config_manager.go
+++ b/pkg/kubevip/config_manager.go
@@ -210,7 +210,10 @@ func isValidInterface(iface string) error {
 	// userspace has set operational state. Interface must be considered for user
 	// data as setting operational state has not been implemented in every driver."
 	if attrs.OperState == netlink.OperUnknown {
-		log.Warningf("the status of the interface %s is unknown. Ensure your interface is ready to accept traffic, if so you can safely ignore this message")
+		log.Warningf(
+			"the status of the interface %s is unknown. Ensure your interface is ready to accept traffic, if so you can safely ignore this message",
+			iface,
+		)
 	} else if attrs.OperState != netlink.OperUp {
 		return fmt.Errorf("%s is not up", iface)
 	}
--- a/pkg/kubevip/config_manager.go
+++ b/pkg/kubevip/config_manager.go
@@ -3,7 +3,6 @@ package kubevip
 import (
 	"fmt"
 	"io/ioutil"
-	"net"
 	"os"
 	"strconv"
 	"strings"
@@ -203,16 +202,16 @@ func isValidInterface(iface string) error {
 	if err != nil {
 		return fmt.Errorf("get %s failed, error: %w", iface, err)
 	}
+	attrs := l.Attrs()
 
-	if l.Attrs().Flags&net.FlagLoopback == net.FlagLoopback {
-		return nil
-	}
-
-	if l.Attrs().Flags&net.FlagPointToPoint == net.FlagPointToPoint {
-		return nil
-	}
-
-	if l.Attrs().OperState != netlink.OperUp {
+	// Some interfaces (included but not limited to lo and point-to-point
+	//	interfaces) do not provide a operational status but are safe to use.
+	// From kernek.org: "Interface is in unknown state, neither driver nor
+	// userspace has set operational state. Interface must be considered for user
+	// data as setting operational state has not been implemented in every driver."
+	if attrs.OperState == netlink.OperUnknown {
+		log.Warningf("the status of the interface %s is unknown. Ensure your interface is ready to accept traffic, if so you can safely ignore this message")
+	} else if attrs.OperState != netlink.OperUp {
 		return fmt.Errorf("%s is not up", iface)
 	}
 


pkg/kubevip/config_manager.go
```diff
@@ -203,16 +202,16 @@ func isValidInterface(iface string) error {
...
-	}
-
-	if l.Attrs().OperState != netlink.OperUp {
+	// Some interfaces (included but not limited to lo and point-to-point
+	//	interfaces) do not provide a operational status but are safe to use.
+	// From kernek.org: "Interface is in unknown state, neither driver nor
+	// userspace has set operational state. Interface must be considered for user
+	// data as setting operational state has not been implemented in every driver."
+	if attrs.OperState == netlink.OperUnknown {
+		log.Warningf("the status of the interface %s is unknown. Ensure your interface is ready to accept traffic, if so you can safely ignore this message")
```
thebsdbox, line 213: Linting error here :-(

pkg/kubevip/config_manager.go
```diff
@@ -203,16 +202,16 @@ func isValidInterface(iface string) error {
...
+	} else if attrs.OperState != netlink.OperUp {
 		return fmt.Errorf("%s is not up", iface)
 	}
```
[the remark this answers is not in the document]
claudio4: I see your point but for just two checks if prefer a good old if.
All the other states are no go, check https://www.kernel.org/doc/html/latest/networking/operstates.html for refence.

pkg/kubevip/config_manager.go
claudio4: My golangci-lint kept crashing (turnouts it doesn't like being installed via Nix) so I didn't notice. Sorry

mkumatag:
@thebsdbox wondering when are we releasing a build with this fix?

thebsdbox:
@mkumatag About 2 hours ago 😂

thebsdbox:
https://github.com/kube-vip/kube-vip/pkgs/container/kube-vip/19806349?tag=v0.4.4

Issue #394, opened 2022-05-13
Kube-vip fails to take lease after upgrading from 0.3.6 to 0.4.4


ggaurav10:
**Describe the bug**
Kube-vip fails to take lease after upgrading from 0.3.6 to 0.4.4
Looks like the hostname is not replaced in the later version when getting new client from config
old code replaces hostname in all cases: https://github.com/kube-vip/kube-vip/blob/v0.3.6/pkg/cluster/clusterLeader.go#L52-L72
new code replaces hostname only if `path` is empty: https://github.com/kube-vip/kube-vip/blob/main/pkg/cluster/clusterLeaderElection.go#L44-L60

**To Reproduce**
1. Run `kube-vip start` with `vip_arp` set to `true`
2. Upgrade from 0.3.6 to 0.4.4

**Expected behavior**
leader election works after upgrade

**Environment (please complete the following information):**
 - OS/Distro: [e.g. Ubuntu 1804]
 - Kubernetes Version: [e.g. v.1.18]
 - Kube-vip Version: 0.4.4

**`Kube-vip.yaml`:**
```

```

**Additional context**

AxiomSamarth:
Thanks for raising this issue @ggaurav10! I want to add to it.

> new code replaces hostname only if path is empty: https://github.com/kube-vip/kube-vip/blob/main/pkg/cluster/clusterLeaderElection.go#L44-L60

I think, the new code replaces the hostname with `os.Hostname()` if **_path is empty_** and if **_`kube-vip` runs out-of-cluster (i.e., not inCluster)_**. The issue we faced is during an inCluster deployment where it gets started as a static pod and won't have the VIP up before trying to connect to the API server. 

So, there is a need to set the `Api-endpoint` to the host to ensure connectivity.

AxiomSamarth:
@bradbury26 So, what is your expectation?

bradbury26:
@AxiomSamarth I would expect the logic to work the same as v0.3.9 in that regardless of whether or not an admin.conf has been supplied the host should be set to kubernetes:6443 if EnableControlPlane is true. Unless there is a genuine reason as to why this was changed in later versions?

AxiomSamarth:
Okay. Thanks for the elaboration, @bradbury26 . Let me blame the code and try to dig in if I can find an explanation for the change introduced. Let me get back to you in a couple of days.

PR #395, opened 2022-05-13
Set api-endpoint to host for incluster deployment

AxiomSamarth:
This PR is a proposed fix for issue #394

### DIFF
--- a/pkg/cluster/clusterLeaderElection.go
+++ b/pkg/cluster/clusterLeaderElection.go
@@ -43,9 +43,13 @@ type Manager struct {
 func NewManager(path string, inCluster bool, port int) (*Manager, error) {
 	var hostname string
 
-	// If the path passed is empty and not running in the cluster,
+	// If inCluster is set then it will likely have started as a static pod or won't have the
+	// VIP up before trying to connect to the API server, we set the API endpoint to this machine to
+	// ensure connectivity. Else if the path passed is empty and not running in the cluster,
 	// attempt to look for a kubeconfig in the default HOME dir.
-	if len(path) == 0 && !inCluster {
+	if inCluster {
+		hostname = fmt.Sprintf("kubernetes:%v", port)
+	} else if len(path) == 0 && !inCluster {
 		path = filepath.Join(os.Getenv("HOME"), ".kube", "config")
 
 		// We modify the config so that we can always speak to the correct host


thebsdbox:
Ah cool, thanks for the PR! testing now :D

thebsdbox (approved):
`/lgtm`

AxiomSamarth:
Thank you so much for the lightning-fast review & merge, @thebsdbox ! ❤️

AxiomSamarth:
@thebsdbox How should the pod manifest look like with `inCluster` set to true? Somehow, my static pod does not seem to start.

```
spec:
  containers:
  - args:
    - start     
    - --inCluster=true
```

Is this how it should look like or should the manifest have the flag `--inCluster` supplied in a different way?

thebsdbox:
```
export KVVERSION=v0.4.4
alias kube-vip="docker run --network host --rm ghcr.io/kube-vip/kube-vip:$KVVERSION"
kube-vip manifest daemonset     --interface $INTERFACE    --inCluster    <other flags>
```

thebsdbox:
A new release tagged `main` should appear soon with your code changes in it.

ggaurav10:
@AxiomSamarth @thebsdbox do we need to handle the following case too - `if len(path) == 0 && inCluster`?
Also, should we use `hostname = os.Hostname():port` instead of `hostname = kubernetes:port`? This way we wouldn't break existing use cases where pod spec isn't using the following field:
```
  hostAliases:
  - hostnames:
    - kubernetes
    ip: 127.0.0.1
```

wdyt?

ggaurav10:
@AxiomSamarth @thebsdbox do we need to handle the following case too - `if len(path) == 0 && inCluster` by setting `path = filepath.Join(os.Getenv("HOME"), ".kube", "config")`?
Also, should we use `hostname = os.Hostname():port` instead of `hostname = kubernetes:port`? This way we wouldn't break existing use cases where pod spec isn't using the following field:
```
  hostAliases:
  - hostnames:
    - kubernetes
    ip: 127.0.0.1
```

wdyt?

AxiomSamarth:
@thebsdbox I think, the if condition to check if the `inCluster` flag is set is not needed [here](https://github.com/kube-vip/kube-vip/blob/b9014ae32a180620bd61e3f6c8b03c45147adaae/pkg/cluster/clusterLeaderElection.go#L50-L52). 

We can have the `hostname = fmt.Sprintf("kubernetes:%v", port)` as default which gets updated to `hostname = fmt.Sprintf("%s:%v", id, port)` it it is out-of-cluster [here](https://github.com/kube-vip/kube-vip/blob/b9014ae32a180620bd61e3f6c8b03c45147adaae/pkg/cluster/clusterLeaderElection.go#L53-L62).

This way, user will also not have to undergo the process of setting up `serviceaccounttoken`, `ca.crt`,  `KUBERNETES_SERVICE_HOST `& `KUBERNETES_SERVICE_PORT ` to generate the `clientSet` when `inCluster` flag is set. 

WDYT?

Issue #330, opened 2021-12-18
[Feature] Distribute VIPs in different nodes


yaocw2020:
**Is your feature request related to a problem? Please describe.**
All the VIPs will be at the same node, which causes a traffic bottleneck.

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

thebsdbox:
The idea here is to do a leaderElection per vip. This would distribute them across all kube-vip pods in the cluster.

yaocw2020:
> The idea here is to do a leaderElection per vip. This would distribute them across all kube-vip pods in the cluster.

But they would not be average.

thebsdbox:
Implemented!

Issue #258, opened 2021-08-20
Implement externalTrafficPolicy options


enkelprifti98:
**Is your feature request related to a problem? Please describe.**
I would like to have visibility into client source IPs. The `externalTrafficPolicy` option allows that as described here:

https://kubernetes.io/docs/tasks/access-application-cluster/create-external-load-balancer/#preserving-the-client-source-ip

**Describe the solution you'd like**

Kube-vip should behave based on what the `externalTrafficPolicy` options has been set to for the service. Default should be `externalTrafficPolicy: Cluster`

**Describe alternatives you've considered**

MetalLB implements the `externalTrafficPolicy` option as shown here:

https://metallb.universe.tf/usage/#traffic-policies

**Additional context**
None.

PR #407, opened 2022-06-08
Initial design of leaderElection per service

thebsdbox:
This will add in a new capability to create a services watcher, on a new service all nodes will participate in a leaderElection (the lease being the service name). The leader will then advertise the VIP to the network, this will enable two things:

- Spread VIPs over ARP (layer 2), not an elegant solution.. but a solution non-the-less
- Allow external mode local, allowing direct traffic to pods (leader Election will be limited to nodes where pods are running locally)

### DIFF
--- a/cmd/kube-vip.go
+++ b/cmd/kube-vip.go
@@ -105,7 +105,7 @@ func init() {
 	kubeVipCmd.PersistentFlags().StringVar(&initConfig.Annotations, "annotations", "", "Set Node annotations prefix for parsing")
 
 	// Control plane specific flags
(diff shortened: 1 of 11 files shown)


thebsdbox:
Fixes #330 and #258 .. and probably some others.

thebsdbox:
Cc/ @cellebyte

Cellebyte:
@thebsdbox will test it in my small production cluster. 
The Table Mode I think will be a separate topic as it is another way of distributing routes.

thebsdbox:
It's still draft at the moment, once it is merged in (with docs and testing) you should be able to just grab the nightly build.

PR #410, opened 2022-06-21
This enables local traffic policy

thebsdbox:
This ensures a node won't participate in the leaderElection when no local pods are found

`External Traffic Policy:  Local`


JhoLee:
@thebsdbox Great change, THANKS! When will it be released? 🙃 I want this feature to deploy our Ingress Controller as Deployment, with setting `spec.externalTrafficPolicy=Local`.

PR #422, opened 2022-07-14
Fixes close nil channel panic

scuzhanglei:


thebsdbox (approved):
Fantastic, thanks for this catch!

PR #434, opened 2022-08-07
Allow fixed virtual MAC for DHCP VIPs

### LINKED ISSUE #354: Support DHCP static leases.

Currently, when deploying a service with `loadBalancerIP: 0.0.0.0`, it gets a random MAC address. I'd like to assign static DHCP leases in my router for some services, either based on a specific MAC address or preferably a client identifier.

I saw that in https://github.com/kube-vip/kube-vip/pull/229 a `kube-vip.io/hwaddr` annotation was added. I'm not sure if this is supposed to work as I expected it to: I added a `kube-vip.io/hwaddr: b6:82:16:b2:86:b2` to one of my services and a static lease with this MAC address to my router. It seems like `kube-vip` then actually tries to use this MAC address, however the service IP stays `<pending>` and the logs show:

```
Could not bring up interface [vip-bf85e6f5] : address already in use
```

Allow specifying a fixed MAC address or client identifier using an annotation which will be used to get the DHCP static lease.

- Using a static IP for the service, but in that case the IP needs to be managed both in the service definition and the router.
- Maybe UPNP would work for my use case, however as far as I understand it, `enableUPNP` will currently enable it globally. I would only want to enable it for some services, not all.

**Additional context**

N/A

valtzu:
Use IP instead of MAC address to determine if the virtual interface was already previously configured. This makes it possible to pre-define `kube-vip.io/hwaddr` annotation on a `LoadBalancer` and therefore allows defining a static DHCP lease for it.

You can test this by using `valtzu/kube-vip:v0.5.0` instead of `ghcr.io/kube-vip/kube-vip:v0.5.0`.

Closes #354

### DIFF
--- a/pkg/manager/instance.go
+++ b/pkg/manager/instance.go
@@ -168,7 +168,7 @@ func (i *Instance) startDHCP() (chan string, error) {
 	}
 
 	var initRebootFlag bool
-	if i.dhcpInterfaceHwaddr != "" {
+	if i.dhcpInterfaceIP != "" {
 		initRebootFlag = true
 	}
 
--- a/pkg/service/services_dhcp.go
+++ b/pkg/service/services_dhcp.go
@@ -65,7 +65,7 @@ func (sm *Manager) createDHCPService(newServiceUID string, newVip *kubevip.Confi
 	}
 
 	var initRebootFlag bool
-	if newService.dhcpInterfaceHwaddr != "" {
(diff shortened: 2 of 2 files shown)


thebsdbox:
@yaocw2020 can you have a quick look at this?

yaocw2020 (approved):
LGTM

PR #435, opened 2022-08-15
Enable kube-vip to only respond to svc with kube-vip's lb class

abhinavmpandey08:
Currently, there's no way to restrict `kube-vip` to only acknowledge services with kube-vip's LB class `kube-vip.io/kube-vip-class`.

For example, if you have two services A and B:
- service A has lb class set to `kube-vip.io/kube-vip-class`
- service B no has lb class set

`kube-vip` will currently advertise IPs for both services A and B

With this change, if a user passes the `--lbClassOnly` flag, `kube-vip` will only advertise IPs for service A since it has `kube-vip`'s LB class but will ignore service B

If a user does not pass the `--lbClassOnly`, there will no change from the current behavior


thebsdbox (approved):
Took a little bit of mental arithmetic to understand the "if" logic, but looks good.

PR #449, opened 2022-09-13
Add random MAC address generation for DHCP

daper:
I was facing an issue when a new macvlan link was added as it was added with `nil` mac; letting the underlying software generate it. Therefore the DHCP request was being made with a non existent MAC and responses were never catched.

This PR adds one randomly generated MAC address when it's empty.


thebsdbox (approved):
I guess 00:00:6C is private, I don't know what that mean for switch/vendor devices that filter .. happy to merge as is and see if it causes issue.

PR #476, opened 2022-11-07
Adds tags to rules so that we can garbage collect.

thebsdbox:


thebsdbox:
This tags all kube-vip created rules so that we can garbage collect the old ones when we restart kube-vip.

Issue #393, opened 2022-05-02
Kube-vip does not work with newer Equinix Metal annotations


enkelprifti98:
**Describe the bug**
The Equinix Metal CCM adds updated node annotations from [CCM 3.3.0](https://github.com/equinix/cloud-provider-equinix-metal/releases/tag/v3.3.0) onwards and kube-vip does not work anymore.

**To Reproduce**
1. Provision k8s cluster with Equinix Metal CCM + kubevip in [annotations mode](https://kube-vip.chipzoller.dev/docs/usage/equinix/#using-annotations).

```
kube-vip manifest daemonset \
  --interface $INTERFACE \
  --services \
  --bgp \
  --annotations metal.equinix.com \
  --inCluster | k apply -f -
```

This is potentially caused by the annotation changes made [here](https://github.com/equinix/cloud-provider-equinix-metal/pull/204).

displague:
The newer service annotations are documented in this table <https://github.com/equinix/cloud-provider-equinix-metal/#configuration> and look like the following:

```
      metal.equinix.com/bgp-peers-0-bgp-pass: ""
      metal.equinix.com/bgp-peers-0-node-asn: "65000"
      metal.equinix.com/bgp-peers-0-peer-asn: "65530"
      metal.equinix.com/bgp-peers-0-peer-ip: 10.1.40.130
      metal.equinix.com/bgp-peers-0-src-ip: 10.1.40.131
      metal.equinix.com/network-4-private: 10.1.40.128/25
```

enkelprifti98:
Hey @thebsdbox ! Any update on this issue?

thebsdbox:
This is now merged, support will in `latest` and the upcoming `0.5.11` release.

PR #521, opened 2023-02-23
check for new equinix ccm annotations

ocobleseqx:
This PR aims to fix issue https://github.com/kube-vip/kube-vip/issues/393

With this solutions it should work with both legacy and new annotations format since it looks for `metal.equinix.com/bgp-peers-0-<info>` or `metal.equinix.com/<info>` an as I understand it, it only works with one neighbour, so `metal.equinix.com/bgp-peers-1-<info>` annotations are not needed

The existing test https://github.com/kube-vip/kube-vip/blob/a92c536de25683f93161819c6630b62538166f16/pkg/manager/watcher_test.go#L67-L83 might be enough to check if this change works

### DIFF
--- a/pkg/manager/manager_bgp.go
+++ b/pkg/manager/manager_bgp.go
@@ -37,7 +37,7 @@ func (sm *Manager) startBGP() error {
 			log.Error(err)
 		}
 
-		// We're using Packet with BGP, popuplate the Peer information from the API
+		// We're using Packet with BGP, populate the Peer information from the API
 		if sm.config.EnableBGP {
 			log.Infoln("Looking up the BGP configuration from packet")
 			err = equinixmetal.BGPLookup(packetClient, sm.config)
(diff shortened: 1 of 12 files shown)


thebsdbox:
This looks fantastic, thanks so much.. the docs are now being moved to the website repo.

pkg/manager/manager_bgp.go
```diff
@@ -37,7 +37,7 @@ func (sm *Manager) startBGP() error {
 			log.Error(err)
 		}
 
-		// We're using Packet with BGP, popuplate the Peer information from the API
+		// We're using Packet with BGP, populate the Peer information from the API
```
displague, line 40: `s/Packet/Equinix Metal/`?  likewise for `s/packet/Equinix Metal/` below in the log message.

ocobleseqx: @thebsdbox should I open a new PR to update those Packet refs?

thebsdbox: No, happy to keep this all under one EM change.

thebsdbox (approved):
Awesome, `/LGTM`

Issue #523, opened 2023-02-27
IPv6: Leader should recover when address goes into dadfailed state


tylerschultz:
**Describe the bug**

We've encountered a situation with IPv6 where the leader node has added the address to the device, but the address is unreachable. It was seen that the kube-vip address is in dadfailed state. Kube-vip is unaware that the address is in a failed state and takes no corrective action. The kernel or anything otherwise takes no corrective action either, as far as we can tell. We were able to restore connectivity to the kube-vip address by deleting it from the device, and allowing kube-vip to re-add it in it's periodic check, [in the EnableARP loop](https://github.com/kube-vip/kube-vip/blob/main/pkg/cluster/service.go#L131-L137). 

Our guess as to why this happens: When a leader election happens, the old leader has not yet finished deleting the address from it's device, and the new leader has added the address to it's device. This overlap means that IPv6's duplicate address detection algorithm kicks in, and sets the address to the `dadfailed` state.

**To Reproduce**
This problem seems to occur occasionally. We haven't found any steps to reproduce this problem reliably.

**Expected behavior**
Because the address is 'owned' by the kube-vip leader, we are of the opinion that the leader should periodically check to see if the address is in `dadfailed` state, and if so, remove and re-add the address. The scenario that we saw, the previous leader was no longer using the kube-vip address, and only the new leader was holding the address, but in a dadfailed state.

**Screenshots**
N/A

**Environment (please complete the following information):**
 - OS/Distro: Ubuntu 20.04
 - Kubernetes Version:  v.1.24
 - Kube-vip Version:  0.4.2

**`Kube-vip.yaml`:**
N/A

**Additional context**
We intend to open a PR to address this problem.

PR #524, opened 2023-03-01
Readd VIP if address has dadfailed flag

tylerschultz:
Remove and re-add the address to the interface in the case that a duplicate address is detected. Without this change, kube-vip will take no corrective action, and the address will remain in dadfailed state. Removing and re adding the address causes duplicate address detection to happen again, hopefully successfully the next time. This behavior only applys to IPv6 addresses, as IPv4 addresses cannot get into the dadfailed state.

We suspect dadfailed happens when a new leader is elected and the old leader is too slow or fails to delete the IP address.

We attempted to add a test, but we find it impossible to reliably trigger a dadfailed state on the leader node. During manual testing we were occasionally trigger the scenario and see that this PR has the desired effect.

fixes: #523

### DIFF
--- a/pkg/cluster/service.go
+++ b/pkg/cluster/service.go
@@ -104,9 +104,10 @@ func (cluster *Cluster) vipService(ctxArp, ctxDNS context.Context, c *kubevip.Co
 		//ctxArp, cancelArp = context.WithCancel(context.Background())
 
 		ipString := cluster.Network.IP()
+		isIPv6 := vip.IsIPv6(ipString)
 
 		var ndp *vip.NdpResponder
-		if vip.IsIPv6(ipString) {
+		if isIPv6 {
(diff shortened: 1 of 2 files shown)


tylerschultz:
We reviewed our PR and realized there's a block of duplicate service related code that ensures the VIP and sends gratuitous updates. We realized that the duplicate code is also prone to the same `dadfailed` errors. We've extracted the common code to a function that both code paths now use.

We've manually tested the change for the control plane VIP code path, and see it has the desired effect.

We've not manually tested the service related code path, we're unsure how to do so. The service-e2e's look to have substantial tests. This PR relies on those tests.

We're done making updates for this PR, and we now think this PR is ready for review. Sorry about the spam.

thebsdbox (approved):

thebsdbox:
`/lgtm`

PR #540, opened 2023-04-14
Adds nftables as an option, and fixes a panic

thebsdbox:
Fixes a panic and ensures rules are cleaned when using nftables.


thebsdbox:
Merging this for now.. the linting issues are in old code that will be removed in the future.

Issue #497, opened 2022-12-27
Control  plane balancing broken with dual stack


xhejtman:
**Describe the bug**
If nodes are running in dual stack configuration, i.e., having both ipv4 and ipv6 addresses, `lb_enable` cannot be used as it balances ipv4 to both ipv4 and ipv6 addresses in wrong way:
```
ipvsadm -L -n
IP Virtual Server version 1.2.1 (size=4096)
Prot LocalAddress:Port Scheduler Flags
  -> RemoteAddress:Port           Forward Weight ActiveConn InActConn
TCP  10.16.62.54:6443 rr
  -> 10.16.62.30:6443             Local   1      0          0         
  -> 32.1.7.24:6443               Local   1      0          0  
```

that ip `32.1.7.24` is actually begining of ipv6 address `2001:718:`. The node has local `10.16.62.30` ipv4 address and also public ipv6 prefix `2001:718:`.

**Environment (please complete the following information):**
 - OS/Distro: Ubuntu 22.04
 - Kubernetes Version: 1.24.7
 - Kube-vip Version: 0.5.7

thebsdbox:
This has been implemented.

PR #554, opened 2023-04-30
Use correct address family for ipvs destinations

sykesm:
When IPv6 addresses are present in node.Status.Addresses, the IPVSLoadBalancer parses the addresses and attempts to add them as an IPv4 destinations. This results in useless entries in the service table that only hold the last 32 bits of address.

This change does two things:
- uses the appropriate address family when creating the service
- ignores backend addresses that use a different address family than the service

This avoids the immediate problem of using the wrong address family for IPv6 destinations and the useless entries in the service table.

#497 appears to report this problem


thebsdbox:
This should be fine to merge, the errors are on deprecated code that I'm going to rip out anyway.

thebsdbox (approved):
`/lgtm`

Issue #479, opened 2022-11-09
Restricts user access to ports other than the load balancer service port through the VIP


yaocw2020:
**Is your feature request related to a problem? Please describe.**
It's not secure to allow users access to ports other than the load balancer service port through the VIP, so we have to restrict it.

**Describe the solution you'd like**
Add some iptables rules to deny connections to ports other than the load balancer service port.

**Describe alternatives you've considered**

**Additional context**

thebsdbox:
Could easily do this with either iptables or eBPF..

PR #560, opened 2023-05-07
Set iptables rules to restrict user access to ports other than the load balancer service port through the VIP

yaocw2020:
Set IPTables rules to Restricts user access to ports other than the load balancer service port through the VIP. Moreover, considering DHCP and any other protocol port requirement, those ports should be reachable.

By the way, we also modify the watch_services.go to support changing service load balancer IP.

IPTables Rules example:
```
-A INPUT -d 172.19.104.147/32 -p tcp -m tcp --dport 80 -m comment --comment "default/primary kube-vip load balancer IP" -j ACCEPT
-A INPUT -d 172.19.104.147/32 -p udp -m udp --dport 68 -m comment --comment "default/primary kube-vip load balancer IP" -j ACCEPT
-A INPUT -d 172.19.104.147/32 -m comment --comment "default/primary kube-vip load balancer IP" -j DROP
```

Related issue: https://github.com/kube-vip/kube-vip/issues/479

### DIFF
--- a/pkg/vip/address.go
+++ b/pkg/vip/address.go
@@ -1,15 +1,24 @@
 package vip
 
 import (
+	"fmt"
+	"os"
+	"strconv"
+	"strings"
 	"sync"
 
 	"github.com/pkg/errors"
+	log "github.com/sirupsen/logrus"
 	"github.com/vishvananda/netlink"
 	"golang.org/x/sys/unix"
+	v1 "k8s.io/api/core/v1"
+
+	"github.com/kube-vip/kube-vip/pkg/iptables"
 )
 
 const (
 	defaultValidLft = 60
+	iptablesComment = "%s kube-vip load balancer IP"
 )
 
 // Network is an interface that enable managing operations for a given IP
@@ -21,6 +30,7 @@ type Network interface {
 	IsSet() (bool, error)
 	IP() string
 	SetIP(ip string) error
+	SetServicePorts(service *v1.Service)
 	Interface() string
 	IsDADFAIL() bool
 	IsDNS() bool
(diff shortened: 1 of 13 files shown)


yaocw2020:
The CI error is introduced by other PRs. We may need to submit a new PR to solve it.

pkg/vip/address.go
```diff
@@ -139,6 +150,136 @@ func (configurator *network) AddIP() error {
...
+	isPortsRuleExisting := make([]bool, len(configurator.ports))
+
+	// delete rules of ports that are not in the service
+	rules, err := ipt.List(iptables.TableFilter, iptables.ChainInput)
+	if err != nil {
+		return fmt.Errorf("could not list iptables rules: %w", err)
+	}
+	for _, rule := range rules {
+		// only handle rules with kube-vip comment
+		if iptables.GetIPTablesRuleSpecification(rule, "--comment") == comment {
```
futuretea, line 188:
```go
if iptables.GetIPTablesRuleSpecification(rule, "--comment") != comment {
	continue
}
```

thebsdbox:
As in a flag/env on the implementation, as in it's current form it is would be the default behaviour.
Something like `--secureService`?

thebsdbox:
Awesome looks good to go! If there is nothing else, I'll give this PR one last look over tomorrow (Awesome work) 😄

yaocw2020:
> Awesome looks good to go! If there is nothing else, I'll give this PR one last look over tomorrow (Awesome work) 😄

Thanks. If merged, could you help draft a new release? Our product whose new version is close to release needs this new feature.

thebsdbox (approved):
`/LGTM`

PR #562, opened 2023-05-15
Add annotation kube-vip.io/ignore-service-security

yaocw2020:
Skip service security configuration for the service whose annotation kube-vip.io/ignore-service-security is true.

Related issue: https://github.com/kube-vip/kube-vip/issues/479


thebsdbox:
Looks good, can you add this to the `/website` repo so people know about it?

PR #564, opened 2023-05-18
Add BGP password support for Equinix Metal

enkelprifti98:
The Equinix Metal API integration was missing support for BGP session passwords so this PR adds the BGP password parameter.


enkelprifti98:
Signed-off-by: Enkel Prifti <enkelprifti@live.com>

PR #567, opened 2023-05-19
Add support for multiple BGP peers when using Equinix Metal annotations

### LINKED ISSUE #543: Support both peers via Equinix Metal annotations
Addendum to, #521. Add support for both peers. We mistakenly thought that kube-vip could only handle one peer when fixing up the annotations support.

Get both equinix-metal bgp peers and add them to the comma separated list.

enkelprifti98:
Using Equinix Metal annotations was limited to only the first BGP peer. This PR adds support for multiple BGP peer by looking at all bgp peer node annotations.

Fixes #543


cprivitere (approved):
lgtm

PR #575, opened 2023-05-30
Increase client QPS to reduce clientside k8api throttling

megakid:
When using kubevip with serviceElection, I have seen many clientside throttling errors that breach 'default' lease periods on leadership elections.

This will allow the client library to perform more queries and consequently hit it's client side rate limiters less.

See https://github.com/coreos/prometheus-operator/blob/9d68ecf289d711c66bef39d2f83429265abc6986/pkg/k8sutil/k8sutil.go#L79-L97
and https://github.com/kubernetes/ingress-nginx/blob/e4d53786e771cc6bdd55f180674b79f5b692e552/pkg/ingress/controller/launch.go#L252-L259


pkg/k8s/client.go
```diff
@@ -32,7 +32,16 @@ func NewClientset(configPath string, inCluster bool, hostname string) (*kubernet
 
 func restConfig(kubeconfig string, inCluster bool) (*rest.Config, error) {
 	if kubeconfig != "" && !inCluster {
-		return clientcmd.BuildConfigFromFlags("", kubeconfig)
+		cfg, err = clientcmd.BuildConfigFromFlags("", kubeConfig)
```
thebsdbox, line 35: I think `cfg, err = clientcmd.BuildConfigFromFlags("", kubeConfig)` should be `cfg, err := clientcmd.BuildConfigFromFlags("", kubeconfig)`

pkg/k8s/client.go
```diff
@@ -32,7 +32,16 @@ func NewClientset(configPath string, inCluster bool, hostname string) (*kubernet
...
+	} else {
+		cfg, err = rest.InClusterConfig()
```
thebsdbox, line 37: Again this probably needs `:=`

thebsdbox (changes requested):
A few fixes, as in it's current state this won't build.

thebsdbox:
Apologies for the delay I've been on PTO 🏖️

thebsdbox:
There are still some compilation issues with this?

thebsdbox:
Can we get these small issues fixed, I'd love to merge this shortly.

thebsdbox:
This still has a compilation problem, but I'll merge it and fix it (just so we can get it in).

PR #576, opened 2023-05-31
Update go dependencies

mrueg:
Make sure we use recent versions.


thebsdbox:
I have no idea why the linter keeps dying :-( .. what happens if you run the make check command locally?

mrueg:
fwiw, there's a breaking change: https://github.com/cloudflare/ipvs/commit/4d94efc6ff9d55895734dc2e17ee1411eb04891f which is using net/netip now, I've not bumped the dep because of this.

mrueg:
@thebsdbox this should be ready to go now.

go.mod
```diff
@@ -1,33 +1,33 @@
 module github.com/kube-vip/kube-vip
 
-go 1.18
+go 1.19
```
lubronzhan, line 3: Should it be 1.20 since you updated dockerfile to use 1.20 as base image

mrueg: not necessarily, as kube-vip would still build with 1.19

PR #578, opened 2023-06-19
enable leaderelection for bgp in CP mode

marc-cerebras:
Hi kube-vip, in order to implement a BGP based VIP with leader election, I had to make this tweak.

I have tested this on a test cluster in control plane however a notable deficiency in my testing is that I have not tried this change in services mode. It does not look like there is an automated test for BGP so I could use some advice on how to proceed with this change- Thank you!

### DIFF
--- a/pkg/cluster/clusterLeaderElection.go
+++ b/pkg/cluster/clusterLeaderElection.go
@@ -167,7 +167,7 @@ func (cluster *Cluster) StartCluster(c *kubevip.Config, sm *Manager, bgpServer *
 		}
 	}
 
-	if c.EnableBGP {
+	if c.EnableBGP && bgpServer == nil {
 		// Lets start BGP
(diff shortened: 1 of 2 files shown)


thebsdbox (approved):

marc-cerebras:
> I'm intrigued about the use case

Sure - one simple reason is for predictability of which node is receiving traffic in situations where there's multiple ToR switches where our controlplane is on multiple racks, we want to have this configuration option.

The other reason is for a non-controlplane use-case where we're launching kv in control-plane mode - we don't want clients to get connection resets in the case that one node hosting a kv follower goes down which triggers routers to reconfigure, resetting connections.

> (either way happy to merge)!

👍

marc-cerebras:
@yastij any objections or suggestions prior to merge?

thebsdbox:
Looks good to me!

PR #579, opened 2023-06-20
Fix makefile default target

runsisi:


runsisi:
i think `.DEFAULT_GOAL` is a special variable, it should be defined using `:=` operator [^1] ?

[^1]: https://www.gnu.org/software/make/manual/html_node/Special-Variables.html

found the issue when building w/o target specified:

```bash
❯ make
go mod tidy
go: downloading k8s.io/api v0.26.2
...
go: downloading github.com/google/pprof v0.0.0-20210407192527-94a9f03dee38
test -z ""
test -z  || echo "[WARN] Fix formatting issues with 'make fmt'"
golangci-lint run
make: golangci-lint: No such file or directory
make: *** [Makefile:96: check] Error 127
```

PR #585, opened 2023-07-12
Allow specifying annotations for created Lease resources

timosluis:
It would be nice to be able to watch for changes for certain Lease resources, and not others. This PR adds the ability to add annotations to the Lease resources being created by kube-vip, allowing for easy filtering of the leases.


thebsdbox:
Very cool, thanks!

PR #587, opened 2023-07-21
Update the DHCP workflows

rikatz:
Debugging some DHCP issues here, I've found the following improvements:

* DHCP library does not need to be replaced anymore
* The request/release/renewal could be simplified to rely only on the nclient4 and some DHCP modifiers.

For example, when releasing an IP, previously it was forcing the usage of unicast client. But, the [ReleaseFromAck](https://github.com/insomniacslk/dhcp/blob/master/dhcpv4/dhcpv4.go#L295) already sets a false broadcast request.

The same applies to rebind/initReboot, which could be simplified to a new discovery request (multicast) but specifying their IP address.

There is a follow-up coming on the next PR, that we need to correctly parse if errors of `nclient4.ErrNak` (today they are also not parsed, just returned). 

In this case, the client should go into requesting a new IP without trying to rebind it, because the server may be communicating that the IP is not available anymore (this can happen on renew, rebind, initReboot).

This means, that a new IP address can be allocated, so the syncServices reconciliation should be triggered to reflect the new allocated IP.

I will come late today (or next week) with a fix for this one as well :)


rikatz:
found  problem on renew, fixing it  :) please don't merge it yet

rikatz:
oook I think this is fine now :)

thebsdbox:
Ooo, just one small issue @rikatz and we should be good to merge 👍

rikatz:
buuuuh got by linters... trying again (sorry for the delay, day was a mess)

rikatz:
I've run the tests locally and things passed now

thebsdbox:
Tests are running!

thebsdbox:
💥 boom

mrueg:
Is https://github.com/harvester/dhcp/commit/28b38eafefe3fcd4d9ce0d3fd605ae54070489e9 / https://github.com/insomniacslk/dhcp/pull/458 not needed anymore? I assume that is why the fork was used in the beginning?

PR #595, opened 2023-08-14
Updating version in makefile

jkossis:
It looks like when v.0.6.1 was cut, the makefile wasn't updated to reflect the new version. This affects manifest generation (not too sure the implications beyond this).


thebsdbox:
Going to merge this and re-release 👍

PR #606, opened 2023-09-04
Add environment variable for routing table type

timosluis:
We would like to be able to bind the virtual IP addresses for outgoing connections, which only seems to work if we add our vip subnet to the local routing table with type local. This pull request adds the option to specify which type should be used when adding entries to the routing table.


thebsdbox:
Is this ready to merge?

thebsdbox:
There are a few complaints from the CI, can these be addressed. I'd love to merge this in!

thebsdbox:
Going to merge and fix locally.

PR #607, opened 2023-09-04
Added create rbac settings section for kind deployment

hellt:
Hey @thebsdbox, all,
I have been trying kube-vip for the first time. Started with `kind` and realized that the usage instructions (if followed as-is) doesn't quite get the LB in a working state.

The RBAC settings were missing and hence the daemonset was never reaching out to running state.

Sorry if that is obvious step, but I would have a better user journey if it was mentioned, hence this PR.

Cheers.


thebsdbox:
Ah! 

Apologies, the documentation all lives in the website repository! `/docs` is going to be removed from the main repo on the next release.

Issue #611, opened 2023-09-08
Egress IP for usage with CiliumEgressGatewayPolicy (add/remove node labels alongside with virtual ip)


fibbs:
I am trying to get outbound connections from a cluster to go out with a defined, dedicated IP address which is not the node ip of one of the cluster nodes. My cluster uses Cilium as CNI with kube-proxy-replacement activated.
I would like to use CiliumEgressGatewayPolicy CRD objects to achieve this. In the definition of a [CiliumEgressGatewayPolicy](https://docs.cilium.io/en/stable/network/egress-gateway/) object you have to specifically define which node the connection has to be routed outbound (by label selector) and optionally you can additionally define which of the ip addresses bound on an interface of that node should be used. There is no "automatic mode" to "discover" on which node a specific IP might be bound and just use this, as the now deprecated CiliumEgressNatPolicy CRD definitions seemed to allow. Also, cilium does not manage adding, removing and migrating IP addresses from one node to the other in case of a node goes down etc. This by its entirety is an "enterprise feature".

Now, in order to achieve a good working and also fault proof way, I came along kube-vip as a possible solution to manage the virtual IP address for me. The fact that kube-vip actually really adds the IP to the interface, other than MetalLB, comes in very handy here. I also noticed that there already is an "Egress" mode in kube-vip but that uses iptables to do the actual routing. I would prefer using kube-vip just for managing the IP and Cilium for the routing part, as described above. For this, the only thing really needed would be an additional functionality inside kube-vip to manage a node label (for example, `kube-vip.io/has-ip-address=<virtual-ip>`) alongside with the virtual IP assignment. If the VIP moves over to another kubernetes node, the according label would have to be removed from the old and set on the new node. With this, the CiliumEgressGatewayPolicy would route the traffic just over the correct node and ip address.

Additional to that would be benefitial the ability to manage more than just one IP address at a time with kube-vip, which is something which I really wonder why it is not something more requested by users. Are there plans to extend kube-vip in this manner?

thebsdbox:
I believe this issue has now been fixed, please re-open if this isn't the case.

PR #617, opened 2023-09-19
ARP: Add node labeling for ARP mode DaemonSet deployment.

mjtrangoni:
This feature is related to #611, to configure the _**CiliumEgressGatewayPolicy**_ *EgressIP* correctly.

As this is not a very common usage for this project, there is a default disabled configuration flag added, `enable_node_labeling`.

The *ClusterRole* also needed patch rights to adjust the node labels.

Let me know what you think about the change, and where is the best place to document this feature.

Thank you!


sa-ChristianAnton:
done some extensive tests with Cilium and this patched version of kube-vip. It's working perfectly for me.

thebsdbox:
Excellent, we love Cilium! I'll cut a new release soon!

aevenot:
Hello,
I'm trying to make the enable_node_labeling features work but it's not.
I have added the env enable_node_labeling=true in my daemonset but the label doesn't appear. I have deployed the latest version and I have also patched the clusterrole.

What do I have to do to make it work?

PR #624, opened 2023-10-03
Enable unit and e2e tests in CI

g-gaston:
Maybe there was a reason to not run unit tests but I thought it could be useful to run what we already have and it makes it easier for folks to add more as they write new code.

A couple extra things:
* Fixed 2 broken tests
* Separated unit tests, e2e tests, linter and scans in different jobs so they run concurrently.

### DIFF
--- a/pkg/manager/watcher_test.go
+++ b/pkg/manager/watcher_test.go
@@ -77,8 +77,8 @@ func TestParseNewBgpAnnotations(t *testing.T) {
 	node.Annotations = map[string]string{
 		"bgp/bgp-peers-0-node-asn": "65000",
 		"bgp/bgp-peers-0-peer-asn": "64000",
-		"bgp/bgp-peers-0-peer-ip":  "10.0.0.254",
-		"bgp/bgp-peers-0-src-ip":   "10.0.0.1,10.0.0.2,10.0.0.3",
+		"bgp/bgp-peers-0-peer-ip":  "10.0.0.1,10.0.0.2,10.0.0.3",
+		"bgp/bgp-peers-0-src-ip":   "10.0.0.254",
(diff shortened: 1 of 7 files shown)


pkg/manager/watcher_test.go
```diff
@@ -77,8 +77,8 @@ func TestParseNewBgpAnnotations(t *testing.T) {
 	node.Annotations = map[string]string{
 		"bgp/bgp-peers-0-node-asn": "65000",
 		"bgp/bgp-peers-0-peer-asn": "64000",
-		"bgp/bgp-peers-0-peer-ip":  "10.0.0.254",
-		"bgp/bgp-peers-0-src-ip":   "10.0.0.1,10.0.0.2,10.0.0.3",
+		"bgp/bgp-peers-0-peer-ip":  "10.0.0.1,10.0.0.2,10.0.0.3",
+		"bgp/bgp-peers-0-src-ip":   "10.0.0.254",
```
g-gaston, line 81: If I'm not mistaken, these two expected values were swapped, the code that generates the peers looks correct, it was just the wrong expectation

testing/e2e/e2e_test.go
```diff
@@ -98,7 +98,7 @@ var _ = Describe("kube-vip broadcast neighbor", func() {
 
 			manifestPath := filepath.Join(tempDirPath, "kube-vip-ipv4.yaml")
 
-			for i := 0; i < 2; i++ {
+			for i := 0; i < 3; i++ {
```
g-gaston, line 101: Maybe a bit slower but using 3 nodes removes the flakiness since etcd won't lose quorum by deleting only one node.

thebsdbox:
Looks good and all tests have passed! (Thanks)

PR #629, opened 2023-10-17
Fix etcd e2e tests in GitHub actions

g-gaston:
In order to run the kubelet in kind containers on the vms provided by github, the kubelet needed some extra configuration.

I also added some setup for the kind network so the the etcd tests don't depend on being run together with the other e2e tests.


thebsdbox (approved):
Fantastic!

Thankyou

Issue #614, opened 2023-09-15
Is it possible to set a custom hostname on DHCP lease request ?


shkuviak:
**Is your feature request related to a problem? Please describe.**
I deployed kube-vip in my **microk8s** cluster using **DHCP** to assign IP provided by DHCP server to services. 

It's working great. 

Now I would like to access those IP via a hostname.

DHCP offers the possibility to send a client hostname when requesting an IP. Then, DHCP server registers the hostname in DNS server (**Active Directory** Infrastructure) to access the machine by domain name: `my-hostname.my.custom.domain`

I cannot find any way to specify a hostname to the DHCP client. Is this feature implemented ?

**Describe the solution you'd like**
I would like to specify a hostname for a specific *LoadBalancer* service to be able to access it via domain name resolution provided by my domain controller.

Example: while exposing a deployment:
```
kube expose deployment my-deployment --port 80 --target-port 8080 --type=LoadBalancer --name=my-service --load-balancer-ip=0.0.0.0 --hostname=my-deployment-hostname.my.custom.domain
```

Then, i'll be able to access the deployment via `curl my-deployment-hostname.my.custom.domain`

**Additional context**
my kube-vip deployment manifest:
```yaml
apiVersion: apps/v1
kind: DaemonSet
metadata:
  creationTimestamp: null
  labels:
    app.kubernetes.io/name: kube-vip-ds
    app.kubernetes.io/version: v0.6.2
  name: kube-vip-ds
  namespace: kube-system
spec:
  selector:
    matchLabels:
      app.kubernetes.io/name: kube-vip-ds
  template:
    metadata:
      creationTimestamp: null
      labels:
        app.kubernetes.io/name: kube-vip-ds
        app.kubernetes.io/version: v0.6.2
    spec:
      containers:
      - args:
        - manager
        env:
        - name: vip_arp
          value: "true"
        - name: port
          value: "6443"
        - name: vip_interface
          value: enp0s31f6
        - name: vip_cidr
          value: "32"
        - name: svc_enable
          value: "true"
        - name: vip_leaderelection
          value: "true"
        - name: vip_leasename
          value: plndr-cp-lock
        - name: vip_leaseduration
          value: "5"
        - name: vip_renewdeadline
          value: "3"
        - name: vip_retryperiod
          value: "1"
        - name: address
          value: "my-custom-hostname.my.custom.domain" # -> Ignore it, I misunderstood this feature at the begining
        - name: prometheus_server
          value: :2112
[...]

PR #664, opened 2023-11-16
Add annotation to specify DHCP lease hostname

shkuviak:
Following my previous issue #614 , i made little changes to be able to specify hostname sent to dhcp server when requesting lease using an annotation as the following:
```yaml
apiVersion: v1
kind: Service
metadata:
  name: nginx-dhcp
  annotations:
    kube-vip.io/loadbalancerHostname: mydhcp-test
spec:
  loadBalancerIP: 0.0.0.0
  ports:
  - name: http
    port: 80
    protocol: TCP
    targetPort: 80
  selector:
    app: hello-world
  type: LoadBalancer
```

This makes DHCP aware of the hostname and makes the service accessible though local hostname resolution:
```
curl mydhcp-test.lan
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,7 @@ TARGET := kube-vip
 .DEFAULT_GOAL := $(TARGET)
 
 # These will be provided to the target
-VERSION := v0.6.3
+VERSION := v0.6.4
 
 BUILD := `git rev-parse HEAD`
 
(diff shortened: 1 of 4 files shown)


thebsdbox:
Nice!

thebsdbox (approved):
`/LGTM`

This is great, thanks!

thebsdbox (approved):

PR #686, opened 2023-12-15
Added first code-path for leaderelection in tableMode

Cellebyte:
* This is my first try to implement a kube-vip without leaderElection.
* Lets see what is needed.


thebsdbox:
So if I'm reading this right.. No leader election so all pods (daemonset? or tainted nodes or something like that) will all react when they see a service, and add to their routing table?

Ideally it's also best to separate out bits of work into separate PRs to keep things a little cleaner (the services comparison code can stay in this PR however 👍 )

Cellebyte:
Close this Pull Request in favor of #720

PR #695, opened 2023-12-21
Add config options for BGP timers

marc-cerebras:
In certain scenarios, it's easier to change the BGP timer settings from the kube-vip side rather than configuring the routers' peer-group timer settings for kube-vip. This is desirable if the router defaults are higher than what we'd like to configure on the kube-vip side for the purpose of quickly reacting to a node down or network partition.

This change adds two new env vars for configuring these settings during the start of peering.

I was able to validate on my local deployment, that the switch side was picking up the modified timer settings.

```
kube-vip manifest yaml envvars
...
    - name: bgp_hold_time
      value: "9"
    - name: bgp_keepalive_interval
      value: "3"
...

Router `show ip bgp neighbors`
...
BGP neighbor is x.x.x.x, remote AS xxxx, internal link
  BGP version 4, remote router ID x.x.x.x, VRF default
  Inherits configuration from and member of peer-group xxx
  Negotiated BGP version 4
  Member of update group 2
  Last read 00:00:01, last write 00:00:01
  Hold time is 9, keepalive interval is 3 seconds

...where the default is 120 40
```

### DIFF
--- a/docs/flags/index.md
+++ b/docs/flags/index.md
@@ -92,6 +92,8 @@ More environment variables can be read through the `pkg/kubevip/config_envvar.go
 |                     | `bgp_multihop`         | Enables eBGP MultiHop                                       | Enable multiHop with a single BGP Peer                                          |
 |                     | `bgp_sourceif`         | Source Interface                                            | Determines which interface BGP should peer _from_                               |
 |                     | `bgp_sourceip`         | Source Address                                              | Determines which IP address BGP should peer _from_                              |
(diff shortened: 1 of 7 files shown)


thebsdbox:
Woah, this is excellent! Fantastic and thank you!

marc-cerebras:
Thanks! I have tested as a static pod and that is working. Now testing as a daemonset deploy and the envvar doesn't seem to pick up. Looking into where it's getting lost. Will close and re-open once I find the cause.

marc-cerebras:
Okay figured it out. The base BGP config is overridden by the annotation parsed config. To avoid this, I passed the base BGP config into the annotation parsing function where it will overlay the envvar driven configs.

PR #724, opened 2024-01-13
fix #723 and allow short hostnames as well

### LINKED ISSUE #723: [Bug] shortName not correctly discovered
https://github.com/kube-vip/kube-vip/blob/3c0e5e327e21e26f1f131c90e0908dc01662ae37/pkg/manager/watch_endpoints.go#L241

This should be `>=` so short host names also work.

As playground currently does not work here the example code.
```
// You can edit this code!
// Click here and start typing.
package main

import (
	"fmt"
	"strings"
)

func getShortname(hostname string) (string, error) {
	if len(hostname) == 0 {
		return "", fmt.Errorf("unable to find shortname from %s as len is 0 %d", hostname, len(hostname))
	}
	hostParts := strings.Split(hostname, ".")
	if len(hostParts) >= 1 {
		return hostParts[0], nil
	}
	return "", fmt.Errorf("unable to find shortname from %s, as hostParts are %v", hostname, hostParts)
}

func main() {
	fmt.Println(getShortname("wuerfelchen-w-1"))
}
```

Cellebyte:
* This fixes #723 
* It ensures that short hostnames are discovered correctly.

### DIFF
--- a/pkg/manager/watch_endpoints.go
+++ b/pkg/manager/watch_endpoints.go
@@ -238,7 +238,7 @@ func getShortname(hostname string) (string, error) {
 		return "", fmt.Errorf("unable to find shortname from %s", hostname)
 	}
 	hostParts := strings.Split(hostname, ".")
-	if len(hostParts) > 1 {
+	if len(hostParts) >= 1 {
 		return hostParts[0], nil
 	}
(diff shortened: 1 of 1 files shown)


thebsdbox:
Flakey CI test, happy to merge this one! Thanks

PR #731, opened 2024-01-16
ExternalTrafficPolicy Cluster was not working as expected

Cellebyte:
* Looks like the `ExternalTrafficPolicy: Cluster` is not correctly behaving in kube-vip.

### DIFF
--- a/pkg/manager/manager_table.go
+++ b/pkg/manager/manager_table.go
@@ -107,7 +107,6 @@ func (sm *Manager) startTableMode() error {
 			},
 		})
 	} else {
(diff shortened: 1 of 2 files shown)


thebsdbox:
Thanks as always for some fantastic work, I’d like to get this into the next release. Anyway I can help?

Cellebyte:
My colleague @p-strusiewiczsurmacki-mobica will take a look at this code part on Monday and probably will ensure that the code is working as desired. Let’s see. He also worked on another feature to implement the same behavior in bgp for the non leader election. 
I would propose waiting on his Pull Requests. I just made this code here to hotfix a broken load balancing setup in my association.

PR #746, opened 2024-02-04
fix: Using log instead of fmt.print

ii2day:
When I was looking at the source code, I noticed that print was being used here, and I felt it would be better to change it to log output.

### DIFF
--- a/cmd/kube-vip-kubeadm.go
+++ b/cmd/kube-vip-kubeadm.go
@@ -1,7 +1,6 @@
 package cmd
 
 import (
-	"fmt"
 	"os"
 
 	"github.com/kube-vip/kube-vip/pkg/kubevip"
@@ -54,7 +53,7 @@ var kubeKubeadmInit = &cobra.Command{
 		}
 		cfg := kubevip.GeneratePodManifestFromConfig(&initConfig, Release.Version, inCluster)
 
-		fmt.Println(cfg)
+		log.Info(cfg)
 	},
 }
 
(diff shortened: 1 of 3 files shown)


lubronzhan:
Good find. Would you mind also finding and fixing other places using fmt to print log?

On a brief search I find blow places as well

https://github.com/kube-vip/kube-vip/blob/34a4352faed124d7d8a0f6af5f352b77146974b3/cmd/kube-vip.go#L172
https://github.com/kube-vip/kube-vip/blob/34a4352faed124d7d8a0f6af5f352b77146974b3/cmd/kube-vip-manifests.go#L57
https://github.com/kube-vip/kube-vip/blob/34a4352faed124d7d8a0f6af5f352b77146974b3/cmd/kube-vip-kubeadm.go#L57

Thank you

ii2day:
> Good find. Would you mind also finding and fixing other places using fmt to print log?
> 
> On a brief search I find blow places as well
> 
> https://github.com/kube-vip/kube-vip/blob/34a4352faed124d7d8a0f6af5f352b77146974b3/cmd/kube-vip.go#L172
> 
> 
> https://github.com/kube-vip/kube-vip/blob/34a4352faed124d7d8a0f6af5f352b77146974b3/cmd/kube-vip-manifests.go#L57
> 
> 
> https://github.com/kube-vip/kube-vip/blob/34a4352faed124d7d8a0f6af5f352b77146974b3/cmd/kube-vip-kubeadm.go#L57
> 
> Thank you

Okay, I've searched globally and changed the place for fmt.print to log

lubronzhan:
Thanks. Dan should have the permission to start the test

/ok-to-test

thebsdbox:
Nice, once this goes through we should be pretty much ready to do the new release

ii2day:
> Nice, once this goes through we should be pretty much ready to do the new release

🎉 By the way, I've noticed that kube-vip's logging doesn't output the number of log lines, which is annoying for people who aren't familiar with kube-vip's code. Is it considered to improve this log output?

PR #761, opened 2024-02-13
Added common endpoint provider interface and fixed route deletion iss…

### LINKED ISSUE #754: [Bug] When kube-vip reconciles in Table/BGP Mode it should reconcile gracefully.
**Describe the bug**
The Bug is not a functional bug, but rather a visual/.

**To Reproduce**
1. Deploy kube-vip without `svc_election` and without `leader_election` in table or BGP mode.
2. Configure two `Services` which have a selector on the same backend pods.
4. Assign both Services the same Loadbalancer IP

**Expected behavior**
Either fail silently when the IP is already there or precompute what needs to be configured should be implemented.

 - OS/Distro: Ubuntu 20.04
 - Kubernetes Version: 1.27.3
 - Kube-vip Version: 0.7.0

**Additional context**
```
time="2024-02-08T09:20:07Z" level=info msg="[endpoint] watching for service [management-nginx-ingress-controller] in namespace [ingress-system]"
time="2024-02-08T09:20:07Z" level=info msg="(svcs) adding VIP [[10.27.195.225]] for [ingress-system/management-nginx-ingress-controller]"
time="2024-02-08T09:20:07Z" level=info msg="[service] synchronised in 0ms"
time="2024-02-08T09:20:07Z" level=info msg="[endpoint] watching for service [management-nginx-ingress-controller-legacy] in namespace [ingress-system]"
time="2024-02-08T09:20:07Z" level=info msg="(svcs) adding VIP [[10.27.195.225]] for [ingress-system/management-nginx-ingress-controller-legacy]"
time="2024-02-08T09:22:47Z" level=info msg="[endpoint]  added route: 10.0.0.0, service: ingress-system/management-nginx-ingress-controller, interface: lo, table: 198"
time="2024-02-08T09:22:47Z" level=error msg="[endpoint] error adding route: file exists\n"
```

p-strusiewiczsurmacki-mobica:
…ues in routing table mode

This PR introduces common interface for `endpoints` and `endpointslices`. This way endpoints and endpointslices watchers do not have to be maintained separately (as both do mostly the same, but using different Kubernetes resources).

Also fixed some issues with routing table cleanup when running in `no-leader-election` mode (e.g. routes were not properly deleted from the node when `local` policy was used if pods were deleted when kube-vip was not running, and then was deployed again).

Also this should fix #754 - already exist error will not be logged if route is already there.

I would recommend to create new tag `v0.7.1` image with those changes if possible.

### DIFF
--- a/pkg/manager/services.go
+++ b/pkg/manager/services.go
@@ -257,11 +257,14 @@ func (sm *Manager) deleteService(uid string) error {
 			}
 		}
 		// TODO: Implement dual-stack loadbalancer support if BGP is enabled
-		if serviceInstance.vipConfigs[0].EnableBGP && ((sm.config.EnableLeaderElection || sm.config.EnableServicesElection) || configuredLocalRoutes[uid]) {
-			cidrVip := fmt.Sprintf("%s/%s", serviceInstance.vipConfigs[0].VIP, serviceInstance.vipConfigs[0].VIPCIDR)
-			err := sm.bgpServer.DelHost(cidrVip)
-			if err != nil {
-				return fmt.Errorf("error deleting BGP host: %v", err)
+		for i := range serviceInstance.vipConfigs {
+			if serviceInstance.vipConfigs[i].EnableBGP {
+				cidrVip := fmt.Sprintf("%s/%s", serviceInstance.vipConfigs[i].VIP, serviceInstance.vipConfigs[i].VIPCIDR)
+				err := sm.bgpServer.DelHost(cidrVip)
+				if err != nil {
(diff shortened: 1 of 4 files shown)


MaxRink:
>I would recommend to retag v0.7.0 image with those changes if possible.
IMO thats not a good idea, as that will lead to people having no clue if they are running a broken or fixed version. IMO the right thing to do is tag 0.7.1 and adding a notice on 0.7.0

thebsdbox (approved):
Thanks, this is a lot cleaner!

Issue #762, opened 2024-02-16
Old routes are not removed if service is modified when kube-vip is down


p-strusiewiczsurmacki-mobica:
**Describe the bug**
When kube-vip is deployed in routing table mode and service is created, rute is created (in table `198` by default). If kube-vip will be down and service will be recreated with another loadbalancer IP address, old route will stay in the routing table.

**To Reproduce**
1. Deploy kube-vip in routing table mode (both non-leader-election and service-leader-election tested).
2. Deploy service with some LB IP (e.g. 172.190.0.10**5**)
3. Delete kube-vip deployment.
4. Delete the service.
5. Recreate the service with different LB IP (e.g. 172.190.0.10**6**).
6. See that routing table has still both IP addresses present:

```
$ docker exec -it kind-worker2 ip route show table 198
172.190.0.105 dev lo 
172.190.0.106 dev lo
```

**Expected behavior**
kube-vip should remove old routes if possible.

**Environment (please complete the following information):**
 - OS/Distro: Ubuntu 22.04
 - Kubernetes Version: v1.27.3
 - Kube-vip Version:v0.7.0

**`Kube-vip.yaml`:**
```
apiVersion: apps/v1
kind: DaemonSet
metadata:
  annotations:
    deprecated.daemonset.template.generation: "4"
  labels:
    app.kubernetes.io/name: kube-vip-svc
    app.kubernetes.io/version: v0.7.0
  name: kube-vip-svc
  namespace: kube-system
spec:
  revisionHistoryLimit: 10
  selector:
    matchLabels:
      app.kubernetes.io/name: kube-vip-svc
  template:
    metadata:
      creationTimestamp: null
      labels:
        app.kubernetes.io/name: kube-vip-svc
        app.kubernetes.io/version: v0.7.0
    spec:
      affinity:
        nodeAffinity:
          requiredDuringSchedulingIgnoredDuringExecution:
            nodeSelectorTerms:
            - matchExpressions:
              - key: node-role.kubernetes.io/control-plane
                operator: DoesNotExist
            - matchExpressions:
              - key: node-role.kubernetes.io/worker
                operator: Exists
      containers:
      - args:
        - manager
        env:
        - name: vip_loglevel
          value: "5"
        - name: vip_arp
          value: "false"
        - name: port
          value: "6443"
        - name: vip_cidr
          value: "32"
        - name: svc_enable
          value: "true"
        - name: svc_election
          value: "false"
        - name: vip_leaderelection
[...]

PR #763, opened 2024-02-16
Added possibility to remove redundant routes

p-strusiewiczsurmacki-mobica:
This PR introduces a possibility to clean reduntant routes created in routing table mode as described in #762 

Created routes will be tagged using `protocol` as explained in [ip-route man](https://man7.org/linux/man-pages/man8/ip-route.8.html) (value `protocol RTPROTO`) which is configurable using `routingProtocol` flag (defaults to ~~`198`~~ `248`, ~~same as table~~).

On every kube-vip start, if `cleanRoutingTable` flag is set `true`, _after 10s of initial delay_, all routes from the configured table that have protocol set to the same value as `routingProtocol`, will be get from node and checked against watched services. If no service will have LB IP same as the route `dst` field, the route will be deleted.

### DIFF
--- a/cmd/kube-vip.go
+++ b/cmd/kube-vip.go
@@ -127,6 +127,7 @@ func init() {
 	// Routing Table flags
 	kubeVipCmd.PersistentFlags().IntVar(&initConfig.RoutingTableID, "tableID", 198, "The routing table used for all table entries")
 	kubeVipCmd.PersistentFlags().IntVar(&initConfig.RoutingTableType, "tableType", 0, "The type of route that will be added to the routing table")
+	kubeVipCmd.PersistentFlags().IntVar(&initConfig.RoutingProtocol, "routingProtocol", 198, "The routing protocol value used to create routes")
 
 	// Behaviour flags
 	kubeVipCmd.PersistentFlags().BoolVar(&initConfig.EnableControlPlane, "controlplane", false, "Enable HA for control plane")
--- a/cmd/kube-vip.go
+++ b/cmd/kube-vip.go
@@ -127,7 +127,8 @@ func init() {
 	// Routing Table flags
 	kubeVipCmd.PersistentFlags().IntVar(&initConfig.RoutingTableID, "tableID", 198, "The routing table used for all table entries")
 	kubeVipCmd.PersistentFlags().IntVar(&initConfig.RoutingTableType, "tableType", 0, "The type of route that will be added to the routing table")
-	kubeVipCmd.PersistentFlags().IntVar(&initConfig.RoutingProtocol, "routingProtocol", 198, "The routing protocol value used to create routes")
+	kubeVipCmd.PersistentFlags().IntVar(&initConfig.RoutingProtocol, "routingProtocol", 248, "The routing protocol value used to create routes")
+	kubeVipCmd.PersistentFlags().BoolVar(&initConfig.CleanRoutingTable, "cleanRoutingTable", false, "Clean routing table of redundant routes on start")
 
 	// Behaviour flags
 	kubeVipCmd.PersistentFlags().BoolVar(&initConfig.EnableControlPlane, "controlplane", false, "Enable HA for control plane")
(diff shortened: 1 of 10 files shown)


cmd/kube-vip.go
```diff
@@ -127,6 +127,7 @@ func init() {
 	// Routing Table flags
 	kubeVipCmd.PersistentFlags().IntVar(&initConfig.RoutingTableID, "tableID", 198, "The routing table used for all table entries")
 	kubeVipCmd.PersistentFlags().IntVar(&initConfig.RoutingTableType, "tableType", 0, "The type of route that will be added to the routing table")
+	kubeVipCmd.PersistentFlags().IntVar(&initConfig.RoutingProtocol, "routingProtocol", 198, "The routing protocol value used to create routes")
```
Cellebyte, line 130: please don't use 198 as a default -> https://github.com/FRRouting/frr/blob/master/zebra/rt_netlink.h#L40-L47

kernel protocols. -> https://github.com/torvalds/linux/blob/master/include/uapi/linux/rtnetlink.h#L274

pkg/kubevip/config_envvar.go
```diff
@@ -136,6 +136,9 @@ const (
 	//						 you should say `vip_routingtabletype=2` for RTN_LOCAL
 	vipRoutingTableType = "vip_routingtabletype" //nolint
 
+	// vipRoutingProtocol - defines what value will be used as protcol when creating routes
```
lubronzhan, line 139: nit: protcol -> protocol

pkg/manager/manager_table.go
```diff
@@ -25,7 +28,17 @@ func (sm *Manager) startTableMode() error {
 	// want to step down
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
-	log.Infof("all routing table entries will exist in table [%d]", sm.config.RoutingTableID)
+	log.Infof("all routing table entries will exist in table [%d] with protocol [%d]", sm.config.RoutingTableID, sm.config.RoutingProtocol)
+
+	go func() {
+		// we assume that after 10s all services should be configured so we can delete redundant routes
+		time.Sleep(time.Second * 10)
```
lubronzhan, line 35: Just a thought, since here it make an assumption, is it safer to turn if off by default and allow user to turn it on if needed?

p-strusiewiczsurmacki-mobica: OK, I've added flag `cleanRoutingTable` which, if true will enable this cleaning and defaults to false.

TBH, I don't like this assumption at all, but I think there's no way to tell if all the services were already processed. At least I cannot see it. :(

thebsdbox:
Looks good, thanks for making this an option to turn on. Is this ready to merge?

Cellebyte:
@thebsdbox I think it is.
I am just buidling again a test image. But @p-strusiewiczsurmacki-mobica already confirmed how to reproduce it in #762  and tested with that case. So I assume I will not find any issues with it.

Cellebyte:
@thebsdbox I found some issues when testing it. 

* Route protocol is not added to old routes from previous version which are in the table and match the service. (Upgrade bug)
* I found a weird crash log which I already shared with @p-strusiewiczsurmacki-mobica.
* I also had some issues with the last version I build from main when running it a longer period, that it started to delete the routes indefinitely from nodes for a. dual stack service . For that I would need to investigate and wait for some days to see if it was a one of a kind or a real issue.

Cellebyte:
@p-strusiewiczsurmacki-mobica I have now a new image running and it does not sync the routes at all after your changes adding the protocol to legacy routes.

Cellebyte:
@p-strusiewiczsurmacki-mobica my over night test went smoothly so I propose removing the draft status again from this PR.

p-strusiewiczsurmacki-mobica:
Changed the status and merged commit messages. I believe this should be ready now.

thebsdbox:
Been through the proposed code changes, and seen @lubronzhan's comments. Looks good to me, there is a bug in the egress code that was introduced once I fix that I'll create a new release!

Issue #755, opened 2024-02-09
kube-vip v0.7.0 broken the control plane load balancing


W0n9:
**Describe the bug**
After upgrade to kube-vip v0.7.0 from v0.6.4，kube-vip will be exited with error with enabled control plane load balancing.

**To Reproduce**
Enable `lb_enable` and upgrade to v0.7.0 from v0.6.4

**Expected behavior**
Works without errors.

**Screenshots**

**Environment (please complete the following information):**
 - OS/Distro: Debian 12
 - Kubernetes Version: K3S v1.27.8+k3s2
 - Kube-vip Version: v0.7.0

**`Kube-vip.yaml`:**
If Possible add in your kube-vip manifest (please remove anything that is confidential)
```yaml
apiVersion: apps/v1
kind: DaemonSet
metadata:
  creationTimestamp: null
  labels:
    app.kubernetes.io/name: kube-vip-ds
    app.kubernetes.io/version: v0.7.0
  name: kube-vip-ds
  namespace: kube-system
spec:
  selector:
    matchLabels:
      app.kubernetes.io/name: kube-vip-ds
  template:
    metadata:
      creationTimestamp: null
      labels:
        app.kubernetes.io/name: kube-vip-ds
        app.kubernetes.io/version: v0.7.0
    spec:
      affinity:
        nodeAffinity:
          requiredDuringSchedulingIgnoredDuringExecution:
            nodeSelectorTerms:
            - matchExpressions:
              - key: node-role.kubernetes.io/master
                operator: Exists
            - matchExpressions:
              - key: node-role.kubernetes.io/control-plane
                operator: Exists
      containers:
      - args:
        - manager
        env:
        - name: vip_arp
          value: "true"
        - name: port
          value: "6443"
        - name: vip_interface
          value: enp1s0
        - name: vip_cidr
          value: "32"
        - name: cp_enable
          value: "true"
        - name: cp_namespace
          value: kube-system
        - name: vip_ddns
          value: "false"
        - name: svc_enable
          value: "true"
        - name: svc_leasename
          value: plndr-svcs-lock
        - name: svc_election
          value: "true"
        - name: vip_leaderelection
          value: "true"
        - name: vip_leasename
          value: plndr-cp-lock
        - name: vip_leaseduration
          value: "5"
        - name: vip_renewdeadline
          value: "3"
        - name: vip_retryperiod
          value: "1"
        - name: lb_enable
          value: "true"
        - name: lb_port
          value: "6443"
[...]

Issue #752, opened 2024-02-06
kube-vip 0.7.0 causes nodes to lose connection and interfaces would lose IPs


Michaelpalacce:
I am writing this, with the full understanding that it's very vague. I tried pretty much the whole day trying to debug what is happening and came to the realization that kube-vip was somehow causing node eth0 interfaces to lose connection. 

This lead to a bunch of other issues of course. I don't have the necessary knowledge to figure out what exactly was happening, but outside of connection loss, this was causing entire network slowdowns, and it was also affecting other devices not connected to k3s at all (like my phone's Wi-Fi).

**To Reproduce**
1. Set up a fresh cluster k3s v 1.29.0
2. Get values from [here](https://github.com/Michaelpalacce/HomeLab/blob/master/cluster/homelab/core/kube-vip/helm-release.yaml)
3. Helm install with provided values.

**Expected behavior**
Create a virtual ip that I can use

**Screenshots**
Can't do this now... at this point, I redid my cluster and restored all my backups. I realize this is not helpful, still deciding to post this if anyone else finds this.

**Environment (please complete the following information):**
 - OS/Distro: Ubuntu Server 20.04
 - Kubernetes Version: v.1.28.3, v.1.28.5, v.1.29.0
 - Kube-vip Version: 0.7.0 (works with 0.6.4)

**`Kube-vip.yaml`:**
```yaml
image:
    repository: ghcr.io/kube-vip/kube-vip
    tag: "v0.6.4"

config:
    address: "192.168.1.2"

env:
    vip_interface: ""
    vip_arp: "true"
    lb_enable: "true"
    lb_port: "6443"
    vip_cidr: "32"
    cp_enable: "true"
    svc_enable: "true"
    vip_leaderelection: "false"
```

Michaelpalacce:
This is still observed with both 0.7.1 and 0.7.2, not sure if anything is changed... Essentially when I run these versions the whole cluster loses connection internally and etcd goes haywire trying to elect a new leader constantly

Michaelpalacce:
Should be reopened @thebsdbox

Cellebyte:
@Michaelpalacce you should not run cp_enable and lb enable with the same daemonset.

Michaelpalacce:
So enabling leader election would solve the problem? Will try it out.

@Cellebyte if you can, can you please explain why this is a bad thing?

Michaelpalacce:
This is what happens immediately:

[...]

PR #765, opened 2024-02-19
Fix IPVS service error: netlink receive invalid argument

lou-lan:
After upgrading IPVS to 0.10.1, it is necessary to specify the Netmask.
If it is necessary to parse the netmask from a specific location, I will also update this PR. Comments are welcome.

Fix https://github.com/kube-vip/kube-vip/issues/755
Fix https://github.com/kube-vip/kube-vip/issues/752


thebsdbox:
I think the net mask should be fine for the control plane VIP (thanks)

Issue #771, opened 2024-02-23
Investigate the flaky e2e test


lubronzhan:
**Describe the bug**
Now the e2e test is kind of flaky.  Better to figure out why it's flaky and try fix it or make it happen less often, maybe by tuning the timeout for example

Example:
https://github.com/kube-vip/kube-vip/actions/runs/8012085050/job/21886685171
https://github.com/kube-vip/kube-vip/actions/runs/7992122294/job/21824906442
https://github.com/kube-vip/kube-vip/actions/runs/7990752712/job/21820248635

```
  kube-vip-test801781902-ipv4-control-plane
  STEP: 2024-02-22 23:06:09.464968739 +0000 UTC m=+138.646686161: checking that the Kubernetes control plane nodes are still accessible via the assigned IPv4 VIP with little downtime @ 02/22/24 23:06:09.464
  [FAILED] in [It] - /home/runner/work/kube-vip/kube-vip/testing/e2e/e2e_test.go:412 @ 02/22/24 23:06:39.465
  << Timeline

  [FAILED] Timed out after 30.000s.
  Failed to connect to VIP
  Expected
      <int>: 500
  to equal
      <int>: 200
  In [It] at: /home/runner/work/kube-vip/kube-vip/testing/e2e/e2e_test.go:412 @ 02/22/24 23:06:39.465
------------------------------
```

**To Reproduce**
Create a test PR and make some code changes

**Expected behavior**
The E2E should not be flaky

**Screenshots**

**Environment (please complete the following information):**
 - OS/Distro: [e.g. Ubuntu 1804]
 - Kubernetes Version: [e.g. v.1.18]
 - Kube-vip Version: [e.g. 0.2.3]

**`Kube-vip.yaml`:**
If Possible add in your kube-vip manifest (please remove anything that is confidential)

**Additional context**

lubronzhan:
Yeah make sense. Maybe we can mark those getting 500 status code as success with warning to reduce some noise. And check why those -1 happens

PR #776, opened 2024-03-01
Fix flaky e2e test

lubronzhan:
…ne VIP
Fix https://github.com/kube-vip/kube-vip/issues/771

Assume VIP is routable even if API server returns 500, since that means VIP is routable, it just might be because etcd is glitching.

Another way to fix is to increase the timeout, but I think better not to increase that since 30s is already a long time
considering the default config
```
- name: vip_leaseduration
      value: "5"
    - name: vip_renewdeadline
      value: "3"
    - name: vip_retryperiod
      value: "1"
```

### DIFF
--- a/testing/e2e/e2e_test.go
+++ b/testing/e2e/e2e_test.go
@@ -393,6 +393,7 @@ func createKindCluster(logger log.Logger, config *v1alpha4.Cluster, clusterName 
 	)).To(Succeed())
 }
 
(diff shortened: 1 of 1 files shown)


thebsdbox (approved):
This is fine for me, Obviously `200` would be better as it proves Kubernetes is up and running.. but we're testing the VIP not the API server.

PR #777, opened 2024-03-01
Fixes conntrack deleting wrong connections and cleaning old SNAT rules

thebsdbox:
The contrack process can be too aggressive, so this introduces additional lookups to ensure only the specified rules are deleted. Additionally an extra process will make sure that when a VIP rule (SNAT) is applied that any existing VIP rules are deleted first. 

Hard to check in testing..


thebsdbox:
Merging this, it's a bit experimental but has been proven in production.

Issue #781, opened 2024-03-08
Control plane load balancing on IPv6 does IPVS service error


RomLecat:
Hi,

**Describe the bug**
I have the same error than #755 in IPv6, however the VIP is advertised as expected (ip a s dev eth0 shows the IPv6 VIP on one master node).
IPv4 VIP works without error.

Logs shows:

```
time="2024-03-07T18:53:02Z" level=info msg="IPVS Loadbalancer enabled for 1.2.1"
time="2024-03-07T18:53:02Z" level=info msg="Kube-Vip is watching nodes for control-plane labels"
time="2024-03-07T18:53:02Z" level=info msg="Gratuitous Arp broadcast will repeat every 3 seconds for [<vip>]"
time="2024-03-07T18:53:02Z" level=info msg="Broadcasting NDP update for <vip> (<mac>) via eth0"
time="2024-03-07T18:53:02Z" level=info msg="ndp: &{false false true <vip> [0xc000051c60]}"
time="2024-03-07T18:53:02Z" level=error msg="Error querying backends file does not exist"
time="2024-03-07T18:53:02Z" level=error msg="Unable to create an IPVS service, ensure IPVS kernel modules are loaded"
time="2024-03-07T18:53:02Z" level=fatal msg="IPVS service error: netlink receive: invalid argument"
```

**To Reproduce**
Helm values:

```yaml
kube-vip:
  image:
    tag: "v0.7.1"
  nameOverride: "kube-vip-v6"
  config:
    address: <vip>
  env:
    vip_arp: "true"
    vip_interface: eth0
    vip_port: "6443"
    lb_enable: "true"
    cp_enable: "true"
    svc_enable: "false"
    vip_cidr: "128"
    prometheus_server: ":2113"
  tolerations:
    - effect: NoSchedule
      key: node-role.kubernetes.io/master
    - effect: NoSchedule
      key: node-role.kubernetes.io/control-plane
      operator: Exists
    - effect: NoExecute
      key: node-role.kubernetes.io/master
  affinity:
    nodeAffinity:
      requiredDuringSchedulingIgnoredDuringExecution:
        nodeSelectorTerms:
          - matchExpressions:
            - key: node-role.kubernetes.io/master
              operator: Exists
          - matchExpressions:
            - key: node-role.kubernetes.io/control-plane
              operator: Exists
```

k3s is started with:

```
'--kube-proxy-arg=proxy-mode=ipvs' \
[...]

PR #783, opened 2024-03-10
Fix netmask of ipvs to consider ipv6 case

lubronzhan:
Fix https://github.com/kube-vip/kube-vip/issues/781


thebsdbox (approved):
Looks good, we eagerly merged the original net mask in order to get ipvs LB working again.

Issue #808, opened 2024-04-08
The enable service security feature is not working in some environment due to the iptables-legacy executable not being found


starbops:
**Describe the bug**

When having a kube-vip setup with an `enable_service_security=true` config, creating new LB type of services results in the "iptables-legacy executable not found" log in the kube-vip pod. Hence kube-vip cannot install the relevant iptables rules to limit the traffic other than the specified ones.

```shell
$ kubectl -n harvester-system logs kube-vip-mgdgr                                                                       
time="2024-04-08T05:22:51Z" level=info msg="Starting kube-vip.io [v0.7.2]"
time="2024-04-08T05:22:51Z" level=info msg="namespace [kube-system], Mode: [ARP], Features(s): Control Plane:[false], Services:[true]"
time="2024-04-08T05:22:51Z" level=info msg="No interface is specified for VIP in config, auto-detecting default Interface"
time="2024-04-08T05:22:51Z" level=info msg="prometheus HTTP server started"
time="2024-04-08T05:22:51Z" level=info msg="kube-vip will bind to interface [mgmt-br]"
time="2024-04-08T05:22:51Z" level=info msg="Starting Kube-vip Manager with the ARP engine"
time="2024-04-08T05:22:51Z" level=info msg="beginning services leadership, namespace [harvester-system], lock name [plndr-svcs-lock], id [harvester-qjl6m]"
I0408 05:22:51.367476       1 leaderelection.go:250] attempting to acquire leader lease harvester-system/plndr-svcs-lock...
I0408 05:22:51.375358       1 leaderelection.go:260] successfully acquired lease harvester-system/plndr-svcs-lock
time="2024-04-08T05:22:51Z" level=info msg="(svcs) starting services watcher for all namespaces"
time="2024-04-08T05:22:51Z" level=info msg="(svcs) adding VIP [[192.168.48.240]] for [kube-system/ingress-expose]"
time="2024-04-08T05:22:51Z" level=info msg="[service] synchronised in 4ms"
time="2024-04-08T05:22:51Z" level=info msg="(svcs) adding VIP [[192.168.48.51]] for [default/test-vm-lb]"
time="2024-04-08T05:22:51Z" level=warning msg="could not add iptables rules to limit traffic ports: could not create iptables client: exec: \"iptables-legacy\": executable file not found in $PATH"
time="2024-04-08T05:22:51Z" level=info msg="[service] synchronised in 2ms"
```

[...]

PR #809, opened 2024-04-08
fix: add iptables-legacy package

starbops:
Adding `iptables-legacy` package for backward compatibility as Alpine 3.19 defaults to nft and legacy symlinks no longer exist.

Related issue: #808


thebsdbox:
Thanks this fixes a few issues and confusion.

Issue #810, opened 2024-04-09
Kube-vip depends on `nodename == hostname` for service election, which easily breaks


d-uzlov:
**Describe the bug**

It seems like kube-vip compares `sh -c hostname` to k8s node name when determining on which node a pod is located.
This works because k8s automatically assigns node name to `sh -c hostname`.
This doesn't work at all in case you alter node name in any way.

Kube-vip shouldn't depend on underlying hostname. Node names should only be compared to other node names.

The solution whould be to use k8s downward API: https://kubernetes.io/docs/concepts/workloads/pods/downward-api/, like suggested in a similar issue: https://github.com/kube-vip/kube-vip/issues/520#issuecomment-1883800048

**To Reproduce**

1. Create single node kubeadm cluster: `sudo kubeadm init --node-name my-custom-node-name`
2. Remove control plane taint: `kl taint node my-custom-node-name node-role.kubernetes.io/control-plane:NoSchedule-`
3. Deploy kube-vip with service election enabled
4. Deploy test service:

```yaml
---
apiVersion: v1
kind: Service
metadata:
  name: apiserver-lb
  namespace: kube-system
spec:
  type: LoadBalancer
  externalTrafficPolicy: Local
  selector:
    component: kube-apiserver
  ports:
  - port: 6443
    targetPort: 6443
```

5. Check service external IP: `kubectl -n kube-system get svc apiserver-lb`
6. Check connection: `curl https://10.3.7.3:6443 --insecure`

**Expected behavior**

External IP is assugned.
Curl is able to connect to the service via the loadbalancer external IP.

**Actual behavior**

`apiserver-lb` gets the `kube-vip.io/loadbalancerIPs` annotation but its external IP is forever pending.
`status.loadBalancer.ingress` is never assigned.
curl fails with `Connection timed out`.

**Environment (please complete the following information):**
 - Kubernetes Version: 1.29.3
 - Kube-vip Version: 0.7.2

**`Kube-vip.yaml`:**

<details>

```yaml
apiVersion: apps/v1
kind: DaemonSet
metadata:
  labels:
    app.kubernetes.io/name: kube-vip-ds
    app.kubernetes.io/version: v0.7.2
  name: kube-vip-ds
  namespace: kube-system
spec:
  selector:
    matchLabels:
      app.kubernetes.io/name: kube-vip-ds
  template:
    metadata:
      labels:
        app.kubernetes.io/name: kube-vip-ds
[...]

PR #811, opened 2024-04-09
Use node name for service election and lease holder name instead of hostname

d-uzlov:
Addresses this issue:
- https://github.com/kube-vip/kube-vip/issues/810

After this change kube-vip should be able to work with any node names. I tried to remove all notions of hostnames from the code.

This PR adds `vip_nodename` env and `nodeName` command line argument.
kube-vip will fall back to hostname if it is not provided.
By default node name is injected via k8s downward API to all yaml deployments generated by kube-vip.

This PR also removes the logic for short name and FQDN matching, because it is now handled by k8s. Now kubevip only matches the exact node name value from config to exact node name value from endpoint/endpointslice.

This is a quite simple change but it affects a lot of things in the project.

I tested this change locally via this image: `docker.io/daniluzlov/k8s-snippets:kube-vip-0.7.2-nodename2`.
I tested both static pods for apiserver HA, and service loadbalancer with and without service election.
However, I can only test with ARP announcements.
I don't have the setup to test BGP.


pkg/equinixmetal/utils.go
```diff
@@ -29,6 +29,7 @@ func findSelf(c *packngo.Client, projectID string) *packngo.Device {
 	// Go through devices
 	dev, _, _ := c.Devices.List(projectID, &packngo.ListOptions{})
 	for _, d := range dev {
+		// TODO do we need to replace os.Hostname with config.NodeName here?
 		me, _ := os.Hostname()
```
d-uzlov, line 33: Maybe here we also want to use node name as reported by k8s, but since I don't have BGP setup to test it, I left it as is, to avoid breaking stuff.

pkg/manager/watch_endpoints.go
```diff
@@ -82,41 +82,23 @@ func (ep *endpointsProvider) getAllEndpoints() ([]string, error) {
...
+			if address.NodeName != nil && id == *address.NodeName {
+				log.Debugf("[%s] found local endpoint - address: %s, hostname: %s, node: %s", ep.label, address.IP, address.Hostname, *address.NodeName)
+				localEndpoints = append(localEndpoints, address.IP)
+				continue
+			}
+			// 2. Compare the Hostname (only useful if node name is not available)
+			if id == address.Hostname {
+				log.Debugf("[%s] found local endpoint - address: %s, hostname: %s", ep.label, address.IP, address.Hostname)
+				localEndpoints = append(localEndpoints, address.IP)
+				continue
```
d-uzlov, line 101: Here and in `watch_endpointslices.go` I removed short hostname-to-nodename matching because k8s is already providing the correct value.
First I check `address.NodeName` because documentation states "This can be used to determine endpoints local to a node".
`address.Hostname` doesn't seem relevant to me, but I left it here as a second check, just in case this is relevant in some setups. Maybe it should also be removed.

thebsdbox:
Unfortunately this is going to break all kube-vip installations and upgrades moving forward, I think defaulting to detecting to hostname detection in the case of manually specifying one should probably remain.

pkg/manager/manager.go
```diff
@@ -67,6 +68,10 @@ func New(configMap string, config *kubevip.Config) (*Manager, error) {
 	adminConfigPath := "/etc/kubernetes/admin.conf"
 	homeConfigPath := filepath.Join(os.Getenv("HOME"), ".kube", "config")
 
+	if config.NodeName == "" {
+		return nil, errors.New("node name is missing from config")
```
thebsdbox, line 72: I haven't tested yet, but wouldn't going back to os.Hostname() restore the previous functionality/behaviour?

thebsdbox:
This would be great, I think we'll just need to ensure that default/expected functionality remains :)

d-uzlov:
> Unfortunately this is going to break all kube-vip installations and upgrades moving forward, I think defaulting to detecting to hostname detection in the case of manually specifying one should probably remain.

When making these changes I thought of using kube-vip to generate new deployment files. This way there will be no issues during the upgrade.
But if incrementing the image version in an existing pod manifest should be a supported option, I guess we need to create a fallback to the old behavior.

I'll push an update commit for this.

> Additionally when the code runs in control-plane mode host name is needed as there is no node object.

What do you mean by this?
I believe that downward API overall and node name in particular is available right from the start of the cluster.
I tried testing my image both in an existing cluster (upgrading control plane static pods from v0.7.2 one by one), and creating a new cluster with kube-vip bootstrap, and I didn't see any issues in both cases.

pkg/manager/manager.go
d-uzlov: Yeah, it would. I thought of doing it initially, but I thought I could make the code a bit simpler by omitting it.

thebsdbox:
> What do you mean by this?

When kube-vip is managing the HA VIP for the control plane (when being brought up by `kubeadm`), there is a particularly amusing/annoying scenario where kube-vip needs to come up "before" the API server is fully up. The reason is that kube-vip needs to have the HA address working in order for the `kubeadm init` process to complete successfully. Kube-vip will typically use a pod DNS spec to have 127.0.0.1 or the local address resolve to `kubernetes` so that the certificate SAN works. So the `tl;dr` is that the control plane side of things is a little bit complicated 😂

d-uzlov:
Well, kube-vip is already using `127.0.0.1 kubernetes` hosts entry, node hostname is not involved here.
And I tested that current set up works as it is.
So I guess using node name for leases and pod placement checks doesn't affect this, and we don't need to handle it any differently.

pkg/manager/manager.go
d-uzlov: Done

thebsdbox: Oops, looks like this has broken the build. `make check` will run all of the linting locally.

d-uzlov: Yeah, was an embarrassing small error/typo, sorry about that :-)

d-uzlov:
@thebsdbox 
Should I also modify unit/e2e tests to include node name option, and add description to documentation, or would you prefer to do it yourself?

thebsdbox:
> @thebsdbox
> Should I also modify unit/e2e tests to include node name option, and add description to documentation, or would you prefer to do it yourself?

Yes please, that way we can be sure that it's working as expected 😀

d-uzlov:
@thebsdbox I made sure that currently existing tests run fine for me locally, and tried to add tests for node names.

Ideally we would want to create nodes with modified names but kind doesn't seem to allow it, so I tried to at least test something.

For e2e tests I modified the main manifest and the etcd manifest to use the new node name env, and then added a new e2e test that uses manifest without this env, to check that old deployments will still work.

For service tests I figured out I can modify cluster a bit before running kube-vip.
I change hostnames so that they no longer match node names, so that, if you try to run kube-vip without the new node name env/flag, it won't find local pods and will fail to assign loadbalancer IPs to nodes.
I then added the env to the daemonset config accordingly.

d-uzlov:
I also wanted to update documentation but turns out it's in a separate repo, so I guess maybe later, after this PR is merged.
Well, this feature shouldn't need user configuration any way, it should just work passively. Documentation will just help users better understand what the new env/flag means.

Cellebyte:
@thebsdbox we need to make sure that new features get also properly documented, I think we already got some drift between code and docs.

thebsdbox:
> @thebsdbox we need to make sure that new features get also properly documented, I think we already got some drift between code and docs.

100% my plan to separate the repos to make things easier, sadly made it worse. Perhaps updating the templates for a corresponding website PR would make sense.

thebsdbox:
All tests are passing, I'll have one more look over but is this ready to merge?

d-uzlov:
For me this PR is ready to merge, I don't have anything more to change.

PR #832, opened 2024-04-24
fix: unecessary string modification

ShivanshVij:
The usage of `strings.ToLower` in the demo server implementation is unnecessary because it's a hard-coded lowercase string already. 

Tiny fix, I know, but it was really bugging me.

### DIFF
--- a/demo/server/main.go
+++ b/demo/server/main.go
@@ -32,7 +32,7 @@ func main() {
 		}
 	}
 
-	if serverType == strings.ToLower("udp") {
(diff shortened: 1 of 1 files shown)


thebsdbox (approved):
Haha, I love this!

`/lgtm`

ShivanshVij:
I'm not sure the failure here is caused by my changes...

thebsdbox:
The CI test build four clusters does HA and breaks them with a small timeout, we sometimes get flakes.

Issue #835, opened 2024-04-29
The load balancer IP addresses are not being release after the LB-type Services are deleted


starbops:
**Describe the bug**

In ARP mode, kube-vip does not remove the load balancer IP address from the node's network interface when the LB-type Service object is deleted.

**To Reproduce**

1. Follow the [document](https://kube-vip.io/docs/usage/k3s/) to prepare the RBAC and kube-vip DaemonSet manifests
2. Create a K3s cluster consists of three nodes using `k3sup`
3. Create a test Deployment called `blog` with `nginx` image
   ```shell
   kubectl create deployment blog --image=nginx --replicas=3
   ```
4. Create a LB-type Service exposing port 80 of the `blog` Deployment
   ```shell
   kubectl expose deployment blog --port 80 --type=LoadBalancer --overrides='{"metadata": {"annotations": {"kube-vip.io/loadbalancerIPs": "192.168.100.201"}}}'
   ```
   The logs generated by the kube-vip leader Pod:
   ```shell
   $ kubectl -n kube-system logs kube-vip-ds-68ldv
   time="2024-04-29T08:28:59Z" level=info msg="Starting kube-vip.io [v0.8.0]"
   time="2024-04-29T08:28:59Z" level=info msg="namespace [kube-system], Mode: [ARP], Features(s): Control Plane:[true], Services:[true]"
   time="2024-04-29T08:28:59Z" level=info msg="prometheus HTTP server started"
   time="2024-04-29T08:28:59Z" level=info msg="Using node name [k3s-03]"
   time="2024-04-29T08:28:59Z" level=info msg="Starting Kube-vip Manager with the ARP engine"
   time="2024-04-29T08:28:59Z" level=info msg="beginning services leadership, namespace [kube-system], lock name [plndr-svcs-lock], id [k3s-03]"
   I0429 08:28:59.926436       1 leaderelection.go:250] attempting to acquire leader lease kube-system/plndr-svcs-lock...
   time="2024-04-29T08:28:59Z" level=info msg="Beginning cluster membership, namespace [kube-system], lock name [plndr-cp-lock], id [k3s-03]"
   I0429 08:28:59.926733       1 leaderelection.go:250] attempting to acquire leader lease kube-system/plndr-cp-lock...
   time="2024-04-29T08:28:59Z" level=info msg="Node [k3s-02] is assuming leadership of the cluster"
   time="2024-04-29T08:28:59Z" level=info msg="new leader elected: k3s-02"
[...]

PR #837, opened 2024-04-30
fix: set service's uid in activeService map for inactive services

starbops:
Related issue: #835


thebsdbox (approved):
Strange, I'll need to update the tests as this wasn't picked up.

thebsdbox:
Thanks

PR #854, opened 2024-05-17
[Hotfix] Routes get not deleted when kube-vip is running without LeaderElection/ServiceElection in Table Mode

Cellebyte:
* When running kube-vip without `vip_leaderelection` or `svc_election` it only deletes IPs from interfaces but never deleted the Route from the RoutingTable in Table Mode
* This change would introduce a bit of cluttering in the logs, when Routes are not installed on a particular node but kube-vip is trying to delete the route for a service when in TableMode
* Functionally it should not break any existing setup using TableMode it just cleans Routes after the Service gets deleted


thebsdbox:
`/lgtm`

thebsdbox (approved):

thebsdbox (approved):

PR #855, opened 2024-05-17
New service class filtering behavior

### LINKED ISSUE #851: [LoadBalancerClass] Better defaulting for Service objects
**Describe the bug**
When you define a LoadBalancerClass via `env` or cli flag **kube-vip** only reconciles `Service` objects with this `LoadBalancerClass`.
This is expected behaviour and should stay like that.
The problem is when you have **kube-vip** running without a `LoadBalancerClass` defined, it also reconciles `Services` with a `LoadBalancerClass`. This would lead to unexpected behaviour as it could announce IPs which it should not.

e.g.
Ingress controllers without a specific `IngressClass` set, only reconcile Ingresses without a `IngressClass` and ignore all `Ingresses` which have such `IngressClass` defined.

**To Reproduce**
1. Install kube-vip with svc_enable and the mode you like but without LoadBalancerClassName set
2. Create 2 Services

```yaml
# a normal Service without a loadbalancerclass
apiVersion: v1
kind: Service
metadata:
  name: default
spec:
  externalTrafficPolicy: Local
  ports:
  - name: http
    nodePort: 31690
    port: 80
    protocol: TCP
    targetPort: http
  - name: https
    nodePort: 32093
    port: 443
    protocol: TCP
    targetPort: https
  selector: {}
  sessionAffinity: None
  type: LoadBalancer
status:
  loadBalancer:
    ingress:
    - ip: 10.0.0.0
```

```yaml
# Service with a loadBalancerClass set
apiVersion: v1
kind: Service
metadata:
  name: default
spec:
  externalTrafficPolicy: Local
  ports:
  - name: http
    nodePort: 31690
    port: 80
    protocol: TCP
    targetPort: http
  - name: https
    nodePort: 32093
    port: 443
    protocol: TCP
    targetPort: https
  selector: {}
  sessionAffinity: None
  type: LoadBalancer
  loadBalancerClass: kube-vip.io/kube-vip-class
status:
  loadBalancer:
    ingress:
    - ip: 10.0.0.1
```

**Expected behavior**
The second service from the example above should not be reconciled by **kube-vip** but rather be skipped.

**Screenshots**

 - OS/Distro: Ubuntu 20.04
 - Kubernetes Version: v1.27.13
 - Kube-vip Version: v0.8.0

p-strusiewiczsurmacki-mobica:
This PR introduces new behavior for LoadBalancer Class Name service filtering. With this new implementation only services that have LB class name set to the value configured in kube-vip (`LoadBalancerClassName` flag) will be processed. This is consistent with implementation in e.g. [MetalLB](https://github.com/metallb/metallb/blob/main/internal/k8s/controllers/service_controller.go#L159) or with how the ingress classes behave.

However this might be a braking change, so it is implemented (for now) as an option. 'Legacy' behavior is a default and can be disabled by setting `LoadBalancerClassLegacyHandling` config flag to `false`. In one of next releases this could become a default setting, and then legacy behavior could be removed altogether.

Fixes #851

### DIFF
--- a/cmd/kube-vip.go
+++ b/cmd/kube-vip.go
@@ -129,6 +129,7 @@ func init() {
 	kubeVipCmd.PersistentFlags().BoolVar(&initConfig.EnableServicesElection, "servicesElection", false, "Enable leader election per kubernetes service")
 	kubeVipCmd.PersistentFlags().BoolVar(&initConfig.LoadBalancerClassOnly, "lbClassOnly", false, "Enable load balancing only for services with LoadBalancerClass \"kube-vip.io/kube-vip-class\"")
 	kubeVipCmd.PersistentFlags().StringVar(&initConfig.LoadBalancerClassName, "lbClassName", "kube-vip.io/kube-vip-class", "Name of load balancer class for kube-VIP, defaults to \"kube-vip.io/kube-vip-class\"")
+	kubeVipCmd.PersistentFlags().BoolVar(&initConfig.LoadBalancerClassLegacyHandling, "lbClassNameLegacyHandling", true, "Use legacy LoadBalancer class name handling (e.g. accepting services both with empty and non-empty class)")
 	kubeVipCmd.PersistentFlags().BoolVar(&initConfig.EnableServiceSecurity, "onlyAllowTrafficServicePorts", false, "Only allow traffic to service ports, others will be dropped, defaults to false")
 	kubeVipCmd.PersistentFlags().BoolVar(&initConfig.EnableNodeLabeling, "enableNodeLabeling", false, "Enable leader node labeling with \"kube-vip.io/has-ip=<VIP address>\", defaults to false")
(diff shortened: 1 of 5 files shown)


thebsdbox:
This looks OK to me, would it be possible to add a service test to check for labels?

p-strusiewiczsurmacki-mobica:
Hi @thebsdbox, I am not really sure. I've just try to play with the tests a bit, but I see 2 issues:

1. I've tried to add LoadbalancerClass to the existing tests, just to check if those would pass with default `kube-vip.io/kube-vip-class` class, but it seems that `v0.9.0` of `kube-vip-cloud-provider` discards any service which has LoadbalancerClass set. Would need to try with current main as it seems some additional controller for loadbalancer classes was added. As service has no loadbalancer IP assigned by cloud-provider, it is discarded by kube-vip as well and the test fails.

2. I can't see any clear way of determining if service was processed or discarded by kube-vip (other than reading logs, obviously :D). I've tried checking `ip neighbour` on the kind nodes, but I am not sure if this is a proper way to do this.

thebsdbox:
I'll look into tests, is this ready to merge?

Cellebyte:
@thebsdbox if you deal with the tests. I think so.

PR #868, opened 2024-06-04
Small correction to docs on retryPeriod property

murphp15:


thebsdbox:
Ah. linting issue, you can use `make check` to see what's wrong :)

Issue #630, opened 2023-10-17
prometheusHTTPServer Set to Blank Fails to Stop Prometheus Server


jjschwarz:
In case #625 it is mentioned that going forward we should use "manager" instead of "start" for all instances of kube-vip.  I made that change and added --prometheusHTTPServer "" to my manifest generation for the secondary IPv6 instance.  That removed the prometheus_server values from the manifest, however on attempting to run the new manifest it still tried to start prometheus on port 2112, causing a crash of the secondary instance for IPv6.  

How should this be approached for secondary instances if we should be using manager?

thebsdbox:
Hmm I wonder if the IPv6 code path looks at the prometheus side of things a little differently. (I'll investigate).. Manager is going to be the only option from `v0.7.0`, so should make it easier as we can clean some of the code path out.

jjschwarz:
Any update on the --prometheusHTTPServer "" not functioning correctly for IPv6?  I would love to get this resolved so we can get rid of using "start" and move to just "manager".

jjschwarz:
So this is still an issue in 0.8.0:

I used this to create the manifest:

sudo ctr run --net-host --rm ghcr.io/kube-vip/kube-vip:v0.8.0 vip /kube-vip manifest pod \
--interface ens192 \
--address 2620:128:2000:254::80 \
--controlplane \
--services \
--arp \
--prometheusHTTPServer "" \
--leaseName plndr-cp-lock6 \
--servicesLeaseName plndr-svcs-lock6 \
--leaderElection  | sed "s/name: kube-vip/name: kube-vip6-ds/g" | sudo tee /etc/kubernetes/manifests/vip6.yaml

It removes the prometheus_server key, but  I am still getting the following error:
time="2024-06-19T18:17:38Z" level=fatal msg="listen:listen tcp :2112: bind: address already in use\n"

thebsdbox:
https://kube-vip.io/docs/usage/multi_tenancy/?query=prom#prometheus-conflicts

It needs to be in the manifest itself, I can add it to the cli if it will
make life easier.

jjschwarz:
For automation purposes it would be handy to be able to generate the proper manifest from the CLI.

jjschwarz:
This appears to be fixed in 0.82 per case 883 and is testing fine for me.

PR #883, opened 2024-06-21
This allows prometheus configuration to be specified in the cli

thebsdbox:
`--prometheusHTTPServer ""` will now create

```
    - name: prometheus_server
```

with no `- value`, which should stop prometheus from starting.


thebsdbox:
Fix for #630

PR #889, opened 2024-06-25
[Bugfix] Parse the routingTableID correctly 

Cellebyte:
Current Behaviour:

kube-vip does not start.
```
time="2024-06-25T11:45:39Z" level=fatal msg="strconv.ParseInt: parsing \"2347908769\": value out of range"
```

With this fix it should start.

References:
* [rtnetlink.h](https://github.com/torvalds/linux/blob/55027e689933ba2e64f3d245fb1ff185b3e7fc81/include/uapi/linux/rtnetlink.h\#L354C15-L354C25) 
* RT_TABLE_MAX=0xFFFFFFFF

### DIFF
--- a/pkg/kubevip/config_environment.go
+++ b/pkg/kubevip/config_environment.go
@@ -329,7 +329,7 @@ func ParseEnvironment(c *Config) error {
 		} else if i < 0 {
 			return fmt.Errorf("no support of negative [%d] in env var %q", i, vipRoutingTableID)
 		}
-		return fmt.Errorf("no support for int64 natively on system only [int%d]", bits.OnesCount(math.MaxUint))
+		return fmt.Errorf("no support for int64, system natively supports [int%d]", bits.OnesCount(math.MaxUint))
 	}
 
 	// Routing Table Type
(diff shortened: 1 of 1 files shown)


thebsdbox (approved):
/lgtm

Cellebyte:
@thebsdbox I am just waiting on some feedback from @schrej

pkg/kubevip/config_environment.go
```diff
@@ -316,11 +319,17 @@ func ParseEnvironment(c *Config) error {
...
 			return err
 		}
-		c.RoutingTableID = int(i)
+		if i >= 0 && i <= math.MaxInt {
+			c.RoutingTableID = int(i)
+			return nil
+		} else if i < 0 {
+			return fmt.Errorf("no support of negative [%d] in env var %q", i, vipRoutingTableID)
+		}
+		return fmt.Errorf("no support for int64 natively on system only [int%d]", bits.OnesCount(math.MaxUint))
```
schrej, line 332: missing a comma here, and for consistency I would use `math.MaxInt` here as well, if that yields the same result.

schrej (approved):

pkg/kubevip/config_environment.go
Cellebyte: nope as it has a signing bit which is zero for a positive number.

Cellebyte: fixed the error message.

PR #897, opened 2024-07-19
Change service interface to `macvlan` when using DHCP

### LINKED ISSUE #871: Consistent DHCP renew failure analysis
Potentially related issues: #572, #580

> ### **TL;DR** 
> **In certain network setups, DHCP renewals performed by `kube-vip` generally fail.** Further implementation issues result in ARP announcements of an IP for which a lease is no longer held.
> 
> The core reason for the renewal failure is that the [L2 source MAC](https://en.wikipedia.org/wiki/Medium_access_control#Functions_performed_in_the_MAC_sublayer) of outgoing DHCP renewal packets is that of the `eth` interface while the [client MAC (CHADDR) specified in the DHCP payload](https://en.wikipedia.org/wiki/Dynamic_Host_Configuration_Protocol#Request) is that of the `macvlan`.
> 
> Some DHCP servers send their reply to the latter instead of the former. In combination with fast-expiring switch MAC-Port tables, this results in response packets being dropped.
> 
> _Furthermore, **`kube-vip` continues to send out ARP announcements for IP leases it no longer holds** due to a failed rebind._
>
> We* propose a set of solutions and seek input from the maintainers. Depending on the preferred one we would be happy to create a Pull Request.
> 
> <sub>*Me and my colleague @MarioArkenberg, who has been instrumental in unravelling this rather involved bug in a wild goose chase over the past couple months 🙈</sub>

This bug is comprised of a root problem that can be tracked down [to a single line of code](https://github.com/kube-vip/kube-vip/blob/c16f634cf7ef4acd3037f4defe48ee911df7e013/pkg/vip/dhcp.go#L281), as well as a handful contributing issues in the DHCP implementation that amplify the symptoms caused by it. Due to this, an explanation of the context is required which can be found below.

Regarding the contributing issues, let us know if you want separate issues to track them. We are not fully confident in providing a proper fix for them but can try based on some prior discussion of the desired solution.

**Note:** `eth` refers to the physical, user-configured `serviceInterface`.

## Root problem

> **Reminder:** In order to perform DHCP `kube-vip` creates a separate `macvlan` interface which is bound to the configured `eth` interface. The acquired IP will be assigned to the `eth` instead of the `macvlan`.

The [DHCP client logic](https://github.com/kube-vip/kube-vip/blob/main/pkg/vip/dhcp.go) of `kube-vip` operates differently for:

- Rebinds & initial lease acquisition 
  - Packets are sent from a raw socket on the `macvlan` interface using its MAC and a `0.0.0.0` source IP address (since the interface has no IP)
- Renewals 
  - Packets are sent from the `eth` interface with its MAC and the leases' IP address

The latter is due to [`pkg/vip/dhcp.go:280`](https://github.com/kube-vip/kube-vip/blob/c16f634cf7ef4acd3037f4defe48ee911df7e013/pkg/vip/dhcp.go#L280) where the DHCP client is initialized as a unicast client based on the IP address from the lease. Since the IP is assigned to the `eth` and not the `macvlan`, the socket is automatically opened on the `eth` interface. The argument containing the interface name containing the `macvlan` is completely ignored by `nclient4`!

### Network setup

Before we continue, we need to take a high-level look at our network infrastructure:

`kube-vip` created the `macvlan` interface which is bound to the physical `eth`. This in turn is connected to the switch. The DHCP server is connected to the same switch.

### Traffic routing in switches

Many switches use MAC tables to decide which ports should receive what traffic. For every received packet it:

- Stores an association between the source MAC and physical port
- Looks up the destination MAC in the table

Based on the result, it decides to which port a packet shall be sent. However, entries expire. That raises the question: _What happens if no entry  is present for the destination MAC?_

In our particular case, the expiry time was 30 minutes and crucially **non-broadcast packets with no port association are being dropped!\***

<sub>\*Some switches forward packets without a known destination to all ports while others only do so for broadcast packets. Either way, it is not consistent and can not be relied upon.</sub>

### Temporal causality

> **Note:** Below is a simplified DHCP flow diagram which might help in understanding this section.
> 
> dhcp-diagram

Now that we have set the stage, let us consider the traffic flow of a usual `kube-vip` DHCP lease acquisition and subsequent renewal:

1. DHCP Discover is sent
    - Source MAC is that of the `macvlan` interface
    - Offer/Request/ACK workflow is performed, lease is acquired
2. DHCP Renew is sent after a while
    - Source MAC is that of the `eth` interface
    - DHCP client MAC (CHADDR) is that of the `macvlan` interface (!)
3. Server sends response
    - Destination is set to the CHADDR instead of the L2 source MAC (!)

This mismatch in source and destination address combined with the previously discussed traffic routing logic causes the response packet to never reach its destination.

packet-trace

Since the `macvlan` interface is exclusively used for DHCP, there is no outbound traffic from it between lease acquisition and the renewal. As a result, the MAC table entry in the switch expires before the renewal is performed.

Normally this is not a problem, since the outbound renewal packet would have a source MAC that matches the responses' destination (creating an entry in the table of the switch). In this case, however, they do not match and thus there is no table entry for the response destination.

Since, as mentioned previously, our switch drops packets for which there is no destination, the response never reaches `kube-vip` and renewals always fail!

### Resulting behaviour

As a result, lease renewals never* succeed and `kube-vip` always resorts to rebinds. Theoretically, failed renews would be caught by rebinds whose implementation uses a proper source+client combination. However if a rebind fails, all-hell-breaks-loose™ due to the contributing issues listed in the next section.

Across ~200 LoadBalancer services we ran into failed rebinds a few times per week which warranted this in-depth analysis focused on the renewal failures. So far, we did not dig deeper into the underlying cause of the rebind failures.

<sub>\*In our case, the kernel assigned the `macvlan` a link-local IPv6 address. This resulted in occasional `ICMPv6 Router Solicitation` requests being broadcast which refreshed the switches' MAC table. To a degree, this masked the issue since some renewals got through if they happened to be sent in close temporal proximity to one of the router solicitations.</sub>

## Contributing issues

This is a list of implementation issues in the DHCP stack of `kube-vip` that we encountered during our analysis. Some only amplified the problem while others may be considered flawed behaviour which never caused trouble so far.

These contributing issues played a big role in the primary bug having the impact it did and should be considered as bugs on their own.

### Lease duration ignored (mild inconvenience)

In our case either `kube-vip` or its upstream dependency `nclient4` ignores the lease duration provided by the DHCP server and instead always assumes `1h`. We have not investigated this yet. That being said, this significantly increased the likelihood of running into the issue.

### No DHCP reinit (minor bug)

When a rebind fails, the lease shall be dropped and the DHCP logic should start over from the beginning. Despite the in-code ASCII flow diagram depicting this, it never happens.

Upon rebind failure, the `DHCPClient.Start` function [simply returns](https://github.com/kube-vip/kube-vip/blob/c16f634cf7ef4acd3037f4defe48ee911df7e013/pkg/vip/dhcp.go#L161) after logging an error. 

Unfortunately, this method is [invoked in a "fire-and-forget" manner](https://github.com/kube-vip/kube-vip/blob/c16f634cf7ef4acd3037f4defe48ee911df7e013/pkg/vip/ddns.go#L45) by just forking it into a goroutine without tracking its status. Once the function returns, it is gone. No reinitialization is performed.

### Expired leases' IP is still announced (major issue)

Not reinitializing in itself is still an error but one that is somewhat easily visible as the LoadBalancer service would looses its IP.

Unfortunately, this is not what happens. It appears that `kube-vip` only monitors for a change in IP address but not for lease expiry. As a result, the service keeps the address and ARP announcements continue being sent out.

The effect of this is that multiple hosts in the same subnet unknowingly announce the same IP (once the DHCP server reassigns the IP) which leads to incoming traffic being randomly directed to either of the two hosts, rendering LoadBalancer services functionally useless with a hard-to-debug failure mode.

### IP not checked before use (minor issue)

According to [RFC 2131, Section 2.2, Paragraph 2](https://datatracker.ietf.org/doc/html/rfc2131#section-2.2) both the server and the client should probe the IP address of a new lease to see if another server still uses it (which is the case here due to the above bug). 

However, in our particular scenario neither the DHCP server nor most clients (both `kube-vip` and other ones) actually perform this check. As a result, the still-in-use-but-technically-free IP is simply reassigned.

If either the server or every client in the network would be spec-compliant, the worst effects of the use-after-expiry bug could've been averted. Though given the wild-west nature of DHCP implementations this sadly did not materialize 🫠

## Fixing the problem

In general, the current behaviour of using different outbound interfaces for binds/rebinds and renewals seems very strange and non-compliant. I personally would favour a variant where renewals are sent from the `macvlan` interface by default. 

That being said, I lack the understanding of the problem some DHCP servers have with the packet originating from an interface that does not have the IP assigned (i.e. what the code-comment describes). If somebody has some input or pointers, let me know!

**We would be happy to create a Pull Request!**

However, since the problem is rather complex and the solution space littered with tradeoffs, we are looking for input from any maintainers (both on which path to choose and whether a fix would be accepted in the first place).

### Potential solutions

We have some ideas for potential solutions but want to give you a chance to come up with your own ones without bias. If you want to, please have a look:

<details>

<summary>Set of potential solutions</summary>

Considering just the underlying root problem — the mismatch between the L2 source MAC and DHCP client MAC for renewals — there are multiple possible solutions that we could come up with:

1. Assign the IP to the `macvlan` interface instead of the `eth` one
    - Very clean solution that should satisfy all DHCP servers
    - Would resolve #766
    - Probably a larger change
        - Per-service `serviceInterface` parameter determined after creation of the `macvlan` interface
        - Egress logic needs to be adapted as well
        - Depends on the exact code paths which are unknown to me (might actually be simple)
    - Couples the DHCP logic with the LoadBalancer logic
1. Remove the [`WithUnicast` option for renewals](https://github.com/kube-vip/kube-vip/blob/c16f634cf7ef4acd3037f4defe48ee911df7e013/pkg/vip/dhcp.go#L280).
    - Simplest possible fix, already verified to solve the issue
    - Based on the code-comment some DHCP servers would probably dislike this
    - It is unclear (to us) if there is a way to use a raw socket (like for rebinds) and still make all DHCP servers happy. It might be possible but more context for that code-comment is needed.
1. Send dummy packets to keep switch tables populated
    - Very hacky solution, unnecessarily spams the network
    - Relies on assumptions about MAC table lifetimes
1. Attempt renewals using both approaches
    - Guaranteed™ to work
    - Somewhat hacky, produces unnecessary traffic
1. Add config flags to select renewal behaviour
    - Allows the user to decide based on their setup
    - No excess traffic "on-the-wire"
    - Requires user understanding of the problem involved
1. Something entirely different :D

</details>

## Environment

- talos-linux `1.6.5`
- Kubernetes `1.29`
- kube-vip `0.8.0`
    - Used for LoadBalancer services only
- Network switch
  - Cisco Nexus 9000 93180YC-EX chassis
  - Nexus OS version 9.2(1)
- DHCP Server
    - Microsoft Windows Server 2019 Standard

### Collected data & packet traces

We do have a large set of packet captures and other data. However, since the above network setup description is somewhat simplified, I am not attaching them to prevent confusion.

In regards to the underlying bug we have a high confidence in our results. It took us weeks to analyse and understand the above causality based on said data. Transferring knowledge about all the details would take a considerable amount of time and we assume that an exact description of the entire context (and thus the ability to understand the raw data) beyond the above would neither be helpful nor necessary for a solution*.

That being said, if anyone wants to get into a deep-dive, we are absolutely open to it!

<sub>*While being fully aware of the [XY problem](https://en.wikipedia.org/wiki/XY_problem) here 🙈</sub>

TilBlechschmidt:
When using DHCP for LB services, the resulting IP was assigned to the `serviceInterface` while most (not all) DHCP communication was done through a per-service `macvlan` interface that is created on demand.

This has been the cause of DHCP lease renewal failures for some users, as the response packets from the server never reached kube-vip. For more details, see issue #871.

To remedy this, the primary interface used for a service is now being set to the `macvlan`. Consequently, the IP address is also assigned to it.

As a result, all traffic related to an LB service (DHCP or not) is now originating from and sent to the per-service `macvlan` interface that is tied to the user-configured `serviceInterface`.

Fixes #871

### DIFF
--- a/pkg/manager/instance.go
+++ b/pkg/manager/instance.go
@@ -125,10 +125,12 @@ func NewInstance(svc *v1.Service, config *kubevip.Config) (*Instance, error) {
 			return nil, fmt.Errorf("error starting DHCP for %s/%s: error: %s",
 				instance.serviceSnapshot.Namespace, instance.serviceSnapshot.Name, err)
 		case ip := <-instance.dhcpClient.IPChannel():
+			instance.vipConfigs[0].Interface = instance.dhcpInterface
 			instance.vipConfigs[0].VIP = ip
 			instance.dhcpInterfaceIP = ip
 		}
 	}
+
 	for _, vipConfig := range instance.vipConfigs {
 		c, err := cluster.InitCluster(vipConfig, false)
 		if err != nil {


TilBlechschmidt:
Despite the [`Contributing.md`](https://github.com/kube-vip/kube-vip/blob/a571d0a9c2e98989d2b1df2961d07b732f2833d6/CONTRIBUTING.md?plain=1#L85) mentioning a `CHANGELOG.md`, I could not find any in the repository or actions workflows.

I have added some context to the commit body, let me know if there is another place to put release notes 🙂

thebsdbox:
Awesome, thanks for the PR. It's difficult to have tests for the DHCP portion within CI so we'll have to watch for any issues that arise from this change moving forward.

thebsdbox:
`/lgtm`

Issue #863, opened 2024-05-29
nodelabel is not applied after a leader change


cdtzabra:
Hello
**Describe the bug**

In my cluster I have 2 nodes (workers) dedicated for the egress with some taints.

I deploy kube-vip only on the both node with toleration

When the kube-vip leader is elected,  the node has the label: `kube-vip/has-ip=my_vip`
And i track that nodelabel `kube-vip/has-ip=my_vip`  and use it actif egress node for traffic.

But when the kube-vip leader change, the label **is not added** it the new node.
I got the error `no node label change needed`:  https://github.com/kube-vip/kube-vip/blob/c683bb879ec1fee25233b683d6abc65fb555f6a1/pkg/manager/node_labeling.go#L48

This break all the outgoing traffic.

My daemonset

```yaml
apiVersion: apps/v1
kind: DaemonSet
metadata:
  creationTimestamp: null
  name: kube-vip-ds
  namespace: kube-system
spec:
  selector:
    matchLabels:
      name: kube-vip-ds
  template:
    metadata:
      creationTimestamp: null
      labels:
        name: kube-vip-ds
    spec:
      affinity:
        nodeAffinity:
          requiredDuringSchedulingIgnoredDuringExecution:
            nodeSelectorTerms:
            - matchExpressions:
              - key: node-role.kubernetes.io/egress
                operator: Exists
            - matchExpressions:
              - key: node-role.kubernetes.io/control-plane
                operator: Exists
      containers:
      - args:
        - manager
        env:
        - name: vip_arp
          value: "true"
        - name: port
          value: "6443"
        - name: vip_interface
          value: ens192
        - name: vip_cidr
          value: "32"
        - name: cp_enable
          value: "true"
        - name: cp_namespace
          value: kube-system
        - name: vip_ddns
          value: "false"
        - name: svc_enable
          value: "true"
        - name: vip_leaderelection
          value: "true"
        - name: vip_leaseduration
          value: "5"
        - name: vip_renewdeadline
          value: "3"
        - name: vip_retryperiod
          value: "1"
        - name: address
          value: 10.xx.yy.ZZ
        - name: vip_loglevel
          value: "5"
        - name: svc_electtion
          value: "false"
        - name: enable_node_labeling
          value: "true"
        image: ghcr.io/kube-vip/kube-vip:v0.8.0
        imagePullPolicy: Always
        name: kube-vip
        resources: {}
        securityContext:
          capabilities:
            add:
            - NET_ADMIN
[...]

PR #923, opened 2024-08-21
Ensuring nodelabel will be always present

cdtzabra:
As reported in the issue #863, nodelabel is missing sometimes.

for example, if the VIP is already associated with a node and you delete the kube-vip pods, once they're re-created, there's a good chance that none of your nodes will include the kube-vip/has-ip=xx label.
As a result, the existence of this label is unreliable.

The aim of this change is to ensure that the nodelabel is always added.

### DIFF
--- a/pkg/manager/node_labeling.go
+++ b/pkg/manager/node_labeling.go
@@ -36,6 +36,7 @@ func applyNodeLabel(clientSet *kubernetes.Clientset, address, id, identity strin
 
 	value, ok := node.Labels[nodeLabelIndex]
 	path := fmt.Sprintf("/metadata/labels/%s", nodeLabelJSONPath)
+	log.Debugf("Received identity: %s - id: %$", identity, id)
 	if ok && value == address {
 		log.Debugf("removing node label `has-ip=%s` on %s", address, id)
 		// Remove label


thebsdbox:
Hello, Sorry for the delay. Would it be possible to address the linting issue?

cdtzabra:
> Hello, Sorry for the delay. Would it be possible to address the linting issue?

Hello

Done

pkg/manager/node_labeling.go
```diff
@@ -36,16 +36,15 @@ func applyNodeLabel(clientSet *kubernetes.Clientset, address, id, identity strin
 
 	value, ok := node.Labels[nodeLabelIndex]
 	path := fmt.Sprintf("/metadata/labels/%s", nodeLabelJSONPath)
-	if (!ok || value != address) && id == identity {
-		log.Debugf("setting node label `has-ip=%s` on %s", address, id)
-		// Append label
-		applyPatchLabels(ctx, clientSet, id, "add", path, address)
-	} else if ok && value == address {
+	log.Debugf("Received identity: %s - id: %$", identity, id)
```
Cellebyte, line 39:
```suggestion
	log.Debugf("Received identity: %s - id: %s", identity, id)
```

thebsdbox:
Going to merge this and then fix in a corresponding PR 😄

PR #937, opened 2024-09-05
This adds ipv6 egress & test

thebsdbox:
The original Egress functionality with IPv4 only, this PR adds the capability for egress addresses to be IPv6.

An additional annotation is required, and the `rbac.yaml` will need updating with pod-list/get in order to auto-detect the pod/service CIDRs.

### DIFF
--- a/pkg/manager/services.go
+++ b/pkg/manager/services.go
@@ -23,6 +23,7 @@ const (
 	requestedIP              = "kube-vip.io/requestedIP"
 	vipHost                  = "kube-vip.io/vipHost"
 	egress                   = "kube-vip.io/egress"
+	egressIPv6               = "kube-vip.io/egress-ipv6"
 	egressDestinationPorts   = "kube-vip.io/egress-destination-ports"
 	egressSourcePorts        = "kube-vip.io/egress-source-ports"
 	activeEndpoint           = "kube-vip.io/active-endpoint"
@@ -142,7 +143,6 @@ func (sm *Manager) addService(svc *v1.Service) error {
 	}
 
 	serviceIPs := fetchServiceAddresses(svc)
-
 	// Check if we need to flush any conntrack connections (due to some dangling conntrack connections)
 	if svc.Annotations[flushContrack] == "true" {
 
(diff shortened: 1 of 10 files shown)


thebsdbox:
cc/ @p-strusiewiczsurmacki-mobica this seems like something you could cast an eye over?

pkg/vip/util.go
```diff
@@ -104,6 +104,24 @@ func IsIPv6(address string) bool {
...
+
+func IsIPv6CIDR(cidr string) bool {
+	_, n, _ := net.ParseCIDR(cidr)
+	if n == nil {
+		return false
+	}
+	_, bits := n.Mask.Size()
+	return bits/8 == net.IPv6len
+}
+
```
p-strusiewiczsurmacki-mobica, line 124: Determining wheteher IP address is v4 or v6 is tricky, but I am not sure if length is always accurate to determine that. Maybe it would be better to just go with `net` package [IP.To4()](https://pkg.go.dev/net#IP.To4) and  [IP.To16()](https://pkg.go.dev/net#IP.To16)  functions?

pkg/manager/services.go
```diff
@@ -162,13 +162,32 @@ func (sm *Manager) addService(svc *v1.Service) error {
...
+		// We will need to modify the iptables rules
+		err = sm.iptablesCheck()
+		if err != nil {
+			log.Errorf("Error configuring egress for loadbalancer [%s]", err)
+		}
+		var podIP string
+		errList := []error{}
+
+		// Should egress be IPv6
+		if svc.Annotations[egressIPv6] == "true" {
```
p-strusiewiczsurmacki-mobica, line 174: With this new annotation kube-vip will check for IPv6 address and if there is no `egressIPv6` annotation it will process service's egress as IPv4.
Wouldn't that effectively disable any support for dualstack services?

p-strusiewiczsurmacki-mobica (commented):
Hi @thebsdbox I've taken a quick look at the code. I think it's really great work, just added 2 comments for further discussion.

pkg/manager/services.go
thebsdbox: This change only affects anything `egress` related, the services/load balancers are all created as usual but if the correct annotations exist then egress rules are created too. The dual stack service test is all passing in the CI/CD.

pkg/vip/util.go
thebsdbox: The `To4()` function is for IP addresses, not CIDR ranges which is the reason I had to add these additional two functions.

Issue #951, opened 2024-09-22
UPNP only forwards one port on services with multiple ports.


l0wl3vel:
**Describe the bug**

When enabling port forwarding on a service with multiple ports exposed only one port gets forwarded.

**To Reproduce**
1. Install kube-vip v.0.8.3 and enable UPNP Forwarding
2. Install ingress-nginx
3. only port 80 gets a port forward

**Expected behavior**
All Ports get forwarded

**Environment (please complete the following information):**
 - OS/Distro: NixOS 24.05
 - Kubernetes Version: 1.30.4+k3s1
 - Kube-vip Version: 0.8.3

**`Kube-vip-values.yaml`:**

```yaml
    config:
      address: "192.168.178.247"
      namespace: kube-vip

    # Check https://kube-vip.io/docs/installation/flags/
    env:
      vip_interface: ""
      vip_arp: "true"
      lb_enable: "true"
      lb_port: "6443"
      vip_cidr: "30"
      cp_enable: "true"
      svc_enable: "true"
      svc_election: "true"
      vip_leaderelection: "true"
      enableUPNP: "true"
```

PR #952, opened 2024-09-22
UPNP: Assume that multiple ports can exist on a service

l0wl3vel:
Fixes UPNP forwarding only the port with the lowest numeric value.

Fixes: #951


thebsdbox:
Can you fix the merge conflict from adding in your previous PR?

Thanks in advance. 👍

l0wl3vel:
@thebsdbox Done

PR #954, opened 2024-09-24
Use IGD:2 and Pinhole Port Forwarding

l0wl3vel:
Implement Pinhole Firewall traversal.

Changes:
- Replace UPNP Implementation with https://github.com/huin/goupnp
- Use WANIPv6FirewallControl to expose Services to the Internet

### Why?

Consumer Routers (in my case a FritzBox) only allow port forwarding for your own device when using the IGD:1 (Internet Gateway Device: A UPNP role allows control over  router and similar devices). You can disable this security feature on other devices, but not on Fritz Boxes, which are quite popular in Germany.

After messing around with it a bit I found out that the IGD:2  WANIPv6FirewallControl control endpoints work much in the same way like port forwarding but my FritzBox does not check the origin of the forwarding request.

We also need to implement Lease Refreshing. It is unclear why it was missing in the previous implementation, since UPNPv2 clearly specifies that a lease time of 0 is not infinity, but invalid. So a perpetual refresher for UPNP forwards is included.

This allows UPNP Port forwarding to work on more routers.

IGD:2 also introduces IPv6 support. But this is out of scope for this PR.

### Open questions

It is unclear if the old IGD:1 implementation is still required. For now I think we should keep both implementations and use the Pinhole style forward first and fall back to Port Mapping if required. IGD:2 is supposed to be backwards compatible and has been around since 2006.

TODO:

- [x] Reimplement IGD:2 Port mapping

Out of scope:

- IPv6 support

Spec: https://www.upnp.org/specs/gw/UPnP-gw-WANIPv6FirewallControl-v1-Service.pdf page 16

### DIFF
--- a/pkg/upnp/upnp.go
+++ b/pkg/upnp/upnp.go
@@ -1,0 +1,120 @@
+package upnp
+
+import (
+	"context"
+	"strings"
+
+	"github.com/huin/goupnp/dcps/internetgateway2"
+	log "github.com/sirupsen/logrus"
+	"golang.org/x/sync/errgroup"
+)
+
+type ConnectionClient interface {
+	AddPortMapping(
+		NewRemoteHost string,
+		NewExternalPort uint16,
+		NewProtocol string,
+		NewInternalPort uint16,
+		NewInternalClient string,
+		NewEnabled bool,
+		NewPortMappingDescription string,
+		NewLeaseDuration uint32,
+	) (err error)
+
+	GetExternalIPAddress() (
+		NewExternalIPAddress string,
+		err error,
+	)
+}
+
+// Gather WANIPv6FirewallControl1 clients in the network and treats errors as non-critical.
+// Use this to create Pinholes in the UPNP IGD2
+func GetWANIPv6FirewallControl1ClientsCtx(ctx context.Context) []internetgateway2.WANIPv6FirewallControl1 {
+	c, errors, err := internetgateway2.NewWANIPv6FirewallControl1ClientsCtx(ctx)
+	var validClients []internetgateway2.WANIPv6FirewallControl1
+	if err != nil {
+		log.Warnf("Unable to query for WAN Firewall Clients [%s]", err.Error())
+	}
+	for _, e := range errors {
+		log.Warnf("UPNP Gateway responded with an error while querying WAN Firewall Client [%s]", e.Error())
+	}
+	for _, c := range c {
+		if c != nil {
+			validClients = append(validClients, *c)
+		}
+	}
+	return validClients
+}
+
+// Gather UPNPConnectionClients in the network and treats errors as non-critical. Use this to find out the external IPs of the network and configure port forwarding
+func GetConnectionClients(ctx context.Context) []ConnectionClient {
+	tasks, _ := errgroup.WithContext(ctx)
+	// This code sucks and I did not manage to make it more concise.
+	//Turns out using []*type instead of []type prevents you from casting to an interface.
+	// If you have any better ideas please take a stab at it.
+	var ip1Clients []*internetgateway2.WANIPConnection1
+	var ip1Error []error
+	tasks.Go(func() error {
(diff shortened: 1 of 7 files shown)


thebsdbox:
This looks great, should I hold off on the other PR whilst this is in motion?

l0wl3vel:
Thanks. Could you please still merge the other PR? They are independent but build to the greater goal of improving the use-case of kube-vip behind consumer routers and NAT setups.

thebsdbox:
Is there any update on this?

thebsdbox:
Was this meant to be closed? 😀

l0wl3vel:
Not really. I was rebasing and pushed the branch without my changes. Github seems to auto-close PRs if the source and destination have the same HEAD

l0wl3vel:
We will have to implement UPNP lease refresh next, since UPNP does not support infinite leases by spec. Right now any UPNP lease only works for an hour, after which kube-vip needs to get restarted to refresh the lease.

 I can't see any requeue logic like [kube-builder controllers use](https://book.kubebuilder.io/getting-started.html#reconciliation-process) anywhere. Are there any plans on implementing something like that globally in kube-vip? Or do you think I should implement reconciler/requeuer just for the UPNP logic? I would really like to hear your thoughts @thebsdbox

thebsdbox:
Will review first thing tomorrow. We've largely just focussed on client-go watchers for most things. The design around moving to kube builder and crds might/probably is a massive change.

l0wl3vel:
Thanks, but no rush. I am going to take another look at dealing with the UPNP clients in a more readable way tomorrow, so I converted to a draft again.

> The design around moving to kube builder and crds might/probably is a massive change.

Can imagine that. Thanks for clarifying

l0wl3vel:
@thebsdbox Could you please trigger a rerun of the pipeline? The tests ran without any issue on my branch.

pkg/upnp/upnp.go
```diff
@@ -0,0 +1,127 @@
...
+			log.Warnf("Unable to find WANIPv6FirewallControl1Clients for Gateway %s [%s]", gatewayURL, err.Error())
+		}
+	}
+	return gatewayClients
+}
+
+// Gather UPNPConnectionClients in the network and treats errors as non-critical. Use this to find out the external IPs of the network and configure port forwarding
+func GetConnectionClients(ctx context.Context) []ConnectionClient {
+	tasks, _ := errgroup.WithContext(ctx)
+	// This code sucks and I did not manage to make it more concise.
```
thebsdbox, line 59: haha, love this comment

pkg/upnp/upnp.go
```diff
@@ -0,0 +1,127 @@
...
+	clients := GetConnectionClients(ctx)
+	gatewayClients := make([]Gateway, len(clients))
+
+	for i := range clients {
+		gatewayClients[i] = Gateway{ConnectionClient: clients[i], WANIPv6FirewallControlClient: nil}
+		gatewayURL := clients[i].GetServiceClient().Location
+		if wanipv6clients, err := internetgateway2.NewWANIPv6FirewallControl1ClientsByURLCtx(ctx, gatewayURL); err == nil {
+			gatewayClients[i].WANIPv6FirewallControlClient = wanipv6clients[0]
+		} else {
+			log.Warnf("Unable to find WANIPv6FirewallControl1Clients for Gateway %s [%s]", gatewayURL, err.Error())
```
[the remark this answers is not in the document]
l0wl3vel: Done

pkg/upnp/upnp.go
l0wl3vel: Yeah, had a lot of fun with the UPNP library. It is really not fun to work with half a dozen auto generated SOAP clients. Unfortunately this seems to be the intended way to do this: https://github.com/huin/goupnp/blob/main/GUIDE.md#internet-gateways

Are you ok with this implementation?

PR #955, opened 2024-09-27
Fixed lastKnownGoodEndpoint not being set when in non-leader-election…

### LINKED ISSUE #948: [Bug] When only relying on Endpoint Events LoadBalancer Service could run out of sync
**Describe the bug**
When a delete event of an `Endpoint` object is not correctly captured by kube-vip or the corresponding kube-vip instance is currently running but the client is not correctly watching events. 
The state between advertised IPs and active `Endpoint` of a `Service` can drift apart.

This can especially happen when you don't run kube-vip with leader election and multiple instances advertise the same IP.

e.g.
* RouteTable setup without leader election for Anycast
* BGP setup without leader election for Anycast

**To Reproduce**
1. Install kube-vip in table mode without leader-election
2. remove rbac for kube-vip to be able to watch endpoints
3. reschedule workload which is backend for a service of type loadBalancer onto a different node.
4. give kube-vip the rbac rights back to be able to watch endpoints.

**Expected behavior**
If we do not catch the event, we should rely on a periodic reconcile of all endpoints on a node.
The schedule for the periodic reconcile should be configurable.
For this specific use-case using RouteTable mode we could use the already implemented option:

``` yaml
#### cleanRoutingTable
env:
- name: vip_cleanroutingtable
  value: 'true'
```

**Screenshots**

 - OS/Distro: Linux
 - Kubernetes Version: [e.g. v.1.18]
 - Kube-vip Version: v0.8.3

**`Kube-vip.yaml`:**
```
apiVersion: apps/v1
kind: DaemonSet
metadata:
  labels:
    app.kubernetes.io/name: kube-vip
    app.kubernetes.io/version: v0.8.3
  name: kube-vip
spec:
  revisionHistoryLimit: 10
  selector:
    matchLabels:
      app.kubernetes.io/name: kube-vip
  template:
    metadata:
      creationTimestamp: null
      labels:
        app.kubernetes.io/name: kube-vip
        app.kubernetes.io/version: v0.8.3
    spec:
      containers:
      - args:
        - manager
        env:
        - name: vip_arp
          value: "false"
        - name: port
          value: "6443"
        - name: vip_interface
          value: lo
        - name: vip_cidr
          value: "32"
        - name: vip_leaderelection
          value: "false"
        - name: svc_enable
          value: "true"
        - name: vip_routingtable
          value: "true"
        - name: vip_cleanroutingtable
          value: "true"
        - name: prometheus_server
          value: :2112
        - name: svc_election
          value: "false"
        - name: disable_service_updates
          value: "true"
        image: ghcr.io/kube-vip/kube-vip:v0.8.3
        imagePullPolicy: Always
        name: kube-vip
        ports:
        - containerPort: 2112
          name: metrics
          protocol: TCP
        resources: {}
        securityContext:
          allowPrivilegeEscalation: false
          capabilities:
            add:
            - NET_ADMIN
            drop:
            - ALL
          readOnlyRootFilesystem: true
          runAsNonRoot: false
      dnsPolicy: ClusterFirst
      hostNetwork: true
      priorityClassName: system-node-critical
      restartPolicy: Always
      securityContext: {}
      serviceAccount: kube-vip
      serviceAccountName: kube-vip
  updateStrategy:
    rollingUpdate:
      maxSurge: 0
      maxUnavailable: 1
    type: RollingUpdate

p-strusiewiczsurmacki-mobica:
… mode

This PR should fix #948 

It seems that when non-leader-election mode was used, `lastKnownGoodEndpoint` was not always updated in case of issues with kubeapi connection.

### DIFF
--- a/pkg/manager/watch_endpoints.go
+++ b/pkg/manager/watch_endpoints.go
@@ -239,19 +239,18 @@ func (sm *Manager) watchEndpoint(ctx context.Context, id string, service *v1.Ser
 							stillExists = true
 						}
 					}
-					// If the last endpoint no longer exists, we cancel our leader Election
-					if !stillExists && leaderElectionActive {
-						if sm.config.EnableServicesElection || sm.config.EnableLeaderElection {
+					// If the last endpoint no longer exists, we cancel our leader Election, and set another endpoint as last known good
+					if !stillExists {
+						if leaderElectionActive && (sm.config.EnableServicesElection || sm.config.EnableLeaderElection) {
 							log.Warnf("[%s] existing [%s] has been removed, restarting leaderElection", provider.getLabel(), lastKnownGoodEndpoint)
 							// Stop the existing leaderElection
 							cancel()
+							// disable last leaderElection flag
+							leaderElectionActive = false
 						}
 						// Set our active endpoint to an existing one
 						lastKnownGoodEndpoint = endpoints[0]
-						// disable last leaderElection flag
-						leaderElectionActive = false
 					}
-
 				} else {
 					lastKnownGoodEndpoint = endpoints[0]
 				}


Cellebyte:
@p-strusiewiczsurmacki-mobica are you still able to reproduce the bug with your changed code?

p-strusiewiczsurmacki-mobica:
@Cellebyte No, as far as I can test it, after this change I didn't see this behavior occurring.

Issue #922, opened 2024-08-21
When using kube-vip controlplane VIP on lo interface it has host scope


tenyo:
**Is your feature request related to a problem? Please describe.**
When using controlplane VIP on the loopback `lo` interface it gets created with a scope of `host` instead of `global`.
One problem in our case is that causes Cilium to ignore that IP and not consider it in its firewall policies.

**Describe the solution you'd like**
An ability to specify `host` or `global` scope, or just default to `global` when using the loopback.

**Additional context**

Using kube-vip v0.8.2

Example of current lo config with the kube-vip VIP:
```
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet 150.50.50.5/32 scope host lo
       valid_lft forever preferred_lft forever
```

Kube-vip installed with:
```
        ctr run --rm --net-host ghcr.io/kube-vip/kube-vip vip /kube-vip manifest pod \
          --interface "lo" \
          --vip "150.50.50.5" \
          --controlplane \
          --leaderElection true
```

tenyo:
I believe this is currently being set here: https://github.com/kube-vip/kube-vip/blob/main/pkg/vip/address.go#L108-L111

PR #956, opened 2024-09-27
Add loInterfaceGlobalScope flag for enabling global scope for vip lo interface

tenyo:
Fixes https://github.com/kube-vip/kube-vip/issues/922

This PR adds a new flag to allow for overriding the current default behavior when creating a kube VIP on the localhost (lo) interface, which automatically sets the scope to `host`. Enabling the new `loInterfaceGlobalScope` flag will prevent kube-vip from setting a `host` scope and will instead use the `global` scope.


thebsdbox (approved):
Looks great, I was wondering if it would make sense to have a scope setting via a string “host” / “global” etc. in case of other scopes? If not this is great.

tenyo:
> Looks great, I was wondering if it would make sense to have a scope setting via a string “host” / “global” etc. in case of other scopes? If not this is great.

I thought about it but I don't believe any of the other scopes would ever be used so went for the simpler boolean.

thebsdbox:
This will be available in the dev image built at midnight, I'll try and get a release out soon.

Issue #831, opened 2024-04-23
Support table mode for CP mode


MaxRink:
**Is your feature request related to a problem? Please describe.**
Currently only BGP, APR and WG modes are implemented for the CP mode.

**Describe the solution you'd like**
We would like to have the table mode with an healthcheck of the local API-Server on each node, so that we can utilize ECMP Load Balancing for the API-Server

thebsdbox:
Makes sense, I don't think it would be a large amount of effort... all the functions are there, they just need wiring together! 🔌

Cellebyte:
Done

PR #957, opened 2024-10-01
Routing Table mode for Control Plane

p-strusiewiczsurmacki-mobica:
This PR introduces Routing Table mode for Control Plane mode as described in #831.

### DIFF
--- a/pkg/backend/backend.go
+++ b/pkg/backend/backend.go
@@ -1,0 +1,69 @@
+package backend
+
+import (
+	"fmt"
+	"time"
+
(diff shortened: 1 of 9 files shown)


thebsdbox:
Amazing stuff! I presume this is going to be a little difficult for an e2e test? Although it appears not be impacting the existing ARP functionality.

p-strusiewiczsurmacki-mobica:
The only test I can think of would be to deploy kube-vip, create service and check if IP address and route were added using `docker exec` to be honest.

thebsdbox (approved):
Looks good to me, thanks for this addition!

PR #926, opened 2024-08-23
Auto network interface discovery for services

p-strusiewiczsurmacki-mobica:
This PR fixes #903 

It introduces feature to automatically discover proper networking interface for the service based on the service's loadbalancer's IP.

To enable this feature `vip_servicesinterface` should be set to value `auto` (to enable it globally) or `kube-vip.io/serviceInterface` annotation should be set to `auto` to enable this for selected service.

PR #961, opened 2024-10-09
Added auto subnet discovery for services in ARP mode

p-strusiewiczsurmacki-mobica:
This PR adds auto subnet discovery for services deployed in ARP mode.

As for now all service VIP addresses have subnet of /32 and /128 for IPv4 and IPv6 respectively, if `vip_cidr` is not defined. After this changes if `vip_address` was not provided in config and `vip_cidr` will be set to `auto,auto` (first value is for IPv4, second for IPv6 addresses) kube-vip will try to find suitable subnet based on addresses available on the `vip_servicesinterface`. This is complementary to #926 . 

Example:

Config:
vip_servicesinterface: "eth0"
vip_cidr: "auto,auto"

Interface on the node:

```
14: eth0@if15: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP group default 
    link/ether 02:42:ac:12:00:04 brd ff:ff:ff:ff:ff:ff link-netnsid 0
    inet 172.18.0.4/16 brd 172.18.255.255 scope global eth0
    inet 192.0.3.103/17 scope global eth0
```

Service configured for address 192.168.3.200
```yaml
annotations:
    kube-vip.io/loadbalancerIPs: "192.0.3.200"
```

As an address `192.0.3.200` is in the same subnet as `192.0.3.103/17`, the configured IP address will be `192.0.3.200/17`.

### DIFF
--- a/cmd/kube-vip.go
+++ b/cmd/kube-vip.go
@@ -289,6 +289,14 @@ var kubeVipManager = &cobra.Command{
 			log.Fatalln("no features are enabled")
 		}
 
+		if !initConfig.EnableARP && strings.Contains(initConfig.VIPCIDR, kubevip.Auto) {
+			log.Fatalln("auto subnet discovery cannot be used outside ARP mode")
+		}
+
+		if strings.Contains(initConfig.VIPCIDR, kubevip.Auto) && initConfig.Address != "" {
+			log.Fatalln("auto subnet discovery cannot be used if VIP address was provided")
+		}
+
 		// If we're using wireguard then all traffic goes through the wg0 interface
 		if initConfig.EnableWireguard {
 			if initConfig.Interface == "" {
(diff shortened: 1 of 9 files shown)


Cellebyte:
@p-strusiewiczsurmacki-mobica maybe we need a rebase here and I think this should be verified as well for the next release.

thebsdbox:
Would be nice to have this in the upcoming release, is there a timeline for this?

Cellebyte:
I think a final validation tomorrow should be enough. We would just need a rebase :)

thebsdbox:
I'll create a PR to bump the version number in the makefile.

thebsdbox (approved):
This looks great, I'm starting to feel like we're slowly becoming an IP parsing project with IP address allocation on top 😅

PR #993, opened 2024-11-12
Services aren't removed when there are no endpoints.

thebsdbox:
For some reason the endpoint watcher had a check that meant that when "local" endpoints went to 0 then it would stop being part of the leader election, however it appears that it was wrapped with only being allowed when routing mode was enabled. This has broken a few things.

This make sense @p-strusiewiczsurmacki-mobica

### DIFF
--- a/pkg/manager/watch_endpoints.go
+++ b/pkg/manager/watch_endpoints.go
@@ -373,7 +373,7 @@ func (sm *Manager) watchEndpoint(ctx context.Context, id string, service *v1.Ser
 				}
 
 				// If there are no local endpoints, and we had one then remove it and stop the leaderElection
-				if lastKnownGoodEndpoint != "" /*&& sm.config.EnableRoutingTable*/ {
+				if lastKnownGoodEndpoint != "" {
 					log.Warnf("[%s] existing [%s] has been removed, no remaining endpoints for leaderElection", provider.getLabel(), lastKnownGoodEndpoint)
 					if err := sm.TeardownEgress(lastKnownGoodEndpoint, service.Spec.LoadBalancerIP,
 						service.Annotations[egressDestinationPorts], service.Namespace); err != nil {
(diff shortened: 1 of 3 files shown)


pkg/manager/watch_endpoints.go
```diff
@@ -373,7 +373,7 @@ func (sm *Manager) watchEndpoint(ctx context.Context, id string, service *v1.Ser
 				}
 
 				// If there are no local endpoints, and we had one then remove it and stop the leaderElection
-				if lastKnownGoodEndpoint != "" && sm.config.EnableRoutingTable {
+				if lastKnownGoodEndpoint != "" /*&& sm.config.EnableRoutingTable*/ {
```
thebsdbox, line 376: This shouldn't be gated around routing table mode? @p-strusiewiczsurmacki-mobica

Cellebyte: I agree.
I think the idea was to hide it behind the table mode feature flag back then.
But after the code got refactored this snippet was just missed to adapt to the new call stack.

thebsdbox: OK cool, I'll clean the code and add a final commit. I'll take an action to refactor the CI/CD testing to create more errors in future to catch this.

Cellebyte:
@thebsdbox we reproduced your issue in one of our cluster.
And the changes here are fixing the issue.

Cellebyte:
@p-strusiewiczsurmacki-mobica maybe you can elaborate, why you added the check here.

p-strusiewiczsurmacki-mobica:
@Cellebyte @thebsdbox I think I wanted to run `sm.TeardownEgress` only in routing table mode and didn't think that through.

Anyway, shouldn't be this check performed in additional if (inside `if lastKnownGoodEndpoint != ""`) to not run  `sm.TeardownEgress` in BGP mode?

Cellebyte:
@p-strusiewiczsurmacki-mobica egress is not dependent on the node. But rather if it is enabled for the service we are reconciling.

Issue #988, opened 2024-11-08
MP-BGP support


chdxD1:
**Is your feature request related to a problem? Please describe.**
In some cases one might want to have a single IPv4 or IPv6 session and use multi-protocol BGP to advertise both address families (IPv4 and IPv6) over the same session.

**Describe the solution you'd like**
gobgp support multi-protocol BGP, adding the address families to the peer configuration is already enough to exchange both route types.

**Describe alternatives you've considered**
Separate sessions for IPv4 and IPv6 (but might not be desired)
 
**Additional context**
One issue with IP4 over IPv6 sessions (or IPv4 over IPv6 sessions) is the default next-hop mapping for the other address family that is not used to establish the session:

MP-BGP on IPv4 session
```
     Network          Next Hop            Metric LocPrf Weight Path
 *>  192.168.100.1/32 100.64.1.2                             0 64512 ?
 *>  2a01::1/128      ::ffff:6440:102                        0 64512 ?
```
MP-BGP on IPv6 session
```
     Network          Next Hop            Metric LocPrf Weight Path
 *>  192.168.100.1/32 2a01:10::2                             0 64512 ?
 *>  2a01::1/128      2a01:10::2                             0 64512 ?
```

In some cases, given that the worker nodes are dualstack (have an IPv4 and IPv6 address) it might be of interest to apply a policy on outgoing announcements to overwrite with the respective interface address. This would require a larger extension of the BGP functionality of kube-vip (and maybe even auto discovering the other address families IP address) to gain something like:

MP-BGP on IPv4 session (with policy)
```
     Network          Next Hop            Metric LocPrf Weight Path
 *>  192.168.100.1/32 100.64.1.2                             0 64512 ?
 *>  2a01::1/128      2a01:10::2                             0 64512 ?
```
MP-BGP on IPv6 session (with policy)
```
     Network          Next Hop            Metric LocPrf Weight Path
 *>  192.168.100.1/32 100.64.1.2                             0 64512 ?
 *>  2a01::1/128      2a01:10::2                             0 64512 ?
```

chdxD1:
For awareness: I did a small PoC with an adapter kube-vip version to validate this functionality with gobgp but it's a bit messy: https://github.com/kube-vip/kube-vip/commit/70798b8e9f898f22ed73fca31b4dec6ba2084dcb

thebsdbox:
Very nice !

chdxD1:
[...]

PR #1003, opened 2024-11-28
Support for MP-BGP

p-strusiewiczsurmacki-mobica:
This should implement #988

### DIFF
--- a/demo/server/main.go
+++ b/demo/server/main.go
@@ -60,7 +60,7 @@ func main() {
 				fmt.Println("error: ", err)
 			}
 
-			ServerConn.WriteTo(buf[0:n])
+			ServerConn.WriteTo(buf[0:n], addr)
 		}
 	}
 }
--- a/pkg/bgp/peers.go
+++ b/pkg/bgp/peers.go
@@ -10,6 +10,7 @@ import (
 	"github.com/golang/protobuf/ptypes" //nolint
 	"github.com/golang/protobuf/ptypes/any"
 	api "github.com/osrg/gobgp/v3/api"
+	"github.com/vishvananda/netlink"
 )
 
 // AddPeer will add peers to the BGP configuration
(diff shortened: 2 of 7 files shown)


thebsdbox:
Looks like it is coming along 😄 @p-strusiewiczsurmacki-mobica / @chdxD1

Cellebyte:
@chdxD1 could you review? :)

thebsdbox:
What is left on this PR?

Cellebyte:
@thebsdbox  I think a review is left from @chdxD1

p-strusiewiczsurmacki-mobica:
> @thebsdbox I think a review is left from @chdxD1

Not really. I still can't get this to work. :( Hence it's a draft. As soon as it will be read for a review I'll open the PR.

p-strusiewiczsurmacki-mobica:
Hi @thebsdbox , @Cellebyte , @chdxD1 - I think I've finally got it to work, so I am opening PR for the review.

@chdxD1 , can you test if this works as you expected? I believe you have much more experience with stuff like BGP and FRR than me.
 
As for configuration, currently global configuration can be set using `mpbgp_nexthop`, `mpbgp_ipv4` and `mpbgp_ipv6` flags is possible, but it can be overridden with peer-specific config using `bgp_peers` flag as in the example below:
 
 ```
        - name: mpbgp_nexthop
          value: "fixed"
        - name: mpbgp_ipv4
          value: "192.168.200.200"
        - name: mpbgp_ipv6
          value: "4f46:52c1:42ee:5394:e75b:bd47:abef:c168"
        - name: bgp_peers
          value: "[fd00:1111:2222:3333:c7d9:7235:6bf7:5d52]:65501::false/mpbgp_nexthop=auto_sourceif"
 ```
 
Now, I believe we would like to use CRDs for configuration instead of this. Should adding CRDs be part of this PR, or should I create another PR for that? What do you think?

thebsdbox:
As much as I'd like to move to CRDs, I think we probably should save that for a later day. From my limited understanding the PR looks good, there are a bunch of helper functions `GetInterfaceByIP()` etc.. would it make sense to bundle them with the other helper functions in the `vip` pkg? Just thinking of sprawling helper functions throughout the code base.

thebsdbox (approved):
This all looks good to me, as you're the primary users of this functionality I'm happy to give the go/no-go to yourselves on this PR.

PR #1042, opened 2025-01-15
Move away from logrus

thebsdbox:
This replaces the usage of logrus with a different logging library.

### DIFF
--- a/cmd/kube-vip.go
+++ b/cmd/kube-vip.go
@@ -9,9 +9,9 @@ import (
 	"strings"
 	"time"
 
+	log "github.com/gookit/slog"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
-	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 	"github.com/vishvananda/netlink"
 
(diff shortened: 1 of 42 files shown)


cmd/kube-vip.go
```diff
@@ -110,7 +110,7 @@ func init() {
 	kubeVipCmd.PersistentFlags().StringVarP(&initConfig.Namespace, "namespace", "n", "kube-system", "The namespace for the configmap defined within the cluster")
 
 	// Manage logging
-	kubeVipCmd.PersistentFlags().Uint32Var(&logLevel, "log", 4, "Set the level of logging")
+	kubeVipCmd.PersistentFlags().Uint32Var(&logLevel, "log", 600, "Set the level of logging")
```
p-strusiewiczsurmacki-mobica, line 113: Wouldn't that break (or rather change the output of) logging in existing deployments without any level translation layer? It seems that values that represent levels in `gookit/slog` are different than those in `logrus` so, for example, existing deployment with `vip_loglevel` set to 5 (currently `info` if I recall correctly) will result in different logging if someone will update the image without updating the `vip_loglevel` value.

p-strusiewiczsurmacki-mobica (commented):
The changes seems to look great overall. I am just worried about logging levels I've mentioned in the other comment.

One thing I am not sure is if `gookit/slog` is the best choice here. It seems to be nice, compact logging package, but as kube-vip is moving from logrus because it is no longer actively maintained, I would consider to use something more widely used like `zap`, `zerolog` or maybe the `log/slog` which was introduced to the standard library in go 1.21, to be sure the logging library wouldn't have to be changed again anytime soon. But that's just a proposition anyway.

thebsdbox:
I liked the output format, especially the file/line number. Presumably I can do this with log/slog and some formatting?

thebsdbox:
Will close this pull request in favour of `log/slog`

PR #1049, opened 2025-01-22
MP-BGP CRD support

p-strusiewiczsurmacki-mobica:
This PR is followup for #1003 and is build upon it.

It adds project structure for `kubebuilder` which enables developers to easily generate CRDs for required API objects. Simple CRD for BGP peers has been added as a PoC.

For now the BGPPeer CRD will only override the usual MP-BGP config, so the CRD objects have to be deployed before `kube-vip`'s start. I believe that moving to more operator-like approach with continuous reconciliation could be part of separate PR based on this one.


thebsdbox:
image

Cellebyte:
I will close this for now, until we have the prerequisites in place to deal with a CRD

PR #1073, opened 2025-02-10
Fixed ENV log level setting

p-strusiewiczsurmacki-mobica:
During investigating another issue I found a minor bug:

`log.SetLogLoggerLevel` is only called  in `init()` and `vip_loglevel` env var is never used so is it impossible to use it to change logging level. This should fix that.


Cellebyte (commented):
LGTM

PR #1079, opened 2025-02-17
This adds a health check

thebsdbox:
The env variable `health_check_port` as long an `!= 0` and `> 1024` will enable a health check on `/healthz`.. be aware that kube-vip uses `hostNetwork` so it would be easy for a port conflict.


pkg/kubevip/config_generator.go
```diff
@@ -528,6 +528,16 @@ func generatePodSpec(c *Config, imageVersion string, inCluster bool) *corev1.Pod
 		newEnvironment = append(newEnvironment, mdif...)
 	}
 
+	if c.HealthCheckPort != 0 {
+		healthPort := []corev1.EnvVar{
```
Mimoni44: Fox com

Issue #953, opened 2024-09-23
changes to services are not handled


CrimsonFez:
**Describe the bug**
When changes are made to the Service, those changes are not processed by kube-vip. This only applies to Services after they have been allocated a LoadBalancer.
For example, if you set the LBIP on a service, then change the annotation, it does not allocate the new LBIP.
Additionally, if you change the servicePorts it does not update status.loadBalancer.

**To Reproduce**
1. deploy v0.8.3 with LoadBalancer ARP mode enabled
2. create service
```
apiVersion: v1
kind: Service
metadata:
  name: test
  annotations:
    kube-vip.io/loadbalancerIPs: "192.168.101.1"
spec:
  selector:
    app: myapp
  ports:
  - port: 80
    targetPort: 80
```
3. change annotation or ports on service
4. no logs of change in kube-vip and no reflection of change in service status

**Expected behavior**
I would expect that when changes are made to a service after it has been made a LoadBalancer that they are respected.

**Environment (please complete the following information):**
 - OS/Distro:Debian 12
 - Kubernetes Version: v1.30
 - Kube-vip Version: v0.8.3

**`Kube-vip.yaml`:**
```yaml
apiVersion: apps/v1
kind: DaemonSet
metadata:
  name: kube-vip-lb
  namespace: kube-system
spec:
  selector:
    matchLabels:
      app: kube-vip-lb
  template:
    metadata:
      labels:
        app: kube-vip-lb
    spec:
      containers:
      - image: ghcr.io/kube-vip/kube-vip-iptables:v0.8.3
        imagePullPolicy: Always
        name: kube-vip
        args:
        - manager
        env:
        - name: vip_arp
          value: "true"
        - name: vip_cidr
          value: "32"
        - name: vip_servicesinterface
          value: dummy0
        - name: svc_enable
          value: "true"
        - name: svc_leasename
          value: kube-vip-lb
        - name: svc_election
          value: "false"
        - name: lb_class_only
          value: "false"
        - name: lb_class_name
          value: kube-vip
        - name: enable_service_security
          value: "false"
        - name: prometheus_server
          value: :2113
        resources: {}
        securityContext:
          capabilities:
            add:
            - NET_ADMIN
            - NET_RAW
        terminationMessagePath: /dev/termination-log
        terminationMessagePolicy: File
      - image: ghcr.io/404incorporated/kube-vip-bird2:v1.1.0
        imagePullPolicy: Always
        name: bird
        resources: {}
        securityContext:
[...]

PR #1094, opened 2025-03-07
Fixed service update in BGP and ARP mode

### LINKED ISSUE #998: Changing the IP of LB isn't detected properly by kube-vip
Tested with v0.8.7. Using ArgoCD to sync resources, not sure if that matters. Happy to provide more info if this is hard to repro.

When loadBalancerIP is changed in existing resource of type LoadBalancer, with loadBalancerClass and loadBalancerIP set, kube-vip does not "see" the new IP. 

It does detect that the resource has been updated, but it reports the old IP and not the new one.

```
(svcs) already found existing address [1.1.1.old] on adapter [eth1]
Re-applying the VIP configuration [1.1.1.old] to the interface [eth1]
```

Additionally, if I restart kube-vip, it still doesn't see the new IP on the service. How does it even know the old IP? Maybe it's stored within the lease? or is it coming out of status.loadBalancer.ingress.ip of the LB?

When I delete the LB resource:
```
[(svcs) already found existing address [[1.1.1.old] on adapter [eth1]
[LOADBALANCER] Stopping load balancer
"[VIP] Releasing the Virtual IP [[1.1.1.old]
Removed [fe766212-7fec-4936-83f7-1dcc853f7a55] from manager, [2] advertised services remain
(svcs) [lbname] has been deleted
[endpoints] stopping watching for [lbname] in namespace [ns]
(svc election) service [name] leader lost: [node3]
(svc election) for service [name] stopping"]      
```  
-> good

when i now recreate the LB (with the new IP):
`(svcs) adding VIP [1.1.1.new] via eth1 for [lbname]"    `
-> that works, too

p-strusiewiczsurmacki-mobica:
This **should** fix #998 and #953

While working on the BGP changes I've noticed that services are not properly updated in BGP mode, so decided to give it a go. Tested this in BGP, ARP and RT modes and it seems to work just fine now.

Main changes:

- Added missing status reset to the existing `event.Modified` check.
- Changed the way `fetchServiceAddresses` works - it will now check if IPv4 or IPv6 provided in `s.Spec.LoadBalancerIP` is different from the same family IP address in `s.Status.LoadBalancer.Ingress` and if so will use it to update the address.
- Added mutex for service leader election - in case of update it sometimes happened that the restarted leader (on the same node) tried to acquire service lease before the stopped leader was able to fully step down. This seems to prevent that.

### DIFF
--- a/pkg/manager/services.go
+++ b/pkg/manager/services.go
@@ -3,6 +3,7 @@ package manager
 import (
 	"context"
 	"fmt"
+	"net"
 	"slices"
 	"strings"
 	"time"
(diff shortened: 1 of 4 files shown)


thebsdbox (approved):
Looks good! The logic was originally meant to be pretty simple, and it's grown a little out of control. Thanks for this fix!

Issue #895, opened 2024-07-13
Control Plane Load-Balancing issue with IPVS


Cloud-Mak:
**Describe the bug**
After setting up HA k8s with 3 nodes when tried enabling [LB for control planes](https://kube-vip.io/docs/about/architecture/#control-plane-load-balancing), the ipvsadm is showing odd beheviour. it was only showing 1st master node m1 (which was initiated 1st in k8s) avialable as backend for IPVS LB. Even though lb_enable was set for rest of nodes (m2 & m3) kube-vip.yaml

m1 ipvsadm -ln 

```
IP Virtual Server version 1.2.1 (size=4096)
Prot LocalAddress:Port Scheduler Flags
  -> RemoteAddress:Port           Forward Weight ActiveConn InActConn
TCP  172.30.84.160:6443 rr
  -> 172.30.84.154:6443           Local   1      0          0

```

After rebooting m1 (.154), the VIP shifted to m3 (new leader) and ipvs command on m3 shown only two nodes available as m2 (.155) & m3 (.156) backend for IPVS.  m1 is not getting added to backend for IPVS even after it comes back online.

Here are kube-VIP logs for m3 while m1 is rebooting.

```
time="2024-07-12T16:27:55Z" level=info msg="IPVS Loadbalancer enabled for 1.2.1"
time="2024-07-12T16:27:55Z" level=info msg="Kube-Vip is watching nodes for control-plane labels"
time="2024-07-12T16:27:55Z" level=info msg="Gratuitous Arp broadcast will repeat every 3 seconds for [172.30.84.160/ens160]"
time="2024-07-12T16:27:55Z" level=error msg="Error querying backends file does not exist"
time="2024-07-12T16:27:55Z" level=info msg="Created Load-Balancer services on [172.30.84.160:6443]"
time="2024-07-12T16:27:55Z" level=info msg="Added backend for [172.30.84.160:6443] on [172.30.84.155:6443]"
time="2024-07-12T16:27:55Z" level=info msg="failed check k8s server version: Get \"https://172.30.84.154:6443/version?timeout=10s\": dial tcp 172.30.84.154:6443: connect: connection refused"
time="2024-07-12T16:27:55Z" level=info msg="Added backend for [172.30.84.160:6443] on [172.30.84.156:6443]"
```

Eventually I had to stop kube-vip container on m3 (via crictl stop ID), it created a new kube-vip container immediately and then m1 appeard as background of IPVS.

```
[...]

PR #1096, opened 2025-03-10
Fixed IPVS deletion issues

p-strusiewiczsurmacki-mobica:
I think this should fix the followup issue described in [the comment](https://github.com/kube-vip/kube-vip/issues/895#issuecomment-2702888317) in #895

Upon investigation it turned out that when kube-vip tries to delete IPVS loadbalancer it tries to close `nil` channel which results in panic and pre-mature exit. So I added initialization for this channel.

```
2025/03/10 12:25:02 INFO IPVS Loadbalancer enabled version=" 1.2.1"
2025/03/10 12:25:02 INFO sysctl set net.ipv4.vs.conntrack to 1
2025/03/10 12:25:02 INFO sysctl set net.ipv4.ip_forward to 1
2025/03/10 12:25:02 INFO layer 2 broadcaster starting
2025/03/10 12:25:02 DEBUG layer 2 update ip=10.0.1.10 interface=eth0 ms=3000
2025/03/10 12:25:02 INFO Kube-Vip is watching nodes for control-plane labels
2025/03/10 12:27:29 INFO Received termination, signaling cluster shutdown
panic: close of nil channel

goroutine 154 [running]:
github.com/kube-vip/kube-vip/pkg/loadbalancer.(*IPVSLoadBalancer).RemoveIPVSLB(0xc0003e4180)
	/src/pkg/loadbalancer/ipvs.go:139 +0x2b
github.com/kube-vip/kube-vip/pkg/cluster.(*Cluster).vipService.func2()
	/src/pkg/cluster/service.go:120 +0x39
created by github.com/kube-vip/kube-vip/pkg/cluster.(*Cluster).vipService in goroutine 89
	/src/pkg/cluster/service.go:118 +0xa2b
```

It also turned out that 2 goroutines were waiting for an OS signal on the same channel. This caused kube-vip to not remove IPVS loadbalancer on exit sometimes (depending on which goroutine got the signal first). Changed that so goroutines should always exit in proper order.

EDIT: It seems that backends were not watched if LB forwarding method was not set as `masquerade`. I've decided to enable backend watching for all forwarding methods as I believe this should be not dependent on the mode. Additionally I've added a piece of code hat will remove a backend from the watch if node will enter `NotReady` state.

### DIFF
--- a/pkg/backend/backend.go
+++ b/pkg/backend/backend.go
@@ -14,8 +14,9 @@ import (
 )
 
 type Entry struct {
-	Addr string
-	Port uint16
+	Addr    string
+	Port    uint16
+	IsLocal bool
 }
 
 type Map map[Entry]bool
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -50,7 +50,8 @@ func startNetworking(c *kubevip.Config) ([]vip.Network, error) {
 
 	networks := []vip.Network{}
 	for _, addr := range addresses {
-		network, err := vip.NewConfig(addr, c.Interface, c.LoInterfaceGlobalScope, c.VIPSubnet, c.DDNS, c.RoutingTableID, c.RoutingTableType, c.RoutingProtocol, c.DNSMode, c.LoadBalancerForwardingMethod, c.IptablesBackend)
+		network, err := vip.NewConfig(addr, c.Interface, c.LoInterfaceGlobalScope, c.VIPSubnet, c.DDNS, c.RoutingTableID,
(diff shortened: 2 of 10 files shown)


Cellebyte:
@p-strusiewiczsurmacki-mobica did you do a test of ipv6 or ipv4+ipv6?

p-strusiewiczsurmacki-mobica:
> @p-strusiewiczsurmacki-mobica did you do a test of ipv6 or ipv4+ipv6?

Not yet, I'll try to do that and will report back.

p-strusiewiczsurmacki-mobica:
@Cellebyte I had to add a few changes but it seems that IPv6 and dualstack clusters should now be good to go with IPVS loadbalancing. :)

Cellebyte (approved):
This code was validate from users in #895 and the ipv6 path was tested by PR maintainer.
The code looks good and we are working on refactoring it anyways.

PR #1109, opened 2025-03-20
Moved endpoint-related code from pkg/manager to pkg/endpoints 

p-strusiewiczsurmacki-mobica:
This PR is a continuation of #1103 

As I think that `pkg/manager` is a bit too big I've decided to move endpoint-related functions to separate `pkg/endpoints`.


Cellebyte:
@p-strusiewiczsurmacki-mobica I like the idea of moving the reconcile methods for CRs into their own package and then implement the reconciliation loop based on the mode configured in kube-vip.

p-strusiewiczsurmacki-mobica:
@Cellebyte this should be ready :)

pkg/manager/watch_endpoints.go
```diff
@@ -183,43 +60,81 @@ func (sm *Manager) watchEndpoint(svcCtx *serviceContext, id string, service *v1.
...
+	return errs
+}
+
+func (sm *Manager) clearBGPHosts(service *v1.Service) {
+	if instance := sm.findServiceInstance(service); instance != nil {
+		sm.clearBGPHostsByInstance(instance)
+	}
+}
+
+func (sm *Manager) clearBGPHostsByInstance(instance *svcs.Instance) {
```
Cellebyte, line 127: this is the cleaner per service, right?

pkg/endpoints/endpoints_bgp.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	return nil
+}
+
+func (b *BGP) deleteAction(service *v1.Service) {
+	b.ClearBGPHosts(service)
+}
+
+func (b *BGP) ClearBGPHosts(service *v1.Service) {
```
Cellebyte, line 97: as the service is similar to the endpoint here ^^

pkg/manager/watch_endpoints.go
p-strusiewiczsurmacki-mobica: this is fixed in #1121

pkg/endpoints/endpoints_bgp.go
p-strusiewiczsurmacki-mobica: this is fixed in #1121

PR #1126, opened 2025-04-09
Added support for RBAC manifest generation

Cellebyte:
* Fixed the currently broken `manager manifest rbac` command.
* It now got 2 new flags
** `--role` which creates a `Role` and `RoleBinding` in the same namespace as the `kube-vip` DaemonSet/StaticPod
** `--rolebinding` which creates a `ClusterRole` cluster-wide and a `RoleBinding` in the same namespace as the `kube-vip` DaemonSet/StaticPod

If non of these two flags are defined it creates `ClusterRoleBinding` and `ClusterRole`.


thebsdbox (approved):
`/lgtm`

PR #1128, opened 2025-04-14
Fixed ARP multiple references issues

p-strusiewiczsurmacki-mobica:


pkg/manager/watch_services.go
```diff
@@ -217,7 +217,7 @@ func (sm *Manager) servicesWatcher(ctx context.Context, serviceFunc func(context
 								}
 							}()
 
-							if (sm.config.EnableRoutingTable || sm.config.EnableBGP) && (!sm.config.EnableLeaderElection && !sm.config.EnableServicesElection) {
+							if (sm.config.EnableARP) || (sm.config.EnableRoutingTable || sm.config.EnableBGP) && (!sm.config.EnableLeaderElection && !sm.config.EnableServicesElection) {
```
Cellebyte, line 220: 🦆 
I think ARP on more then one endpoint could be very bad ❌

p-strusiewiczsurmacki-mobica: Anyway, I think that after I've added separate `ARP manager` we do not need that here anymore, so removed that.

PR #1141, opened 2025-04-26
fix: s/endoints/endpoints/ in RBAC manifest generation code

sdwilsh:
While moving over to using the RBAC manifest generator, this stood out as a clear typo.


Cellebyte (approved):
LGTM

PR #1146, opened 2025-05-01
Replace invalid verb "endpoints" with "update"

nprzy:
As written, the RBAC manifest generator produces a rule that looks like:

```yaml
- apiGroups:
  - ""
  resources:
  - services
  - endpoints
  verbs:
  - list
  - get
  - watch
  - endpoints
```

The very last line of this snippet, "endpoints", doesn't seem correct. This can be checked with the command `kubectl api-resources --verbs=endpoints`, which will show all resource types for which `endpoints` is a valid verb. In my cluster none exist. Based on the hand-written manifest published [here](https://github.com/kube-vip/website/blob/main/content/manifests/rbac.yaml#L19), I think this is supposed to be "update". This PR adjusts the RBAC manifest generation accordingly.


Cellebyte (approved):
LGTM

PR #1166, opened 2025-05-29
Fix/map concurrency

### LINKED ISSUE #1164: [v0.9.1] fatal error: concurrent map read and map write
On a cluster with around 150 LoadBalancer services and 80 nodes we experience kube-vip crashs with error `fatal error: concurrent map read and map write`, worked so far with 0.8.7

**To Reproduce**
1. Upgrade kube-vip 0.8.7 to 0.9.1
2. Replace worker nodes with upgrade Kubernetes 1.30.7 -> 1.31.7
3. kube-vip pod crashed after a while

**config**

<details>

```yaml
    env:
    - name: lb_class_legacy_handling
      value: "false"
    - name: enable_endpointslices
      value: "true"
    - name: lb_class_name
    - name: prometheus_server
      value: :2112
    - name: disable_service_updates
      value: "true"
    - name: vip_nodename
      valueFrom:
        fieldRef:
          apiVersion: v1
          fieldPath: spec.nodeName
    - name: vip_arp
      value: "false"
    - name: port
      value: "6443"
    - name: vip_interface
      value: lo
    - name: vip_subnet
      value: 32,128
    - name: vip_leaderelection
      value: "false"
    - name: svc_enable
      value: "true"
    - name: vip_routingtable
      value: "true"
    - name: vip_cleanroutingtable
      value: "true"
    - name: svc_election
      value: "false"
```
</details>

**Expected behavior**
service running is expected to happen.

**logs**

attached

[kube-vip.log](https://github.com/user-attachments/files/20406476/kube-vip.log)

**Environment: **
 - OS/Distro: Ubuntu 24.04
 - Kubernetes Version: v.1.31.7
 - Kube-vip Version: 0.9.1

Cellebyte:
* this should fix #1164


Makefile
```diff
@@ -5,7 +5,7 @@ TARGET := kube-vip
 .DEFAULT_GOAL := $(TARGET)
 
 # These will be provided to the target
-VERSION := v0.9.1
+VERSION := v0.9.2
```
thebsdbox: Can we bump the makefile in a separate PR, I've some additional changes I'll need to add before we cut a new release. (mainly a cosmetic thing around kernel modules)

pkg/manager/services.go
```diff
@@ -86,6 +113,10 @@ func comparePortsAndPortStatuses(svc *v1.Service) bool {
 }
 
 func (sm *Manager) addService(ctx context.Context, svc *v1.Service) error {
+	// protect against addService whil reading
```
thebsdbox: `whil` -> `while`

thebsdbox (approved):
Looks good, I like the cleanup of the maps for keeping track of active services! :)

Makefile
Cellebyte: Done

pkg/manager/services.go
Cellebyte: done

PR #1169, opened 2025-06-06
Small change to reduce pointless errors

thebsdbox:
> Currently the following 3 kernel modules are mandatory for kube-vip to startup -
iptable_mangle
iptable_nat
iptable_filter

and 

> Based on the above, and the analysis that the customer did on the code (ervice_egress.go) it seems that kube-vip won't be able to work without the above modules. 
Also taking into consideration the feedback from Ubuntu, these modules are considered obsolete in case nftables is used.

The error reported is cosmetic as kube-vip will still start with missing modules. But now if nftables is explicitly used then this test is skipped.


Cellebyte (approved):
LGTM

PR #1174, opened 2025-06-10
Fixed service-tests + minor changes

p-strusiewiczsurmacki-mobica:
This PR introduces 2 changes:

1. It moves `configuredRoutes` maps into `serviceContext` struct. This is rather minor change that might simplify some things - like e.g. instead of remembering to delete the value from the `configuredRoutes` map it is sufficient to delete the service whatsoever sometimes. Additionally fixed some encapsulation.

2. It fixes the `service-tests` target. It turned out that there were a few issues in the tests (e.g. incorrect local address was found so the test server was reachable from the kind cluster, resulting in no traffic to/from loadbalancer/egress etc.). All tests should work fine now. Also, if any issue will occur the test will now explicitly fail in the CI.

Important info:

When running the `service-tests` target the test will try to find an interface with the name starting with `br-` (as used by default by docker/kind). This should be sufficient for the CI. However, if you have multiple docker networks on your dev machine you'll have to use `-networkInterface <interface-name>` flag to explicitly specify the interface to be used.

### DIFF
--- a/pkg/manager/servicesLeader.go
+++ b/pkg/manager/servicesLeader.go
@@ -78,7 +78,6 @@ func (sm *Manager) StartServicesLeaderElection(ctx context.Context, service *v1.
 	}
 
 	svcCtx.isActive = true
-	// activeService[string(service.UID)] = true
 
 	// start the leader election code loop
 	leaderelection.RunOrDie(childCtx, leaderelection.LeaderElectionConfig{
@@ -112,7 +111,6 @@ func (sm *Manager) StartServicesLeaderElection(ctx context.Context, service *v1.
 				}
 				// Mark this service is inactive
 				svcCtx.isActive = false
-				// activeService[string(service.UID)] = false
 			},
 			OnNewLeader: func(identity string) {
 				// we're notified when new leader elected
--- a/pkg/manager/servicesLeader.go
+++ b/pkg/manager/servicesLeader.go
@@ -77,7 +77,7 @@ func (sm *Manager) StartServicesLeaderElection(ctx context.Context, service *v1.
 		return fmt.Errorf("failed to get context for service %q with UID %q: nil context", service.Name, service.UID)
 	}
 
-	svcCtx.IsActive = true
+	svcCtx.isActive = true
 	// activeService[string(service.UID)] = true
 
 	// start the leader election code loop
@@ -105,13 +105,13 @@ func (sm *Manager) StartServicesLeaderElection(ctx context.Context, service *v1.
 			OnStoppedLeading: func() {
 				// we can do cleanup here
 				log.Info("leadership lost", "service", service.Name, "leader", sm.config.NodeName)
-				if svcCtx.IsActive {
+				if svcCtx.isActive {
 					if err := sm.deleteService(service.UID); err != nil {
 						log.Error("service deletion", "err", err)
 					}
 				}
 				// Mark this service is inactive
-				svcCtx.IsActive = false
+				svcCtx.isActive = false
 				// activeService[string(service.UID)] = false
 			},
 			OnNewLeader: func(identity string) {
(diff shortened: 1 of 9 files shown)


testing/services/pkg/deployment/kubernetes.go
```diff
@@ -156,6 +156,10 @@ func (d *Deployment) CreateDeployment(ctx context.Context, clientset *kubernetes
 						{
 							Name:  "kube-vip-web",
 							Image: "docker.io/plndr/e2e:0.0.1",
+							// Image: "wbitt/network-multitool",
```
Cellebyte: can we remove this part?

p-strusiewiczsurmacki-mobica: Sorry. Done :)

pkg/manager/servicesLeader.go
```diff
@@ -77,7 +77,7 @@ func (sm *Manager) StartServicesLeaderElection(ctx context.Context, service *v1.
 		return fmt.Errorf("failed to get context for service %q with UID %q: nil context", service.Name, service.UID)
 	}
 
-	svcCtx.IsActive = true
+	svcCtx.isActive = true
 	// activeService[string(service.UID)] = true
```
Cellebyte: as we now have `svcCtx` this line could be removed as well :)

pkg/manager/servicesLeader.go
```diff
@@ -105,13 +105,13 @@ func (sm *Manager) StartServicesLeaderElection(ctx context.Context, service *v1.
...
-				if svcCtx.IsActive {
+				if svcCtx.isActive {
 					if err := sm.deleteService(service.UID); err != nil {
 						log.Error("service deletion", "err", err)
 					}
 				}
 				// Mark this service is inactive
-				svcCtx.IsActive = false
+				svcCtx.isActive = false
 				// activeService[string(service.UID)] = false
```
Cellebyte: same here 👍

testing/services/pkg/deployment/kubernetes.go
```diff
@@ -297,8 +297,8 @@ func (s *Service) CreateService(ctx context.Context, clientset *kubernetes.Clien
 		// We need to inspect the event and get ResourceVersion out of it
 		switch event.Type {
 		case watch.Added, watch.Modified:
-			// slog.Debugf("Endpoints for service [%s] have been Created or modified", s.service.ServiceName)
 			svc, ok := event.Object.(*v1.Service)
+			// slog.Debugf("Endpoints for service [%s] have been Created or modified", svc.Name)
```
Cellebyte: either uncomment or remove ^^

testing/services/pkg/deployment/services.go
```diff
@@ -84,10 +94,18 @@ func (config *TestConfig) StartServiceTest(ctx context.Context, clientset *kuber
...
 		}
 	}
 
-	slog.Infof("🏆 Testing Complete [%d] passed", config.SuccessCounter)
+	printf := slog.Infof
+	if len(errs) > 0 {
+		printf = slog.Errorf
+	}
+
+	printf("🏆 Testing Complete [%d] passed / [%d] failed", config.SuccessCounter, len(errs))
```
Cellebyte, line 106: use a constant for the string. and call the logger instead of indirect assignment of different log functions to `printf`

testing/services/pkg/deployment/services.go
```diff
@@ -84,10 +94,18 @@ func (config *TestConfig) StartServiceTest(ctx context.Context, clientset *kuber
...
+
+	return errs
```
Cellebyte:
```suggestion
        const testComplete = "🏆 Testing Complete [%d] passed / [%d] failed"
	if len(errs) > 0 {
		slog.Errorf(testComplete, config.SuccessCounter, len(errs)
		return errs
	}
	slog.Infof(testComplete, config.SuccessCounter, len(errs))

	return errs
```

Cellebyte (commented):
Some small change :)

pkg/manager/servicesLeader.go
p-strusiewiczsurmacki-mobica: Done :)

testing/services/pkg/deployment/kubernetes.go
p-strusiewiczsurmacki-mobica: Done :)

pkg/manager/servicesLeader.go
p-strusiewiczsurmacki-mobica: Done :)

PR #1185, opened 2025-06-24
Move towards internal nftables egress

thebsdbox:
This currently will implement IPv4 egress from pods using a load balancer address, without the need for various additional binaries (iptables/nftables). 

Tasks:
- [x] IPv4
- [x] IPv6 
- [x] No SNAT for pod CIDR / service CIDR
- [x] No SNAT for specific networks added as an annotation
- [x] Rules deleted when service is deleted
- [x] Specifies ports

We create an IPv4 Table and and IPv6 Table and a chain for each service, that uses the service UID which makes it easy to clean up services or all kube-vip rules by deleting the table.

## Example of a SNAT from `10.0.0.1` to `11.0.0.2` as long as the destination isn't the `podCIDR` or the `serviceCIDR`
```
   ip saddr { 10.0.0.1 } ip daddr != 10.0.0.0-10.0.255.255 ip daddr != 10.96.0.0-10.96.255.255 snat to 11.0.0.2
```

## Example of SNAT as long as the destination port is 123 for tcp and udp
```
   ip saddr { 10.0.0.1 } tcp dport { 123 } udp dport { 123 } snat to 101.0.0.1
```

### Example of SNAT with IPv6
```
   ip6 saddr { 2001:1::1 } snat to 2001:2::1
```


pkg/nftables/nftables.go
```diff
@@ -0,0 +1,204 @@
...
+	"github.com/google/nftables/expr"
+	"golang.org/x/sys/unix"
+)
+
+var Accept = nftables.ChainPolicyAccept
+
+const (
+	NatTable  = "kube_vip"
+	SNatChain = "kube_vip_snat"
+	DNatChain = "kube_vip_dnat"
```
thebsdbox: This isn't required as we're never doing DNAT

go.mod
```diff
@@ -8,6 +8,7 @@ require (
 	github.com/florianl/go-conntrack v0.4.0
 	github.com/golang/protobuf v1.5.4
 	github.com/google/go-cmp v0.7.0
+	github.com/google/nftables v0.2.0
```
rikatz: hum, question here: I can see there's a 0.3.0 tag, with a bunch of changes. Is it worth using it instead?

thebsdbox: Good point, I'll fix that!

pkg/manager/service_egress.go
```diff
@@ -180,6 +182,31 @@ func (sm *Manager) configureEgress(vipIP, podIP, namespace string, annotations m
...
+	if internalEgress != "" {
+		// Create an array of CIDRs that we wont SNAT to.
+		ignoreCIDRs := []string{
+			podCidr,
+			serviceCidr,
+		}
+
+		// Ad anny specifically denied networks
+		if deniedNetworks != "" {
+			networks := strings.Split(deniedNetworks, ",")
```
rikatz: dumb question: can deniedNetworks have space? is it work maybe trimming spaces?

testing/services/pkg/deployment/kubernetes.go
```diff
@@ -238,6 +239,12 @@ func (s *Service) CreateService(ctx context.Context, clientset *kubernetes.Clien
 			"kube-vip.io/egress": "true",
 		}
 	}
+
+	if s.egressInternal {
+		svc.Annotations["kube-vip.io/egress-internal"] = "true"
```
rikatz, line 244: @thebsdbox maybe a suggestion: you also set the annotations here: https://github.com/kube-vip/kube-vip/pull/1185/files#diff-06e954781cf1c4eea903c33aa93db82d2d103ae66b0ae9b0f620280c82082c56R27

maybe in a future all those annotations can belong to some "annotations" package, documented, as constants and we generate the docs for every annotation from a single place + automatic docs for the annotations? :)

rikatz: aaah nevermind, this is the testing package :facepalm: 

And I think it totally makes sense to test the annotation here and be sure it won´t regress if someone changes the constant above. Ignore me

pkg/nftables/nftables.go
```diff
@@ -0,0 +1,204 @@
...
+	"fmt"
+	"log/slog"
+	"net"
+
+	"github.com/google/nftables"
+	"github.com/google/nftables/expr"
+	"golang.org/x/sys/unix"
+)
+
+var Accept = nftables.ChainPolicyAccept
```
rikatz: nit: maybe make it a const (it is a const on original package anyway: https://github.com/google/nftables/blob/main/chain.go#L93C2-L93C19

pkg/nftables/nftables.go
```diff
@@ -0,0 +1,204 @@
...
+	NatTable  = "kube_vip"
+	SNatChain = "kube_vip_snat"
+	DNatChain = "kube_vip_dnat"
+)
+
+func ApplySNAT(podIP, vipIP string, ignoreCIDR []string, IPv6 bool) error {
+
+	conn, err := nftables.New()
+	if err != nil {
+		panic(err)
```
rikatz: why panic and not return the error? 

I think this applies for all below, the function signature already supports errors returns :)

pkg/nftables/nftables.go
```diff
@@ -0,0 +1,204 @@
...
+		Name:     SNatChain,
+		Table:    GetTable(IPv6),
+		Type:     nftables.ChainTypeNAT,
+		Hooknum:  nftables.ChainHookPostrouting,
+		Priority: nftables.ChainPriorityNATSource,
+		Policy:   &policy,
+	}
+}
+
+func FilterTable(conn *nftables.Conn, tableName string) (table *nftables.Table, err error) {
```
rikatz: nit: the return parameters here are inconsistent with the rest of package, those are named values :) 

as you are not even using the "table" name below, maybe remove the named returns here as well?

pkg/nftables/nftables.go
```diff
@@ -0,0 +1,204 @@
...
+	tableList, err := conn.ListTables()
```
rikatz: question: why using ListTables here and not ListTable (or ListTableOfFamily, in case of caring about IPv4 or IPv6)? 

I think the ListTables may be  a bit less expensive (probably not that much, as both ListTable, ListTables, etc uses https://github.com/google/nftables/blob/v0.3.0/table.go#L175) but at least it guarantees you get a single table, which IIUC is what you expect here: check and get a single one, or error if it doesn't exist.

Maybe the reason is because you care about "not having a table is not an error"?

pkg/nftables/nftables.go
```diff
@@ -0,0 +1,204 @@
...
+}
+
+func Flush(conn *nftables.Conn, IPv6 bool) error {
+	if t, err := FilterTable(conn, NatTable); err != nil {
+		return err
+	} else if t != nil {
+		conn.DelTable(t)
+	}
+
+	// These don't return errors, so not 100% sure how to guarentee things were created
```
rikatz: I think, per my reading on the library doc you need to do something like .Flush after every command you care being transactional? IIUC everytime you call something like DelTable, AddTable, it will just add it to a batch/buffer of commands to be sent, so probably need some .Flush() for all the stuff you care, or just being sure/clear that the errors will appear on the last statement (return conn.Flush())

pkg/nftables/nftables.go
```diff
@@ -0,0 +1,204 @@
...
+
+	// original pod IP
+	if net.ParseIP(podIP) == nil {
+		return nil, errors.New("ip is invalid")
+	}
+
+	// Create an element using our pod IP
+	elements := []nftables.SetElement{
+		{
+			Key: net.ParseIP(podIP).To4(),
```
rikatz: shouldn´t this reflect the IPv6 bool flag, instead of doing a To4 directly?

pkg/manager/service_egress.go
thebsdbox: yeah, I think for safety reasons we potentially should do that..

`10.0.0.0/24, 192.168.0.0/24` etc would probably crash the function :|

pkg/nftables/nftables.go
rikatz: or maybe this can just be dropped, I can see you already use nftables.ChainPolicyAccept below

pkg/nftables/nftables.go
```diff
@@ -0,0 +1,204 @@
...
+				FromData: start.To4(),
+				ToData:   end.To4(),
+			},
+		}
+		// // Add expression to the rule
+		rule.Exprs = append(rule.Exprs, expression...)
+	}
+
+	// Create SNAT rule
+	if net.ParseIP(vipIP) == nil {
```
rikatz: nit: maybe do this on the beginning of this function as well (together with the ParseIP you are doing for podIP)

pkg/nftables/nftables.go
```diff
@@ -0,0 +1,204 @@
...
+		return nil, errors.New("output_ip is not a valid ip")
+	}
+
+	expression = []expr.Any{
+		&expr.Immediate{
+			Register: 1,
+			Data:     net.ParseIP(vipIP).To4(),
```
rikatz: already doing this on line 179, maybe can join the info

pkg/nftables/nftables.go
thebsdbox: gone

go.mod
thebsdbox: done

pkg/nftables/nftables.go
thebsdbox: fixed

pkg/nftables/nftables.go
thebsdbox: fixed

aojea:
@thebsdbox I was talking today with @rikatz about using the nftables netlink interface ... and I love it and I move all my projects to it so sharing some lessons learned:
- use the userspace syntax to generate the bytecode or to bootstrap, you can get the golang representation from it, see https://github.com/aojea/kindnet/blob/main/docs/nftables-netlink.md and https://github.com/google/nftables/pull/298
- add tests that validate that your netlink representation matches the userspace representation, 
   - https://github.com/aojea/kindnet/blob/main/pkg/fastpath/fastpath_test.go 
   - https://github.com/aojea/kindnet/blob/main/pkg/nat64/proxy_test.go
   - with some magic you can also not require root to run those tests using rootless like in https://github.com/kubernetes-sigs/kube-network-policies/blob/main/pkg/networkpolicy/controller_test.go#L277-L450

I also have people asking me when to use it and I only recommend it if you are very concerned about your images dependencies, this allows you to build distroless images, your stability since the netlink interface is much stable than the usersapces tools,  if you have some knowledge of low level networking at the kernel and require to implement complex expressions tthat may not be available on the userspace tools, ... . For people that want to get something going then it is simpler to use https://github.com/kubernetes-sigs/knftables that wraps the userspace binaries

thebsdbox:
Thanks for the hints and tips @aojea! This has been something I've been wanting to implement for quite a while, ever since I spotted that the nftables pkg finally supported SNAT 👍. The syntax is pretty bonkers, having to use strace or debug modes to work out the payloads is pretty grim, but I've managed to implement pretty much everything I needed at this point. The rules are a lot cleaner than they were when implemented in iptables too 😄

pkg/nftables/nftables.go
thebsdbox: fixed :-)

pkg/nftables/nftables.go
thebsdbox: Moved this to the more simpler functions instead of pulling everything and then looping through!

pkg/nftables/nftables.go
thebsdbox: done.

thebsdbox:
Finally !

thebsdbox:
closing this one as there's a fresh one that's easier to review.

PR #1188, opened 2025-06-26
Basic e2e tests for BGP mode.

p-strusiewiczsurmacki-mobica:
This PR introduces basic E2E tests for services handing in BGP mode.

It uses GoBGP which runs on the host, therefore this test has to be run with **root privileges**. One idea to loose this requirement would be to run GoBGP in cluster but I was unable to find any reliable GoBGP docker image, and maintaining image for this seems might be a bit too much for the e2e test I think.

This test requires name of the kind network interface. By default it will use first interface named `br-*` (sufficient for CI) but if one have more than one interface with this name, it should be specified with env variable, e.g.: `NETWORK_INTERFACE=br-36d13ecf5bde`.

Docker client, which is used here to obtain IP address of the kind container might also not like the docker daemon version, this can be mitigated with another env variable like this `DOCKER_API_VERSION=1.50`

So command to run test might look for example like this (PATH is passed for access to go binaries from in sudo env):
```
sudo -E PATH=$PATH NETWORK_INTERFACE=br-36d13ecf5bde DOCKER_API_VERSION=1.50 make e2e-tests129-bgp
```

The test will check:
- BGP with IPv4 and IPv6
- MP-BGP for IPv4 over IPv6 session both for fixed hop address and `auto_sourceif` address
- MP-BGP for IPv6 over IPv4 session both for fixed hop address and `auto_sourceif` address

`auot_sourceip` is not tested as it requires container's  IP address to be known before kube-vip starts, but for the tests we use kube-vip deployed as a manifest, which is created before the container's creation.


thebsdbox:
Code looks good, is the root privilege tied down to the fact that it needs to bind to a low port < 1024?

p-strusiewiczsurmacki-mobica:
> Code looks good, is the root privilege tied down to the fact that it needs to bind to a low port < 1024?

Yeah, I believe that's the case.

PR #1189, opened 2025-06-26
Propagate root context instead of creating new ones

rikatz:
Context are being created everywhere :) 

This change makes the context part of a "root context" from the command, that will be then passed down to other parts of the code.

This may help in the future on setting context cancelation, timeouts but also passing down loggers from context, or in case there's a decision to implement OTEL to have a single context for traces, etc.


thebsdbox:
Initially this all looks good, main concern is ensuring that a service cleanup or something that could cause a `cancel()` doesn't wipe everything out.

rikatz:
hum, yeah, I will review all my changes now regarding old context vs context that are created with cancelation somewhere else, and maybe do some tests here.

Maybe I can think on a way to, using fake client + some dumb services, call context cancelation and guarantee that other resources still exists as well

Cellebyte:
@rikatz we are currently refactoring some things, so your PR will become outdated again in the next weeks.
@thebsdbox do you have an idea for tests to ensure, the context is not wiping everything?

rikatz:
@Cellebyte no rush from my side, this was kind of a best effort to improve the context usage :) but no rush for real. I can look at it after the refactor!

Cellebyte:
@rikatz I think now, you could take a look again on the current main branch 👍

rikatz:
folks, sadly I wont have much time on the next few weeks to test my changes, do you want to keep with it or "winning team we don't change the formation" and we can close this? :)

thebsdbox:
I'll keep with it, mainly in the hope I can coerce you to keep helping..

thebsdbox:
I know you're super busy, what's your thoughts on getting this completed? If not I might take it over from you? (if that's cool with you?)

rikatz:
I am worried of breaking things that are currently working 😬

And then not having time to investigate or fix them (aka I dont wanna be THAT kind of person that comes, does a PR, breaks things, disappears)

I guess if this is not being an issue right now I would be tempted to close and make the change with more calm, wdyt?

thebsdbox:
Love a drive by PR. I'll keep it open until I can find some time to re-implement it 😄

Cellebyte:
Let’s close it :)

PR #1206, opened 2025-07-21
Use insecure client only in E2E tests

p-strusiewiczsurmacki-mobica:
This PR should fix #1190

It seems that mixing secure and insecure clients was not a great idea. This PR ensures that insecure client will only be used in E2E testing environment.

PR #1209, opened 2025-07-23
Removed insecure client use in E2E tests

p-strusiewiczsurmacki-mobica:
This PR is a folloup to #1206 

It removes insecure client altogether in favor of adding additional certSANs to the kind cluster. 63 addresses (or less, if the address will fall out of the docker's interface CIDR)  will be added to the certs starting from the docker bridge address.
In case that there are many docker bridges (or other interfaces named `br-XXX`) the network interface for RT tests has to be provided with NETWORK_INTERFACE=<name> env variable similar to what's described in #1188

### DIFF
--- a/pkg/backend/backend.go
+++ b/pkg/backend/backend.go
@@ -2,8 +2,6 @@ package backend
 
 import (
 	"fmt"
-	"os"
-	"strconv"
 	"time"
 
 	log "log/slog"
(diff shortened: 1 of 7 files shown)


Cellebyte:
This is a much better solution then the insecure client approach.
Thanks @p-strusiewiczsurmacki-mobica for investigating.

PR #1210, opened 2025-07-27
Don't execute the e2e-tests in parallel

Cellebyte:
This will prolong the execution of the CI e2e-tests, 
but at least does not require to restart them all the time.


Cellebyte:
@thebsdbox this reduces the e2e-test runs by 4 jobs as they only run on merge requests.

thebsdbox:
Makes sense. We need to remove the before v1.29 test as it EOL too.

thebsdbox (approved):

PR #1213, opened 2025-07-28
Adds the internal egress functionality

thebsdbox:
This currently will implement IPv4 egress from pods using a load balancer address, without the need for various additional binaries (iptables/nftables). 

Tasks:
- [x] IPv4
- [x] IPv6 
- [x] No SNAT for pod CIDR / service CIDR
- [x] No SNAT for specific networks added as an annotation
- [x] Rules deleted when service is deleted
- [x] Specifies ports

We create an IPv4 Table and and IPv6 Table and a chain for each service, that uses the service UID which makes it easy to clean up services or all kube-vip rules by deleting the table.

## Example of a SNAT from `10.0.0.1` to `11.0.0.2` as long as the destination isn't the `podCIDR` or the `serviceCIDR`
```
   ip saddr { 10.0.0.1 } ip daddr != 10.0.0.0-10.0.255.255 ip daddr != 10.96.0.0-10.96.255.255 snat to 11.0.0.2
```

## Example of SNAT as long as the destination port is 123 for tcp and udp
```
   ip saddr { 10.0.0.1 } tcp dport { 123 } udp dport { 123 } snat to 101.0.0.1
```

### Example of SNAT with IPv6
```
   ip6 saddr { 2001:1::1 } snat to 2001:2::1
```


pkg/manager/service_egress.go
```diff
@@ -206,6 +208,29 @@ func (sm *Manager) configureEgress(vipIP, podIP, namespace string, annotations m
...
 
+	// Use the internal egress implementation
+	if internalEgress != "" {
+		// Create an array of CIDRs that we wont SNAT to.
+		ignoreCIDRs := []string{
+			podCidr,
+			serviceCidr,
+		}
+
+		// Ad anny specifically denied networks
```
Cellebyte, line 265: nit `typo` 
```suggestion
		// Ad any specifically denied networks
```

testing/services/main.go
```diff
@@ -38,6 +38,7 @@ func main() {
 	flag.BoolVar(&t.LeaderActive, "leaderActive", false, "Perform a test on the active leader")
 	flag.BoolVar(&t.LocalDeploy, "localDeploy", false, "Perform a test on the active leader")
 	flag.BoolVar(&t.Egress, "egress", false, "Perform an egress test")
+	flag.BoolVar(&t.EgressInternal, "egressInternal", false, "Perform an egress test, using the internal functionality")
```
Cellebyte, line 41: Is this option really only needed for testing?

pkg/nftables/nftables.go
```diff
@@ -0,0 +1,412 @@
...
+					tcpElements = append(tcpElements, nftables.SetElement{Key: binaryutil.BigEndian.PutUint16(uint16(port))})
+				case "udp":
+					//nolint:gosec
+					udpElements = append(udpElements, nftables.SetElement{Key: binaryutil.BigEndian.PutUint16(uint16(port))})
+				default:
+					slog.Error("[egress]", "unknown protocol", data[0])
+				}
+			}
+		}
+		if len(tcpElements) != 0 {
```
Cellebyte, line 293: maybe we should also support sctp as it is officially supported by k8s/kube-proxy as well?

testing/services/main.go
Cellebyte: What is it's purpose? :)

thebsdbox: It creates a service with the correct annotation for using the internal egress functionality.

Cellebyte: ahh :)

pkg/nftables/nftables.go
thebsdbox: Done

pkg/manager/service_egress.go
thebsdbox: Fixed

Cellebyte (approved):
LGTM :) just needs rebase I think 👍

PR #1215, opened 2025-07-29
Kubernetes 0.33.3 API and endpoint slices as the default moving forward

thebsdbox:
Breaking change as the flag for enabling endpointSlices is now gone as it is the default.. If you require the older functionality then you have to specify it with `EnableEndpoints` for the deprecated behaviour.


Cellebyte:
I think it looks good and needs docs :)

Issue #1201, opened 2025-07-12
It is not possible assign IP for services with same name in different namespaces


KonstantinNosov:
**Describe the bug**
If there are two services with same name but in diferent namesapces (in my case namespace is per environment dev/prod/uat) and both have `kube-vip.io/loadbalancerIPs` annotation with unique ip, only one gets ip assigned

**To Reproduce**
1. Create 2 services named `mysvc` in different namespaces
2. Add `kube-vip.io/loadbalancerIPs` annotations with different ips 
3. check status in service live manifest - on will have `status.loadBalancer` empty another will have it indicating ip assignment

**Environment (please complete the following information):**
 - OS/Distro: ubuntu 24.04
 - Kubernetes Version:  v1.32.1+k3s1
 - Kube-vip Version:  0.9.1

**`Kube-vip.yaml`:**
If Possible add in your kube-vip manifest (please remove anything that is confidential)

```yaml
apiVersion: apps/v1
kind: DaemonSet
metadata:
  annotations:
    argocd.argoproj.io/tracking-id: infrastructure-prod_kubevip-core:apps/DaemonSet:kube-system/kubevip-core
    deprecated.daemonset.template.generation: '12'
    exordis/subsystemDescription: infrastructure subsystem
  creationTimestamp: '2025-03-11T21:36:01Z'
  generation: 12
  labels:
    exordis/application: infrastructure-kubevip-core
    exordis/environment: prod
    exordis/product: infrastructure
    exordis/subsystem: infrastructure
  name: kubevip-core
  namespace: kube-system
  resourceVersion: '95149324'
  uid: f4a2f38f-a5ce-4cc7-a108-bfcb68504d8c
spec:
  revisionHistoryLimit: 10
  selector:
    matchLabels:
      app.kubernetes.io/instance: kubevip-core
      app.kubernetes.io/name: kubevip-core
  template:
    metadata:
      annotations:
        exordis/subsystemDescription: infrastructure subsystem
        kubectl.kubernetes.io/restartedAt: '2025-04-16T20:47:36Z'
      creationTimestamp: null
      labels:
        app.kubernetes.io/instance: kubevip-core
        app.kubernetes.io/name: kubevip-core
        exordis/application: infrastructure-kubevip-core
        exordis/environment: prod
        exordis/product: infrastructure
        exordis/subsystem: infrastructure
    spec:
      containers:
        - args:
            - manager
            - '--prometheusHTTPServer='
          env:
            - name: cp_enable
[...]

PR #1233, opened 2025-08-13
Issue 1201 - uniquely identify a service lease combining namespace + service name

makeittotop:
This PR fixes [this issue](https://github.com/kube-vip/kube-vip/issues/1201) by adding namespace to uniquely identify a service lease, instead of merely relying on the service name.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -166,7 +166,7 @@ kind-quick:
 	kubectl create configmap --namespace kube-system kubevip --from-literal range-global=172.18.100.10-172.18.100.30
 	kubectl apply -f https://raw.githubusercontent.com/kube-vip/kube-vip-cloud-provider/main/manifest/kube-vip-cloud-controller.yaml
 	kind load docker-image --name kube-vip $(REPOSITORY)/$(TARGET):$(DOCKERTAG)
-	docker run --network host --rm $(REPOSITORY)/$(TARGET):$(DOCKERTAG) manifest daemonset --services --inCluster --arp --servicesElection --interface  eth0 | kubectl apply -f -
+	docker run --network host --rm $(REPOSITORY)/$(TARGET):$(DOCKERTAG) manifest daemonset --services --inCluster --arp --servicesElection --interface  eth0 | sed 's|image:.*$ |image: $(REPOSITORY)/$(TARGET):$(DOCKERTAG)|'| kubectl apply -f -
 
 kind-reload:
 	kind load docker-image $(REPOSITORY)/$(TARGET):$(DOCKERTAG) --name kube-vip
--- a/pkg/services/leader.go
+++ b/pkg/services/leader.go
@@ -45,7 +45,8 @@ func (p *Processor) StartServicesWatchForLeaderElection(ctx context.Context) err
 
 // The startServicesWatchForLeaderElection function will start a services watcher, the
 func (p *Processor) StartServicesLeaderElection(ctx context.Context, service *v1.Service) error {
-	serviceLease := fmt.Sprintf("kubevip-%s-%s", service.Namespace, service.Name)
+	serviceLease := fmt.Sprintf("kubevip-%s", service.Name)
+	serviceLeaseId := fmt.Sprintf("kubevip-%s/%s", serviceLease, service.Namespace)
 	log.Info("new leader election", "service", service.Name, "namespace", service.Namespace, "lock_name", serviceLease, "host_id", p.config.NodeName)
 	// we use the Lease lock type since edits to Leases are less common
 	// and fewer objects in the cluster watch "all Leases".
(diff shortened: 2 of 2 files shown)


pkg/services/leader.go
```diff
@@ -45,7 +45,7 @@ func (p *Processor) StartServicesWatchForLeaderElection(ctx context.Context) err
 
 // The startServicesWatchForLeaderElection function will start a services watcher, the
 func (p *Processor) StartServicesLeaderElection(ctx context.Context, service *v1.Service) error {
-	serviceLease := fmt.Sprintf("kubevip-%s", service.Name)
+	serviceLease := fmt.Sprintf("kubevip-%s-%s", service.Namespace, service.Name)
```
Cellebyte: @makeittotop this lease is created inside the namespace a service is created in, so it is not required to add the namespace in the lease name as the lease api is namespaced.

makeittotop: @Cellebyte yes, you're  right in that the lease is created inside a namespace, so ideally having the same name (based on the service) should not really overlap with a lease of the same name inside a different namespace, but here is what happens in my opinion - 

https://github.com/kube-vip/kube-vip/blob/main/pkg/services/leader.go#L17-L19

```

var (
	svcLocks map[string]*sync.Mutex
)

...

if _, ok := svcLocks[serviceLease]; !ok {
svcLocks[serviceLease] = new(sync.Mutex)
}


	svcLocks[serviceLease].Lock()
	defer svcLocks[serviceLease].Unlock()
```

This is a global map which references leases(locks) with their names. Now the name only consists of the service name which creates a collision when another service of the same name is created in another ns.  That's why adding "namespace" to the lease would uniquely identify it.

The pod throws an error like - 
```
kube-vip-ds-qvcc5 kube-vip E0814 16:44:35.542770       1 leaderelection.go:440] error initially creating leader election record: leases.coordination.k8s.io "kubevip-test-lb-0" already exists
```
and doesn't create the associated lease in that ns. Thus the svc always have loadbalancer IP in pending state 
```
test-lb-0      LoadBalancer   10.96.232.113   <pending>     80:32537/TCP             4m56s
```

Steps to reproduce - 
```
# create a svc in default namespace 
kubectl -ndefault create service loadbalancer test-lb-0 --tcp=80:80 --dry-run=client -o yaml | kubectl apply -f -
# observe that it gets a lb ip 

# create another svc with same name in another ns 
kubectl -n test create service loadbalancer test-lb-0 --tcp=80:80 --dry-run=client -o yaml | kubectl apply -f -
# observe that it will never get an ip for the lb (as shown above)
```

Does this make sense @Cellebyte ?

Cellebyte: Yes :)

Cellebyte: I think, splitting it into a leaseName + leaseIdentifier would be better, otherwise the name gets very long.

The `leaseIdentifier` would then be the combined name of name + namespace using a `/` as a separator.

E.g.

* `kubevip-my-service` becomes `kubevip-my-service/my-namespace`

This would also fit with the new PR raised yesterday -> #1235

makeittotop: I see. I'll take a look into that

makeittotop: Is this (more or less) what you had in mind @Cellebyte?  I ran few tests on a local kind cluster to confirm this works as expected

Makefile
```diff
@@ -166,7 +166,7 @@ kind-quick:
 	kubectl create configmap --namespace kube-system kubevip --from-literal range-global=172.18.100.10-172.18.100.30
 	kubectl apply -f https://raw.githubusercontent.com/kube-vip/kube-vip-cloud-provider/main/manifest/kube-vip-cloud-controller.yaml
 	kind load docker-image --name kube-vip $(REPOSITORY)/$(TARGET):$(DOCKERTAG)
-	docker run --network host --rm $(REPOSITORY)/$(TARGET):$(DOCKERTAG) manifest daemonset --services --inCluster --arp --servicesElection --interface  eth0 | kubectl apply -f -
```
makeittotop, line 169: Made a change to substitute the  docker image loaded into kind into the ds. Let me know if it's not OK.

makeittotop:
renamed var to conform to linter rec.

Makefile
Cellebyte: I think this part is not required, but I need to check for myself ^^

pkg/services/leader.go
Cellebyte: Yes, now we just need to check if we are using this somewhere else as well and if so we can make it a generic function :)

Makefile
Cellebyte: I checked it and update the manifest generation to make it configurable.

pkg/services/leader.go
Cellebyte: LGTM

Cellebyte:
This one will be closed, but I ported your Commits into #1261

PR #1235, opened 2025-08-14
Added common lease for multiple services

### LINKED ISSUE #879: svc_enable=true, having two load balancers with the same IP but different ports will result in the IP appearing on two different servers.
**Describe the bug**

I have a three-node Kubernetes cluster with two load balancers using the same IP but different port numbers. This VIP appears on two nodes simultaneously.

**To Reproduce**
1. Create kube-vip-cloud-provider configmap
```yaml
apiVersion: v1
data:
  cidr-qt-system: 10.106.109.9/32
kind: ConfigMap
metadata:
  annotations:
    meta.helm.sh/release-name: kube-vip-cloud-provider
  labels:
    app.kubernetes.io/managed-by: Helm
  name: kubevip
```
2. Create 2 load balancers with the same ip
```yaml
apiVersion: v1
kind: Service
metadata:
  annotations:
    kube-vip.io/loadbalancerIPs: 10.106.109.9
  labels:
    implementation: kube-vip
  name: traefik-lb-1
spec:
  ports:
  - name: agent
    port: 8000
    protocol: TCP
    targetPort: agent
  selector:
    app.kubernetes.io/instance: traefik
    app.kubernetes.io/name: traefik
  type: LoadBalancer
``` 

```yaml
apiVersion: v1
kind: Service
metadata:
  annotations:
    kube-vip.io/loadbalancerIPs: 10.106.109.9
  labels:
    implementation: kube-vip
  name: traefik-lb-2
spec:
  ports:
  - name: web
    port: 80
    protocol: TCP
    targetPort: web
  selector:
    app.kubernetes.io/instance: traefik
    app.kubernetes.io/name: traefik
  type: LoadBalancer
```
3. To check the creation status of a virtual IP (VIP) on different servers.
```bash
[root@ip-10-106-109-51 ~]# ip addr show ens18
2: ens18: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc fq_codel state UP group default qlen 1000
    link/ether fe:33:6e:ee:11:43 brd ff:ff:ff:ff:ff:ff
    inet 10.106.109.51/24 brd 10.106.109.255 scope global noprefixroute ens18
    inet 10.106.109.9/32 scope global ens18
```

```bash
[root@ip-10-106-109-52 ~]# ip addr show ens18
2: ens18: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc fq_codel state UP group default qlen 1000
    link/ether 52:e8:74:31:5f:db brd ff:ff:ff:ff:ff:ff
    inet 10.106.109.52/24 brd 10.106.109.255 scope global noprefixroute ens18
    inet 10.106.109.9/32 scope global ens18
```

**Expected behavior**

The VIP address appears only on one server.

**Screenshots**

 - OS/Distro: [e.g. CentOS 8.5.2111]
 - Kubernetes Version: [e.g. v.1.24.17]
 - Kube-vip Version: [e.g. v0.8.0]

**`Kube-vip.yaml`:**
```yaml
apiVersion: apps/v1
kind: DaemonSet
metadata:
  annotations:
    deprecated.daemonset.template.generation: "2"
    meta.helm.sh/release-name: kube-vip
    meta.helm.sh/release-namespace: qt-system
  creationTimestamp: "2024-06-13T07:07:36Z"
  generation: 2
  labels:
    app.kubernetes.io/instance: kube-vip
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/name: kube-vip
    app.kubernetes.io/version: 202406061636-db14aabc
    helm.sh/chart: kube-vip-0.1.0
  name: kube-vip
  resourceVersion: "554829"
  uid: a381a14c-1dd3-4077-9aae-e2b1c3ad8b4e
spec:
  revisionHistoryLimit: 10
  selector:
    matchLabels:
      app.kubernetes.io/instance: kube-vip
      app.kubernetes.io/name: kube-vip
  template:
    metadata:
      creationTimestamp: null
      labels:
        app.kubernetes.io/instance: kube-vip
        app.kubernetes.io/name: kube-vip
    spec:
      affinity: {}
      containers:
      - args:
        - manager
        - --prometheusHTTPServer=
        env:
        - name: vip_arp
          value: "true"
        - name: vip_nodename
          valueFrom:
            fieldRef:
              apiVersion: v1
              fieldPath: spec.nodeName
        - name: vip_ddns
          value: "false"
        - name: svc_enable
          value: "true"
        - name: svc_election
          value: "true"
        - name: vip_leaderelection
          value: "true"
        - name: vip_leaseduration
          value: "60"
        - name: vip_renewdeadline
          value: "20"
        - name: vip_retryperiod
          value: "5"
        image: plndr/kube-vip:v0.8.0
        imagePullPolicy: IfNotPresent
        name: kube-vip
        resources:
          limits:
            cpu: 100m
            memory: 128Mi
          requests:
            cpu: 10m
            memory: 16Mi
        securityContext:
          capabilities:
            add:
            - NET_ADMIN
            - NET_RAW
      dnsPolicy: ClusterFirst
      hostNetwork: true
      imagePullSecrets:
      - name: qt-registry
      restartPolicy: Always
      securityContext: {}
      serviceAccount: kube-vip
      serviceAccountName: kube-vip
      tolerations:
      - effect: NoSchedule
        operator: Exists
      - effect: NoExecute
        operator: Exists
  updateStrategy:
    rollingUpdate:
      maxSurge: 0
      maxUnavailable: 1
    type: RollingUpdate
status:
  currentNumberScheduled: 3
  desiredNumberScheduled: 3
  numberAvailable: 3
  numberMisscheduled: 0
  numberReady: 3
  observedGeneration: 2
  updatedNumberScheduled: 3
```
**Additional context**
If I delete svc_election, it will work normally.

p-strusiewiczsurmacki-mobica:
This should fix #879

This PR introduces new annotation `kube-vip.io/leaseName: <name>` that will tie multiple services with single lease (so only one node will be a leader)..

Additionally fixed some issues I found with IP address deletion in ARP mode.


pkg/arp/arp.go
```diff
@@ -12,6 +12,7 @@ import (
 
 type Manager struct {
 	instances map[string]*Instance
```
Cellebyte: should we not use `sync.Map` here as well?

pkg/cluster/service.go
```diff
@@ -330,7 +330,7 @@ func (cluster *Cluster) StartLoadBalancerService(ctx context.Context, c *kubevip
 			return
 		}
 		for i := range cluster.Network {
-			if c.EnableARP && cluster.arpMgr.Count(cluster.Network[i].ARPName()) > 0 {
+			if c.EnableARP && cluster.arpMgr.Count(cluster.Network[i].ARPName()) > 1 {
```
Cellebyte, line 333: So first does election and the rest is skipping election here?

Cellebyte: What is with the other modes, do they work accordingly and we only skip it here because of the single path arp protocol?

Cellebyte:
SinglePath: ARP Mode 
* Services with the same IP but different Ports should be able to listen on the same Node when `externalTrafficPolicy: Cluster`, `externalTrafficPolicy: Local` should not be supported.

MultiPath: BGP, Route Table 
* could function the same, but also does not support `Local` as above.

multiple Services sharing the same LoadBalancer IP but using different ports is only possible in a Service of type `LoadBalancer` which uses `externalTrafficPolicy: Cluster` otherwise it is tricky to select a node, as it needs to have endpoints for both selectors.

pkg/cluster/service.go
p-strusiewiczsurmacki-mobica: It work's like this, but not in this part of the code. This change is an addtional fix for ip address deletion in ARP mode I found along the work.

As for the other modes, I am checking those.

Cellebyte:
When your PR is Ready for Review please ensure while rebasing, that you not overwrite the fix for the service lock. I think you changed the code as well.

pkg/arp/arp.go
p-strusiewiczsurmacki-mobica: Done :)

p-strusiewiczsurmacki-mobica:
@Cellebyte This should be ready for the review.

It turned out that service leader election for BGP mode was never supported, so I've added that for consistency. Also added e2e tests for BGP and RT with service leader election.

> When your PR is Ready for Review please ensure while rebasing, that you not overwrite the fix for the service lock. I think you changed the code as well.

Yes, those should be included.

pkg/services/processor.go
```diff
@@ -266,21 +274,26 @@ func (p *Processor) AddOrModify(ctx context.Context, event watch.Event, serviceF
...
 	if err != nil {
 		return false, fmt.Errorf("(svcs) unable to get context: %w", err)
 	}
-	if svcCtx != nil && svcCtx.IsActive {
+
+	if svcCtx != nil {
+		log.Info("(svcs) deleted", "service name", svc.Name, "namespace", svc.Namespace, "IsActive", svcCtx.IsActive)
+	}
+
+	if svcCtx != nil {
```
Cellebyte: any reason we have the if condition here twice?

p-strusiewiczsurmacki-mobica: Yeah, deleted that.

PR #1240, opened 2025-08-18
check conn.Flush error in nftables SNAT

hensur:


pkg/nftables/nftables.go
```diff
@@ -41,9 +40,11 @@ func ApplySNAT(podIP, vipIP, service, destinationPorts string, ignoreCIDR []stri
 		}
 	}
 	slog.Debug("[egress]", "Creating Chain for service", service, "IPv6", IPv6)
-	// These don't return errors, so not 100% sure how to guarantee things were created
```
thebsdbox, line 44: This comment is about the `conn.AddChain()` function, it doesn't return anything so we can't guarentee that a chain is actually added.

pkg/nftables/nftables.go
```diff
@@ -41,9 +40,11 @@ func ApplySNAT(podIP, vipIP, service, destinationPorts string, ignoreCIDR []stri
...
 	conn.AddChain(GetSNatChain(IPv6, service))
-	conn.Flush()
+	err = conn.Flush()
```
thebsdbox, line 44: :+1:

pkg/nftables/nftables.go
hensur: I was looking at the code (https://github.com/google/nftables/blob/6f574e7fd1b4d07006d1c598cdb9b22789e5e541/chain.go#L143) which does nothing except update the conn's in memory buffer.
The only write seems to be done here: https://github.com/google/nftables/blob/6f574e7fd1b4d07006d1c598cdb9b22789e5e541/conn.go#L262

If we check the error from `Flush()` we can be sure that everything has been applied over the netlink socket.

Because of this I thought the comment might be redundant.

PR #1251, opened 2025-08-26
fix: using DNS name for address

nianyush:


pkg/cluster/clusterDDNS.go
```diff
@@ -12,7 +12,7 @@ import (
 // during runtime if IP changes, startDDNS don't have to do reconfigure because
 // dnsUpdater already have the functionality to keep trying resolve the IP
 // and update the VIP configuration if it changes
-func (cluster *Cluster) StartDDNS(ctx context.Context) error {
+func (cluster *Cluster) StartDDNS(ctx context.Context, vipSubnet string) error {
```
Cellebyte, line 15: is this information not already part of `cluster`

pkg/vip/dns.go
```diff
@@ -13,13 +13,15 @@ type IPUpdater interface {
...
 
 type ipUpdater struct {
-	vip Network
+	vip       Network
+	vipSubnet string
 }
 
 // NewIPUpdater creates a DNSUpdater
-func NewIPUpdater(vip Network) IPUpdater {
+func NewIPUpdater(vip Network, vipSubnet string) IPUpdater {
```
Cellebyte: why do we need to set it here if it is already in the `Network` type and just needs an extension of the interface?

pkg/vip/dns.go
```diff
@@ -53,6 +55,9 @@ func (d *ipUpdater) Run(ctx context.Context) {
 					log.Error("error adding virtual IP", "err", err)
 				}
 
+				if err := d.vip.SetMask(d.vipSubnet); err != nil {
```
Cellebyte: as this value is hardcoded on start or we could run the `auto` mode we don't need to configure it on every ip update.

pkg/cluster/clusterDDNS.go
```diff
@@ -22,6 +22,10 @@ func (cluster *Cluster) StartDDNS(ctx context.Context) error {
 		if err = cluster.Network[i].SetIP(ip); err != nil {
 			return err
 		}
+
+		if err := cluster.Network[i].SetMask(vipSubnet); err != nil {
```
Cellebyte, line 26: I think this should be enough.

pkg/cluster/clusterDDNS.go
nianyush: i don't think we have kubevip.Config access here. Can you show me if i am missing it?

pkg/vip/dns.go
nianyush: Removed

pkg/vip/dns.go
nianyush: removed

pkg/cluster/clusterDDNS.go
Cellebyte: No but, it could be that the required information is already part of cluster :)

nianyush: ummm maybe i am just not familiar enough with this code. How can we call SetMask in clusterDDNS in this case if this is not passed here? Or its just not needed actually? 
https://github.com/kube-vip/kube-vip/pull/1251#discussion_r2301083810

Cellebyte:
fixed in #1276

Issue #1243, opened 2025-08-21
No LB service ip assignment if cp_enable = true


mauhau:
**Describe the bug**

I want to use kube-vip to provide an HA IP for kube-api and IPs for LoadBalancer services. As the system is running in a layer3 fabric, we configured it in routing table mode:

```
        env:
        - name: vip_arp
          value: "false"
        - name: port
          value: "6443"
        - name: vip_nodename
          valueFrom:
            fieldRef:
              apiVersion: v1
              fieldPath: spec.nodeName
        - name: vip_interface
          value: lo
        - name: vip_servicesinterface
          value: lo
        - name: vip_subnet
          value: "32"
        - name: dns_mode
          value: first
        - name: cp_enable
          value: "true"
        - name: cp_namespace
          value: kube-system
        - name: svc_enable
          value: "true"
        - name: svc_election
          value: "false"
        - name: svc_leasename
          value: plndr-svcs-lock
        - name: vip_routingtable
          value: "true"
        - name: address
          value: 10.1.4.50
        - name: prometheus_server
          value: :2112
        image: ghcr.io/kube-vip/kube-vip:v1.0.0
```

As long as `cp_enable` is set to `true` the LoadBalancer services are not handled:

```
2025/08/21 16:38:49 INFO kube-vip.io version=v1.0.0 build=d8877072d4a1dcb2fde2ccd22b97dbf65a1c2a73
2025/08/21 16:38:49 INFO starting namespace=kube-system Mode="Routing Table" "Control Plane"=true Services=true
2025/08/21 16:38:49 WARN the status of the interface is unknown. Ensure your interface is ready to accept traffic, if so you can safely ignore this message interface=lo
2025/08/21 16:38:49 INFO prometheus HTTP server started
2025/08/21 16:38:49 WARN the status of the interface is unknown. Ensure your interface is ready to accept traffic, if so you can safely ignore this message interface=lo
2025/08/21 16:38:49 INFO using node name name=mgmt-01
2025/08/21 16:38:49 INFO Starting Kube-vip Manager with the Routing Table engine
2025/08/21 16:38:49 INFO destination for routes table=198 protocol=248
2025/08/21 16:38:54 INFO added backend ip=10.1.4.50
```

As soon as I set `cp_enable` is set to `false`, the services are handled:

```
2025/08/21 16:40:52 INFO kube-vip.io version=v1.0.0 build=d8877072d4a1dcb2fde2ccd22b97dbf65a1c2a73
[...]

PR #1252, opened 2025-08-27
fix for layer 3 issues with not added ip 

T-Bonhagen:
fix for [issue 1243](https://github.com/kube-vip/kube-vip/issues/1243).

for cp_enable=true and svc_enable=true issue:
just change the order of start up because cp code is not threading.

for ip issue:
just change if causes which might be the problem.

### DIFF
--- a/pkg/cluster/service.go
+++ b/pkg/cluster/service.go
@@ -181,7 +181,7 @@ func (cluster *Cluster) vipService(ctxArp, ctxDNS context.Context, c *kubevip.Co
 				}
 
 				for entry := range *backendMap {
-					log.Debug("entry.Check() for entry:", entry)
+					log.Debug("entry.Check() for entry", "entry", entry)
 					if entry.Check() {
 						log.Debug("entry.Check() true")
 						_, err = network.AddIP(true)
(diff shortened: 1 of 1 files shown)


pkg/cluster/service.go
```diff
@@ -278,15 +283,21 @@ func (cluster *Cluster) StartLoadBalancerService(ctx context.Context, c *kubevip
...
+		log.Debug("c.EnableRoutingTable:",c.EnableRoutingTable," c.EnableLeaderElection ",c.EnableLeaderElection, " c.EnableServicesElection ",c.EnableServicesElection)
 
 		if c.EnableRoutingTable && (c.EnableLeaderElection || c.EnableServicesElection) {
 			err = network.AddRoute(false)
 			if err != nil {
 				log.Warn(err.Error())
+			} else {
+				log.Info("successful add Route")
 			}
-		} else if !c.EnableRoutingTable {
```
Cellebyte, line 287: I am not sure if the swap of the condition is the behaviour wanted.

pkg/manager/manager_table.go
```diff
@@ -149,9 +134,37 @@ func (sm *Manager) startTableMode(id string) error {
 			err = sm.svcProcessor.ServicesWatcher(ctx, sm.svcProcessor.SyncServices)
 			if err != nil {
 				log.Error("Cannot watch services", "err", err)
+			}else{
+				log.Debug("watching services")
 			}
 		}
 	}
+	if sm.config.EnableControlPlane {
```
Cellebyte: can we move this part into a separate PR?
I would like to merge it and it does not change behaviour that much.

Cellebyte: Please ensure go fmt run and commits being signed.

T-Bonhagen: hi, i created a seperate pull request

pkg/cluster/service.go
T-Bonhagen: What do you think about removing the condition completely?

Cellebyte:
@T-Bonhagen sorry that it takes so long. I need to verify that adding the ip without a routing policy is fine when running in routing table mode.

pkg/cluster/service.go
Cellebyte: this part I would need to verify first, as it changes behaviour.

Cellebyte:
please rebase and fix codeql issues :)

T-Bonhagen:
anything else what i should do?

Cellebyte:
No not for now.

PR #1253, opened 2025-08-27
WIP: Add node labels with per service leader election

### LINKED ISSUE #1234: node labeling is not available with per service leader election
I am hesitant to put this ticket in as a "bug" or a "feature request" as I just don't know enough about the kube-vip code base. However what I've learnt from playing around with kube-vip on a vanilla kind cluster, as well as by looking at the code base [here](https://github.com/kube-vip/kube-vip/blob/main/pkg/manager/manager_arp.go#L147) is that 
node labeling via `EnableNodeLabeling` [here](https://kube-vip.io/docs/installation/flags/) is only available in the case of `svc_election: false`. 

```
if sm.config.EnableServicesElection {
		log.Info("beginning watching services, leaderelection will happen for every service")
		err = sm.svcProcessor.StartServicesWatchForLeaderElection(ctx)
		if err != nil {
			return err
		}
	} else {

		log.Info("beginning services leadership", "namespace", ns, "lock name", sm.config.ServicesLeaseName, "id", id)
...
					if sm.config.EnableNodeLabeling {
						applyNodeLabel(sm.clientSet, sm.config.Address, id, identity)
					}
					if identity == id {
						// I just got the lock
						return
					}
					log.Info("new leader elected", "new leader", identity)
				},
			},
		})
	}
```

https://github.com/kube-vip/kube-vip/blob/main/pkg/manager/manager_arp.go#L91

Could this functionality be introduced for per svc leadership election (leases) too? Or am I missing somethnig here?

esperoz:
this fixes #1234 
Label nodes with following flags:
        - name: svc_election
          value: "true"
        - name: enable_node_labeling
          value: "true"
Node label template:
service-provided.kube-vip.io/<service-name>.<service-namespace>: "<service IPs if any>"
On kube-vip shutdown all these labels removed from the node.

### DIFF
--- a/pkg/node/manager.go
+++ b/pkg/node/manager.go
@@ -1,0 +1,137 @@
+package node
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"strings"
+	"time"
+
+	log "log/slog"
+
+	"github.com/kube-vip/kube-vip/pkg/instance"
+	"github.com/kube-vip/kube-vip/pkg/kubevip"
+	"github.com/pkg/errors"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/kubernetes"
+)
+
+const (
+	// labelName is the name of the label that will be added to the node
+	// it is prefix for the label key before "/"
+	labelName = "service-provided.kube-vip.io"
+)
+
+// labelOperation is the operation to perform on the node labels
+type labelOperation string
+
+// labelOperation constants
+const (
+	labelOperationRemove labelOperation = "remove"
+	labelOperationAdd    labelOperation = "add"
+)
+
+// Manager is the label manager for the node
+type Manager struct {
+	// nodeName is the name of the node to manage
+	nodeName string
+
+	// clientSet is the Kubernetes client set to use
+	clientSet *kubernetes.Clientset
+}
+
+// NewManager creates a new Label Manager for the given node
+func NewManager(config *kubevip.Config, clientSet *kubernetes.Clientset) *Manager {
+	return &Manager{
+		nodeName:  config.NodeName,
+		clientSet: clientSet,
+	}
+}
+
+// AddLabel a new label to the node
+func (m *Manager) AddLabel(ctx context.Context, svc *corev1.Service) error {
+	labelKey, labelValue := generateNodeLabelKeyValue(svc)
+	return m.patchNode(ctx, labelOperationAdd, map[string]string{labelKey: labelValue})
+}
+
+// RemoveLabel a label from the node
+func (m *Manager) RemoveLabel(ctx context.Context, svc *corev1.Service) error {
+	labelKey, _ := generateNodeLabelKeyValue(svc)
+	return m.patchNode(ctx, labelOperationRemove, map[string]string{labelKey: ""})
+}
+
+// clean up the node labels
+func (m *Manager) CleanUpLabels(timeout time.Duration) error {
+	log.Debug("cleaning up labels for node", "node", m.nodeName, "timeout", timeout)
+
+	ctx, cancel := context.WithTimeout(context.Background(), timeout)
+	defer cancel()
+
+	// get the node
+	node, err := m.clientSet.CoreV1().Nodes().Get(ctx, m.nodeName, metav1.GetOptions{})
+	if err != nil {
+		return errors.Wrapf(err, "failed to get node %s", m.nodeName)
+	}
+
+	// collect all labels with the prefix to remove
+	labels := map[string]string{}
+	for k := range node.Labels {
+		if strings.HasPrefix(k, labelName) {
+			labels[k] = ""
+		}
+	}
+
+	if len(labels) == 0 {
+		log.Debug("no labels to remove for node", "node", m.nodeName)
+		return nil
+	}
+
+	// patch the node with the labels to remove
+	return m.patchNode(ctx, labelOperationRemove, labels)
+}
+
+// generateNodeLabelKeyValue generates a label key and value for the given service
+func generateNodeLabelKeyValue(svc *corev1.Service) (string, string) {
+	addresses := instance.FetchServiceAddresses(svc)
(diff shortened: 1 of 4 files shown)


Cellebyte:
This should not work when we have multi endpoint configuration where `vip_leader_election` and `svc_election` are turned off.
As kube-vip would misbehave in that case.

esperoz:
> This should not work when we have multi endpoint configuration where `vip_leader_election` and `svc_election` are turned off. As kube-vip would misbehave in that case.

It's true, labeling will work only when services processor can handle services.

isc30:
Hi, this PR seems like the final step to properly have `has-ip` labels for Cilium EgressGatewayPolicy.
Any ETA or prerelease tag where we can test it?

Cellebyte:
@esperoz CI is complaining about one error check missing and could we only enable this feature when election is used? Either per service or for all kind-vip daemons using the same lease.

esperoz:
> is complaining about one error check missing

Fixed

esperoz:
> could we only enable this feature when election is used

From my perspective, it makes sense to use node labeling with BGP. In any case, it is just a fature gate, which can be toggled on or off.

Cellebyte:
@esperoz we just need to introduce a check that if elections are turned off the feature is erroring, as all kube-vip instances would compete on the label on the service. Other than that LGTM :+1: 

editation:
@esperoz sorry, I misread that the label is set on the node object using service.Name and service.Namespace as unique identifiers. Sorry for that.

Unhappy CI :(

```log
2025/09/11 20:35:09 DEBUG [service] add label namespace=default name=kube-vip-service
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0x1e0cb8f]

goroutine 20 [running]:
github.com/kube-vip/kube-vip/pkg/node.(*Manager).patchNode(0x0, {0x2927f30, 0xc000362320}, {0x25a348b, 0x3}, 0xc000403020)
	/src/pkg/node/manager.go:127 +0x48f
github.com/kube-vip/kube-vip/pkg/node.(*Manager).AddLabel(0x0, {0x2927f30, 0xc000362320}, 0xfffffffffffffffc?)
	/src/pkg/node/manager.go:56 +0xbc
github.com/kube-vip/kube-vip/pkg/services.(*Processor).SyncServices(0xc0004129a0, {0x2927f30, 0xc000362320}, 0xc000050a08)
	/src/pkg/services/services.go:68 +0x3b3
github.com/kube-vip/kube-vip/pkg/services.(*Processor).StartServicesLeaderElection.func1({0x2927f30?, 0xc000362320?})
	/src/pkg/services/leader.go:100 +0x45
created by k8s.io/client-go/tools/leaderelection.(*LeaderElector).Run in goroutine 79
	/go/pkg/mod/k8s.io/client-go@v0.33.3/tools/leaderelection/leaderelection.go:220 +0xe6
```

pkg/node/manager.go
```diff
@@ -0,0 +1,137 @@
...
+		return nil
+	}
+
+	// patch the node with the labels to remove
+	return m.patchNode(ctx, labelOperationRemove, labels)
+}
+
+// generateNodeLabelKeyValue generates a label key and value for the given service
+func generateNodeLabelKeyValue(svc *corev1.Service) (string, string) {
+	addresses := instance.FetchServiceAddresses(svc)
```
Cellebyte: that is probably why the CI fails?

`instance` is undefined here.

esperoz: In that case "instance" is pkg/instance package, so it is correct.
I've missed a case with nil clientset (eg, etcd leader election). Added handling

esperoz:
> I misread that the label is set on the node object

Yes, exactly. Thanks!

Cellebyte:
@esperoz please rebase using `—signoff`

esperoz:
> @esperoz please rebase using `—signoff`

done

Cellebyte:
@esperoz There are still stacktraces in the CI.

Please take a look 👍 

```
2025/09/15 13:17:31 DEBUG [service] add label namespace=default name=kube-vip-service
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0x1e1f76f]

goroutine 25 [running]:
github.com/kube-vip/kube-vip/pkg/node.(*Manager).patchNode(0x0, {0x293d170, 0xc00024e320}, {0x25b59a9, 0x3}, 0xc00049ad50)
	/src/pkg/node/manager.go:127 +0x48f
github.com/kube-vip/kube-vip/pkg/node.(*Manager).AddLabel(0x0, {0x293d170, 0xc00024e320}, 0xfffffffffffffffc?)
	/src/pkg/node/manager.go:56 +0xba
github.com/kube-vip/kube-vip/pkg/services.(*Processor).SyncServices(0xc0005a8000, {0x293d170, 0xc00024e320}, 0xc000050788)
	/src/pkg/services/services.go:68 +0x3b3
github.com/kube-vip/kube-vip/pkg/services.(*Processor).StartServicesLeaderElection.func1({0x293d170?, 0xc00024e320?})
	/src/pkg/services/leader.go:100 +0x45
created by k8s.io/client-go/tools/leaderelection.(*LeaderElector).Run in goroutine 19
	/go/pkg/mod/k8s.io/client-go@v0.33.3/tools/leaderelection/leaderelection.go:220 +0xe6
```

esperoz:
> Please take a look

I'm lost finally. Could you please point me to the log, may be test context will help

Cellebyte:
Hey you can find the test case logs here -> https://github.com/kube-vip/kube-vip/actions/runs/17740798247

esperoz:
> Hey you can find the test case logs here

Thanks. Refactored and added debug logs. Waiting for the next run results

Cellebyte:
Ok now it is working, but the labeler is not tested at all.
Could you add test-cases for it as well?

esperoz:
> Could you add test-cases for it as well?

Sure. First of all I will try to extend existing tests to support this feature gate

Cellebyte:
@esperoz if you want you could create a new PR with tests and we can merge this one :)

esperoz:
> @esperoz if you want you could create a new PR with tests and we can merge this one :)

Agree. It will take some time

PR #1254, opened 2025-08-27
feature/config-file-support

buji-code:
This feature allows providing kube-vip with ymal/json file based configuration using a flag and env variable. This is useful in complicated deployment cases where you need to pass different configuration per Node for a kube-vip DaemonSet.

Added test for merging different config methods (file, env, flags).


Cellebyte:
I like your PR, I always thought we had a `ConfigMap` option already. While it is implemented as flag and env variable it is not used. @thebsdbox worth reviewing :) We could deprecate `vip_configmap` as it is not used by any code anymore.

thebsdbox:
This looks good 👍 Kube-vip originally didn't depend on Kubernetes and had it's own Raft implementation (hence external configuration files back in the day) When we move to utilise the Kubernetes leader election all the configuration moved to cmdline or environment variables in the static pod manifest. So thanks for tidying this up 😄

thebsdbox (approved):

PR #1259, opened 2025-09-01
change startup order of ControlPlane and Services

T-Bonhagen:
Fix: ControlPlane was blocking main thread so Services could not start in parallel


thebsdbox (approved):
Thanks for this, wondering how this was missed.

PR #1273, opened 2025-09-16
use the new vip.LookupHost for manifest creation

Cellebyte:
* This should fix manifest creation when using dnsMode without specifying address


thebsdbox (approved):
Nice, looks good.

PR #1275, opened 2025-09-19
feat: Add leaderElection retry logic

slimm609:
Currently if there is any failure in communications when trying to get or check the leader, kubevip immediately fails. This causes the service to remove the vip and restart which can cause a short outage This makes kube-vip much more reliable before failing when running in unstable or unreliable networks.

Instead of failing, if leader election fails for "retryable" conditions, then try up to 2 additional times for a total of 3 before failing. 

- attempt 1
- sleep 1s before attempt 2
- sleep 2s before attempt 3

if the error is something like a etcd lease update or conflict, then try much faster.
- attempt 1
- sleep 250ms before attempt 2
- sleep 500ms before attempt 3

### DIFF
--- a/pkg/cluster/clusterLeaderElection.go
+++ b/pkg/cluster/clusterLeaderElection.go
@@ -14,6 +14,7 @@ import (
 	"github.com/kube-vip/kube-vip/pkg/etcd"
 	"github.com/kube-vip/kube-vip/pkg/k8s"
 	"github.com/kube-vip/kube-vip/pkg/kubevip"
+	"github.com/kube-vip/kube-vip/pkg/leaderelection"
 	"github.com/kube-vip/kube-vip/pkg/loadbalancer"
 
 	log "log/slog"
@@ -25,7 +26,7 @@ import (
 	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/cache"
-	"k8s.io/client-go/tools/leaderelection"
+	k8sleaderelection "k8s.io/client-go/tools/leaderelection"
 	"k8s.io/client-go/tools/leaderelection/resourcelock"
 	watchtools "k8s.io/client-go/tools/watch"
 )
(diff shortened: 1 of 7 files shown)


Cellebyte:
@slimm609 retry is already implemented.
In most cases when leader election is failing the storage is too slow. Then you could tweak the 3 parameters ->
```yaml
    - name: vip_leaseduration
      value: "5"
    - name: vip_renewdeadline
      value: "3"
    - name: vip_retryperiod
      value: "1"

```
I am not sure if adding another retry logic helps here. (ref: https://pkg.go.dev/k8s.io/client-go/tools/leaderelection#LeaderElectionConfig)

> Larger clusters often have a more lenient SLA for API latency. This should be taken into account when configuring the client. The rate of leader transitions should be monitored and RetryPeriod and LeaseDuration should be increased until the rate is stable and acceptably low. It's important to keep in mind when configuring this client that the tolerance to API latency varies inversely to master availability.

thebsdbox:
This is quite an interesting idea to extend the functionality of leader Elections within Kubernetes. The implementation that comes from the `client-go/tools` is quite limited. I quite like the concept of being a bit more deterministic about the election process based upon the type of issues that the current leader might be facing. 

My main concerns are taking on the technical debt of maintaining a package that is dependant on various strings for lookups etc.. But happy to discuss how we move this forward with everyone involved 😄

slimm609:
> @slimm609 retry is already implemented. In most cases when leader election is failing the storage is too slow. Then you could tweak the 3 parameters ->
> 
> ```yaml
>     - name: vip_leaseduration
>       value: "5"
>     - name: vip_renewdeadline
>       value: "3"
>     - name: vip_retryperiod
>       value: "1"
> ```
> 
> I am not sure if adding another retry logic helps here. (ref: https://pkg.go.dev/k8s.io/client-go/tools/leaderelection#LeaderElectionConfig)
> 
> > Larger clusters often have a more lenient SLA for API latency. This should be taken into account when configuring the client. The rate of leader transitions should be monitored and RetryPeriod and LeaseDuration should be increased until the rate is stable and acceptably low. It's important to keep in mind when configuring this client that the tolerance to API latency varies inversely to master availability.

We have tweaked a bunch of these and adjusted them but it is rather limited and when the failure was something like an object modification, it never fetched the new one and just kept erroring over and over for the duration of the retryPeriod so trying to extend to make a little more flexible.

Cellebyte:
Ahh so you mean explicitly this issue?
ref: https://github.com/kubernetes/kubernetes/issues/131867

But I am with @thebsdbox as the error interface needs to be maintained and updated when upstream changes it.

slimm609:
I see that client-go was updated to 0.34.1 in main but its not in the 1.0 release,  I will test to see if the issue is fixed with the newer library and see if that resolves the lease option.

slimm609:
updated the library and tested with the changes and that seemed to help a lot.   Closing

Cellebyte:
Nice :)

PR #1276, opened 2025-09-21
Refactor/kube vip 

Cellebyte:
* Move config from bgp package into kubevip package.
* Move annotation and label keys from different code paths  into there respective central places under kubevip pkg
* Refactor the gathering for Service Addresses.
* Refactor the BGP config interface.
* Moved generic IP functions used in every pkg from vip into utils pkg.


Cellebyte:
@thebsdbox I still don‘t know why it is failing CI.

Cellebyte:
@thebsdbox I think I found it and it was a timing issue and a bug which existed since long time which never bothered us until now...
The svcCtx was not set to nil when we garbage collected a service.

PR #1306, opened 2025-10-25
Feat: preserve vip on leadership loss

slimm609:
When in arp mode, when preserveVipOnLeadershipLoss is true, when a node loses leadership (or can't keep it), this will stop performing ARPs but will wait until a new leader is elected before deleting the arp. This is beneficial for cases where there may be k8s api or etcd instability.

When the flag is set to false (default) it continue to function like normal.

Current behavior:
- Leader loses election
- Deletes VIP immediately  <- VIP becomes unavailable
- New leader elected
- Configure VIP
- Starts ARPing <- VIP becomes available again

New flag behavior:
- Leader loses election
- Stops ARPing BUT keeps VIP <- vip still on current worker
- New leader elected  
- Configure VIP (or keeps it if the same node is elected) 
- Starts ARPing <- takes over traffic
- Cleans up old vip (if on a different kube-vip)

### DIFF
--- a/pkg/vip/dns.go
+++ b/pkg/vip/dns.go
@@ -51,10 +51,10 @@ func (d *ipUpdater) Run(ctx context.Context) {
 					log.Error("setting IP", "address", ip, "err", err)
 				}
 
-			// Normal VIP addition for DNS, use skipDAD=false for normal DAD process
-			if _, err := d.vip.AddIP(false, false); err != nil {
-				log.Error("error adding virtual IP", "err", err)
-			}
+				// Normal VIP addition for DNS, use skipDAD=false for normal DAD process
+				if _, err := d.vip.AddIP(false, false); err != nil {
+					log.Error("error adding virtual IP", "err", err)
+				}
 
 			}
 			time.Sleep(3 * time.Second)


thebsdbox:
Oooh, give me a day or two to have a look.

thebsdbox:
Looks good, can you add the new flag into the documentation?

slimm609:
We found an issue in testing this where ipv6 seems to cause issue but ipv4 seems to work fine.   Want to look at it some more and will update the documentation as well.

slimm609:
Its the IPv6 Duplicate Address Detection that is blocking it because it sees it still in use.   Need to see how to handle this

thebsdbox:
Good day @slimm609, how are things progressing with this work?

slimm609:
> Good day @slimm609, how are things progressing with this work?

Sorry,  was on vacation the past several days.  I am validating some fixes and will hopefully have a fix soon.

thebsdbox:
This looks good! Thanks 👍

thebsdbox:
Some linting issues, but other than that this looks good to merge!

slimm609:
looking at lint issue and updating documentation

slimm609:
https://github.com/kube-vip/website/pull/64

thebsdbox:
Awesome, I'll be creating a new release shortly !

rptaylor:
Is this applicable for services with ExternalTrafficPolicy: local (and svc_election: true) ?
I understand in that case effectively only a node hosting a pod backing the service is eligible to win the election. It sounds like this feature should still be useful in that case.

Are there any risks or downsides to enabling it?

Issue #1337, opened 2025-11-25
UPNP refresher runs even when UPNP is disabled


mkhpalm:
**Describe the bug**
Ever since [v0.8.5](https://github.com/kube-vip/kube-vip/pull/954) we are seeing UPNP logs even with a classic kube-vip configuration that only enables ARP and no LoadBalancer services.

Example logs:

```
...
2025/11/24 22:54:15 INFO Starting UPNP Port Refresher
...
2025/11/24 22:59:15 INFO [UPNP] Refreshing Instances "number of instances"=0
# repeats every 5 minutes
```

We don’t understand why it is trying to do anything with UPNP in the first place, let alone refreshing gateway port forwards to kube-vip managed ports that haven’t (and will not) be created without enabling UPNP.

It looks like the `RefreshUPNPForwards()` goroutine is started **outside** of the `sm.config.EnableUPNP` conditional. Should it be moved inside, e.g., around line 311, so it only runs if UPNP is enabled?

https://github.com/kube-vip/kube-vip/blob/v1.0.2/pkg/manager/manager.go#L313

```golang
if sm.config.EnableARP || sm.config.EnableWireguard {
    if sm.config.EnableUPNP {
        clients := upnp.GetConnectionClients(context.TODO())
        if len(clients) == 0 {
            log.Error("Error Enabling UPNP. No Clients found")
            sm.config.EnableUPNP = false
        } else {
            for _, c := range clients {
                ip, err := c.GetExternalIPAddress()
                if err != nil {
                    log.Error("unable to find IGD2 Gateway address", "err", err)
                }
                log.Info("Found UPNP IGD2 Gateway address", "ip", ip)
            }
        }
    }
    // TODO: It would be nice to run the UPNP refresh only on the leader.
    go sm.svcProcessor.RefreshUPNPForwards()
}
```

**To Reproduce**
1. Fill in the blanks and generate a manifest: 

```
docker run --rm ghcr.io/kube-vip/kube-vip:v1.0.2 manifest pod --interface eth1 --address 1.2.3.4 --controlplane --arp --leaderElection --enableLoadBalancer --lbPort 6443
```

2. Execute the test and notice it logging about UPNP refreshres

**Expected behavior**
To not see anything about UPNP or Services in the logs if those things are disabled.

thebsdbox:
This has now been rectified.

PR #1338, opened 2025-11-25
Do not run UPnP refresher if UPnP is disabled

mkhpalm:
https://github.com/kube-vip/kube-vip/issues/1337: Do not run UPNP Refresher if UPNP is disabled.


thebsdbox (approved):
Yeah. This makes sense. Not sure how we ended up with it the way we did.

Issue #1349, opened 2025-12-12
UPnP needs adjustable LeaseDuration


ivucica:
**Is your feature request related to a problem? Please describe.**

Implementation of UPNP on DD-WRT firmware on TPLink TL-WDR4300 (previous: #1335) expires a portmap within seconds. Observed on Tailscale, Kube-Vip and on manually `POST`-ed XML onto its controlpoint for portmapping.

Explicitly specifying value of `0`, as is done by `miniupnpc` / `upnpc` v2.2.4, makes it last indefinitely.

Since I knew this is on port 1780 and path `/control?WANIPConnection`, I did not observe SSDP or other traffic, and I simply observed the HTTP traffic via `sudo tshark --color -V -Y http -i eth0 'tcp port 1780'` (replacing tcpdump so I can `-Y http` for the displayfilter and `-V` to decode more of the protocol).

I observed Tailscale, Kube-Vip and upnpc traffic.

Then I mapped a particular service I happened to want exposed anyway:

```bash
#!/bin/bash
UPNP=$(mktemp)
#LEASE_DURATION_SEC=3600
LEASE_DURATION_SEC=0
cat > $UPNP <<_EOF
<?xml
        version="1.0"
        encoding="UTF-8"
        ?>
    <s:Envelope
        xmlns:s="http://schemas.xmlsoap.org/soap/envelope/"
        s:encodingStyle="http://schemas.xmlsoap.org/soap/encoding/">
        <s:Body>
            <u:AddPortMapping
                xmlns:u="urn:schemas-upnp-org:service:WANIPConnection:1">
                <NewRemoteHost></NewRemoteHost>
                <NewExternalPort>41641</NewExternalPort>
                <NewProtocol>UDP</NewProtocol>
                <NewInternalPort>41641</NewInternalPort>
                <NewInternalClient>10.1.2.3</NewInternalClient>
                <NewEnabled>1</NewEnabled>
                <NewPortMappingDescription>manual-portmap-tailscale</NewPortMappingDescription>
                <NewLeaseDuration>${LEASE_DURATION_SEC}</NewLeaseDuration>
                </u:AddPortMapping>
            </s:Body>
        </s:Envelope>
_EOF

# --http0.9 might not be required, my initial tests used the wrong endpoint.
curl -v 'http://10.1.2.1:1780/control?WANIPConnection' \
  -X 'POST' \
  -H 'Content-Type: text/xml; charset="utf-8"' \
  -H 'Connection: close' \
[...]

PR #1350, opened 2025-12-12
Support specifying UPNP lease duration as annotation on a service.

ivucica:
The annotation `kube-vip.io/upnp-lease-duration` takes in a duration where the string is parseable by `time.ParseDuration`.

There is no new configuration option that would allow setting this globally, since this new feature is most helpful in limited cases (such as incorrect implementations that have trouble with certain values).

There is a minimal test added for the value returned by the new private helper `upnpLeaseDurationForService`. Due to a lack of dependency injection in `service.Processor`'s `upnpMap`, it is infeasible to add a more complete test at this time, to check what would be passed, exactly, into `AddPinholeCtx` and `AddPortMapping`.

Fixes: #1349

### DIFF
--- a/pkg/kubevip/annotations.go
+++ b/pkg/kubevip/annotations.go
@@ -57,6 +57,9 @@ const (
 	// Enable UPNP on a Service
 	UpnpEnabled = "kube-vip.io/forwardUPNP"
 
+	// Set the UPNP lease duration for a specific service using duration format (e.g., "30s", "1h")
+	UpnpLeaseDuration = "kube-vip.io/upnp-lease-duration"
+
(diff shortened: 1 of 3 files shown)


ivucica:
I have not tested this, as I wrote the patch in a codespace, and I don't yet have a process for building or even self-hosting the image in my local cluster.

ivucica:
I don't think addressing the issue detected by Trivy would be in scope of this PR, so I will not be attempting to address this presubmit.

thebsdbox:
No worries we can raise another pr that bumps the `go.mod` separately. I'm just travelling but I'll review this pr hopefully later.

thebsdbox:
Will merge and fix trivy issues seperately.

PR #1356, opened 2025-12-15
Fixed service DNS resolve

### LINKED ISSUE #1156: IPv4 with DNS address is broken
**Describe the bug**
When using DNS address that resolves to IPv4 as vip address for a service, it fails with `invalid CIDR address: x.x.x.x/128`

**To Reproduce**
1. Add example.com that resolves to IPv4
2. Add `kube-vip.io/loadbalancerIPs: example.com` to service annotation
3. Look at logs

**Additional context**
The problem comes from `NewInstance` in `pkg/cluster/instance.go`.
The subnet is computed from the vip address, but it is still a dns name at this point, so the function wrongly assumes it is ipv6 and sets subnet to 128.

When function `NewConfig` in `pkg/vip/address.go` later on resolves the address, it does not recomputes the subnet and tries to use IPv4 address with 128 subnet and fails to parse it as valid address.

p-strusiewiczsurmacki-mobica:
This PR introduces some changes for DNS resolving and DDNS for both services and control-plane.

- DHCPv6 client was added - the DDNS and DHCP for services/CP can now work with both IPv4 and IPv6 addresses. DHCP mode can be configured with `dhcp_mode` env variable. If not provided, the value of `dns_mode` will be used.

- Dualstack  DHCP can be enabled for services by providing annotation like: 

```yaml
kube-vip.io/loadbalancerIPs: "0.0.0.0,::"
```
- DDNS can be enabled for services with annotation:
```yaml
kube-vip.io/ddns: "true"
```

In case of DualStack services value of `ipFamilyPolicy` field should be respected.

This fixes #1156 (I believe) and possibly other DNS-related issues.

### DIFF
--- a/cmd/kube-vip.go
+++ b/cmd/kube-vip.go
@@ -7,6 +7,7 @@ import (
 	"net/http"
 	"os"
 	"slices"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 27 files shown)


thebsdbox (approved):

jjschwarz:
The documentation is missing for the dhcp_mode at https://kube-vip.io/docs/installation/flags/

PR #1360, opened 2025-12-18
Fixed service deletion when service leader election is enabled

### LINKED ISSUE #1287: When we delete a DualStack Service it has two endpointslices which individually cleanup the Service object
**Describe the bug**
When deleting a Service with both IPFamilies it has two EndpointSlice objects.
This causes the second deletion to fail finding the service.

```
2025/10/05 20:10:56 DEBUG processing endpoint provider=endpointslices ip=10.244.1.8
2025/10/05 20:10:56 DEBUG watcher provider=endpointslices "service name"=kube-vip-service namespace=default endpoints=0 "last endpoint"="" "active leader election"=false
2025/10/05 20:10:56 DEBUG processing endpoint provider=endpointslices ip=fd00:10:244:1::8
2025/10/05 20:10:56 DEBUG watcher provider=endpointslices "service name"=kube-vip-service namespace=default endpoints=0 "last endpoint"="" "active leader election"=false
2025/10/05 20:10:56 ERROR unable to find/stop service uid=f5920295-b83c-4e8b-920d-f7ab1a26a003
```

**To Reproduce**
1. Run the e2e-service-tests
2. Observe the logs.

**Expected behavior**
Don't cleanup twice when the service object is deleted or garbage collected instead of the endpointslices.

**Screenshots**

 - Kube-vip Version: main

p-strusiewiczsurmacki-mobica:
This PR fixes #1287 

It turned out that the issue described by @Cellebyte did occur not only for dualstack services, but all -  kube-vip tried to delete the service in two codepaths if service election was enabled. I've changed where service is tagged as inactive and in what circumstances should be delete called, so the `unable to find/stop service` log should not be displayed now.

Additionally added  service's UID to some logs to ease service tracking a bit and updated the titles of BGP e2e tests to be more distinguishable.

### DIFF
--- a/pkg/services/leader.go
+++ b/pkg/services/leader.go
@@ -54,6 +54,12 @@ func (p *Processor) StartServicesLeaderElection(ctx context.Context, service *v1
 		},
 	}
 
+	go func() {
+		// wait for the service context to end and delete the lease then
+		<-ctx.Done()
+		p.leaseMgr.Delete(service)
+	}()
+
 	svcCtx, err := p.getServiceContext(service.UID)
 	if err != nil {
 		return fmt.Errorf("failed to get context for service %q with UID %q: %w", service.Name, service.UID, err)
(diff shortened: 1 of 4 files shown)


thebsdbox:
Ah fantastic! This has been something I've been wanting to look at for ages too! 🤩

Issue #1318, opened 2025-11-06
Multiple nodes register the IP Addresses


Danixu:
**Describe the bug**

I am using k3s as Kubernetes cluster on my homelab with three nodes. I have noticed problems with some applications that uses websockets , getting connections lost errors. Investigating the problem I have noticed that the ARP requests were sent with two different MAC addreses, and looking at the nodes I have noticed that some of the Load Balancing addresses are duplicated in two nodes.

Aragorn node:
```
# ip addr show dev enp2s0.40 | egrep "192.168.40.102|inet "
    inet 192.168.40.52/24 brd 192.168.40.255 scope global enp2s0.40
    inet 192.168.40.104/32 scope global deprecated enp2s0.40
    inet 192.168.40.111/32 scope global deprecated enp2s0.40
    inet 192.168.40.120/32 scope global deprecated enp2s0.40
    inet 192.168.40.107/32 scope global deprecated enp2s0.40
    inet 192.168.40.115/32 scope global deprecated enp2s0.40
```

Legolas node:
```
# ip addr show dev enp2s0.40 | egrep "192.168.40.102|inet "
    inet 192.168.40.51/24 brd 192.168.40.255 scope global enp2s0.40
    inet 192.168.40.70/32 scope global deprecated enp2s0.40
    inet 192.168.40.104/32 scope global deprecated enp2s0.40
    inet 192.168.40.102/32 scope global deprecated enp2s0.40
    inet 192.168.40.111/32 scope global deprecated enp2s0.40
    inet 192.168.40.120/32 scope global deprecated enp2s0.40
    inet 192.168.40.106/32 scope global deprecated enp2s0.40
    inet 192.168.40.107/32 scope global deprecated enp2s0.40
    inet 192.168.40.115/32 scope global deprecated enp2s0.40
```

The Gimli node right now doesn't have any vip address, just the node address. It happens always, one node is OK with no addresses, and the other two have some repeated.

After removing the IP addresses from the non leader node the problem has gone, but just for a while. After some hours the addresses were there again, and after a reboot also they comes again.

**To Reproduce**
There are no special steps to reproduce it. I have just installed kube-vip, disabled the k3s load balancing feature and added a service as LoadBalancer.

**Expected behavior**
To keep the addresses in just one node, the leader node.

**Environment (please complete the following information):**
 - OS/Distro: Debian 13
 - Kubernetes Version: v1.32.6+k3s1
 - Kube-vip Version: 1.0.1

[...]

PR #1375, opened 2025-12-31
Fix context propagation and panic() calls

p-strusiewiczsurmacki-mobica:
This PR is a followup to #1189

I have moved changes created by @rikatz to the current main (with some additional changes), therefore I tagged @rikatz as a co-author. Hope they will be OK with that. :)

The second commit tries to remove calls to `panic()` (mostly cosmetic change, but might also help with the cleanup I believe). This commit has to be verified carefully. I am still testing this myself, but decided to open this PR as a draft for early review.

EDIT:

~~I have found an issue with handling of endpointslices in dualstack clusters. Sometimes when IPv4 and IPv6 endpointslices were updated one after another the last known good enpoint was changed from IPv4 to IPv6 or the other way around which resulted in cancelling leader context. For this reason BGP test failed randomly. I think that should be fixed now.~~

~~Additionally I've added a pre-pull for `trafik/whoami` image that is used in the e2e test to not download it for every kind cluster, as I have already reached pull limits when running tests in the loop (plus this should also speed up the test a little).~~

~~I did the same for `plndr/e2e:v0.0.1` image in service tests.~~

This was moved to separate PRs - #1379 and #1380

### DIFF
--- a/pkg/kubevip/config_environment.go
+++ b/pkg/kubevip/config_environment.go
@@ -26,7 +26,7 @@ func ParseEnvironment(c *Config) error {
 	if env != "" {
 		logLevel, err := strconv.ParseInt(env, 10, 32)
 		if err != nil {
-			return fmt.Errorf("unable to parse environment variable [vip_loglevel], should be int")
+			return fmt.Errorf("unable to parse environment variable [vip_loglevel], should be int: %w", err)
 		}
 		c.Logging = int32(logLevel)
 	}
--- a/pkg/node/labeler/label_manager.go
+++ b/pkg/node/labeler/label_manager.go
@@ -69,7 +69,7 @@ func (m *Manager) RemoveLabel(ctx context.Context, svc *corev1.Service) error {
 func (m *Manager) CleanUpLabels(timeout time.Duration) error {
 	log.Debug("cleaning up labels for node", "node", m.nodeName, "timeout", timeout)
 
-	// create new contex for labels cleanup (independent)
+	// create new context for labels cleanup (independent)
 	ctx, cancel := context.WithTimeout(context.Background(), timeout)
 	defer cancel()
 
--- a/pkg/manager/manager_bgp.go
+++ b/pkg/manager/manager_bgp.go
@@ -59,15 +59,17 @@ func (sm *Manager) startBGP(ctx context.Context) error {
 		}
 	}()
 
-	// Shutdown function that will wait on this signal, unless we call it ourselves
-	go sm.waitForShutdown(bgpCtx, cancel, cpCluster)
-
 	if sm.config.EnableControlPlane {
 		cpCluster, err = cluster.InitCluster(sm.config, false, sm.intfMgr, sm.arpMgr)
 		if err != nil {
 			return err
 		}
+	}
 
+	// Shutdown function that will wait on this signal, unless we call it ourselves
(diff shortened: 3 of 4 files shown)


Cellebyte:
Is it possible to separate the endpointslice fix into a separate PR and rebase it on main?

p-strusiewiczsurmacki-mobica:
> Is it possible to separate the endpointslice fix into a separate PR and rebase it on main?

Sure, I'll try that.

thebsdbox:
A monstrous change 😄 .. Looks good to me!

pkg/cluster/clusterLeaderElection.go
```diff
@@ -342,7 +343,10 @@ func (cluster *Cluster) runEtcdLeaderElectionOrDie(ctx context.Context, run *run
...
 			OnNewLeader:      run.onNewLeader,
 		},
-	})
+	}); err != nil {
+		return fmt.Errorf("etcd leaderelection: %w", err)
+	}
+	return nil
 }
 
 func (sm *Manager) NodeWatcher(ctxArp context.Context, lb *loadbalancer.IPVSLoadBalancer, port uint16) error {
```
Cellebyte: I am unsure if the ctxArp should stay as a parameter as the node watcher is a general function. Maybe rename would make sense.

Cellebyte:
@p-strusiewiczsurmacki-mobica I am still reviewing your changes. Probably on Saturday I have more time to finish the Review :)

pkg/cluster/clusterLeaderElection.go
p-strusiewiczsurmacki-mobica: Done :)

p-strusiewiczsurmacki-mobica:
@Cellebyte I have added one more commit into this PR. I believe this whole PR could fix #1385 , #1318 and #996 

All those issues have one thing in common I think - the users use  `cp_enable` with `vip_leaderelection` and `svc_enable`. Now, in the the current `main` the control plane leaderelection calls `panic` on exit so kube-vip ends abruptly therefore not cleaning VIPs correctly. I was unable to reproduce this behavior with this PR anymore.

Cellebyte:
@p-strusiewiczsurmacki-mobica I still did not find the time for review. I will try tomorrow.
Thanks for taking the time to investigate :)

pkg/kubevip/config_environment.go
```diff
@@ -26,7 +26,7 @@ func ParseEnvironment(c *Config) error {
 	if env != "" {
 		logLevel, err := strconv.ParseInt(env, 10, 32)
 		if err != nil {
-			panic("Unable to parse environment variable [vip_loglevel], should be int")
+			return fmt.Errorf("unable to parse environment variable [vip_loglevel], should be int")
```
Cellebyte: We should do error wrapping here :)

pkg/node/labeler/label_manager.go
```diff
@@ -69,6 +69,7 @@ func (m *Manager) RemoveLabel(ctx context.Context, svc *corev1.Service) error {
 func (m *Manager) CleanUpLabels(timeout time.Duration) error {
 	log.Debug("cleaning up labels for node", "node", m.nodeName, "timeout", timeout)
 
+	// create new contex for labels cleanup (independent)
```
Cellebyte: Nit pick context

PR #1380, opened 2026-01-09
Added pre-pull for testing images + some minor e2e tweaks

p-strusiewiczsurmacki-mobica:
While working on #1375 I've tried to run the tests in a loop and I believe I've hit some rate limits for `traefik/whoami` image. This image is pulled for every cluster (node) created in the e2e tests. Decided to try to pull it once before the test and load it to into the kind cluster.

Additionally added a local build of `plndr/e2e` image to the e2e tests so the image is built and loaded instead of being pulled. At this point it turned out that the `log` package in the code was changed to `slog`, but the code was not updated, so fixed that and additionally bumped go version to `1.25.5` same as main `kub-vip` image.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -133,20 +133,25 @@ integration-tests:
 	go test -tags=integration,e2e -v ./pkg/etcd
 
 e2e-tests:
+	docker pull ghcr.io/traefik/whoami:v1.11
 	GOMAXPROCS=4 E2E_IMAGE_PATH=$(REPOSITORY)/$(TARGET):$(DOCKERTAG) go run github.com/onsi/ginkgo/v2/ginkgo --tags=e2e -v -p ./testing/e2e ./testing/e2e/etcd
 
 e2e-tests129-arp:
+	docker pull ghcr.io/traefik/whoami:v1.11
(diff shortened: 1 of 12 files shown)


Makefile
```diff
@@ -133,20 +133,25 @@ integration-tests:
 	go test -tags=integration,e2e -v ./pkg/etcd
 
 e2e-tests:
+	docker pull ghcr.io/traefik/whoami:v1.11
 	GOMAXPROCS=4 E2E_IMAGE_PATH=$(REPOSITORY)/$(TARGET):$(DOCKERTAG) go run github.com/onsi/ginkgo/v2/ginkgo --tags=e2e -v -p ./testing/e2e ./testing/e2e/etcd
 
 e2e-tests129-arp:
+	docker pull ghcr.io/traefik/whoami:v1.11
 	GOMAXPROCS=4 TEST_MODE=arp V129=true K8S_IMAGE_PATH=kindest/node:v1.29.0 E2E_IMAGE_PATH=$(REPOSITORY)/$(TARGET):$(DOCKERTAG) go run github.com/onsi/ginkgo/v2/ginkgo --tags=e2e -v -p ./testing/e2e
```
Cellebyte, line 141: could we use a newer kubernetes now?
like 1.33.x?

p-strusiewiczsurmacki-mobica: I think we can. I'll try to create PR for that.

Cellebyte: resolved as this will be done in a separate PR :)

PR #1384, opened 2026-01-13
Improve wireguard

daniel-naegele:
The goal for this pull request is to enable a WireGuard-backed loadbalancing mechanism, for clusters behind (CG)NATs and special setups.
The idea is the following: Have one fixed VPN Node which provides an IP address, this can be selfhosted or a VPN provider like ovpn.com can be used. This WireGuard Node is the only prerequisite. The node which connects to the VPN node also gets the IP assigned and thus the traffic hitting that IP. Note that only node can be connected at any time.

In my scenario, IP is public and Node A is connect to the VPN. When I set controlplane.example to that IP, I can use it for making my Kube API Server HA. If Node A  should fail, a new leader gets elected and and then Node B can connect and receives the traffic for example. Depending on the implementation, this could also easily be used to load balance traffic in the same way, for example to a nginx or traefik ingress. Maybe an integration for `Service`s of type `LoadBalancer` could be implemented.
```
           ┌────────────────────┐        
           │                    │        
           │  External Traffic  │        
           │                    │        
           └────────┬───────────┘        
                    │                    
                    │                    
             ┌──────▼───────┐            
             │              │            
      ┌──────►   VPN NODE   │            
      │      │              │            
      │      └──────────────┘            
      │                                  
      │                                  
┌─────▼─────┐  ┌──────────┐  ┌──────────┐
│           │  │          │  │          │
│  Node A   │  │  Node B  │  │  Node C  │
│           │  │          │  │          │
└───────────┘  └──────────┘  └──────────┘
```

Challenges:
- making sure that only nodes tries to connect to the VPN node.

Current state of the implementation:
- Creating and configuring a WireGuard interface (working, IP is pingable, also from other/external nodes) 
- Forwarding the traffic using nftables on port 6443 from WireGuard to the kubeapiserver running on the host (not working)

To ensure that the VPN connection dies when the pod dies, it runs with `hostNetwork: false`. This prevents a crashing pod/kubelet, and the VPN connection remaining up. I will open a separate PR for the Helm chart later as well.

But a few questions remain:
- Should the traffic coming from the wireguard be loadbalanced, for example ipvs round-robin to all healthy kubeapiservers, if so how?
- (How) should the wireguard implementation be integrated into the current one? For example into `pkg/cluster`

### DIFF
--- a/pkg/manager/manager_wireguard.go
+++ b/pkg/manager/manager_wireguard.go
@@ -2,13 +2,13 @@ package manager
 
 import (
 	"context"
-	"sync/atomic"
 	"fmt"
 	"net"
 	"os"
 	"os/signal"
 	"strconv"
 	"strings"
+	"sync/atomic"
 	"syscall"
 	"time"
 
--- a/pkg/wireguard/wireguard.go
+++ b/pkg/wireguard/wireguard.go
@@ -60,6 +60,11 @@ func (w *WireGuard) Down() error {
 
 // CreateInterface creates the wg interface in the current netns if it doesn't exist
 func (w *WireGuard) createInterface() error {
+	// Validate ListenPort is within valid range for uint32
+	if w.cfg.ListenPort < 0 || w.cfg.ListenPort > 65535 {
+		return fmt.Errorf("ListenPort %d is out of valid range (0-65535)", w.cfg.ListenPort)
+	}
+
 	genericLink := &netlink.GenericLink{
 		LinkAttrs: netlink.LinkAttrs{Name: w.cfg.InterfaceName},
 		LinkType:  "wireguard",
@@ -94,18 +99,13 @@ func (w *WireGuard) createInterface() error {
 		return fmt.Errorf("failed to set mtu %w", err)
 	}
 
-	// Validate ListenPort is within valid range for uint32
-	if w.cfg.ListenPort < 0 || w.cfg.ListenPort > 65535 {
-		return fmt.Errorf("ListenPort %d is out of valid range (0-65535)", w.cfg.ListenPort)
-	}
-	listenPortU32 := uint32(w.cfg.ListenPort)
-
 	// add rule to consult our routing table for IP packets without a special firewall mark
 	// all packets coming wireguard get that special firewall mark by wireguard itself
 	mask := uint32(0xffffffff)
 	err = netlink.RuleAdd(&netlink.Rule{
-		Table:  w.cfg.ListenPort,
-		Mark:   listenPortU32,
+		Table: w.cfg.ListenPort,
+		// the listen port range is already validated
+		Mark:   uint32(w.cfg.ListenPort), //nolint:all
 		Mask:   &mask,
 		Invert: true,
 		Family: netlink.FAMILY_V4,


thebsdbox:
I'm loving all of this !

thebsdbox:
@daniel-naegele is there much more than you want to add to this piece of work?

daniel-naegele:
> @daniel-naegele is there much more than you want to add to this piece of work?

Yeah I still got some issues regarding proper routing. Also I want to incorporate the IPVS loadbalancer, for the following failure mode: kube-vip on a certain node stays healthy, but the kubeapi server on that host goes down. The IPVS loadbalancer will then check the healthiness of the kubeapi server and remove it.

pkg/manager/manager_wireguard.go
```diff
@@ -100,18 +130,41 @@ func (sm *Manager) startWireguard(id string) error {
...
+					addr, _ := netlink.ParseAddr(cfg.Address)
+					lb, err := loadbalancer.NewIPVSLB(addr.IP.String(), 6443, "local", 10, cfg.InterfaceName, cancel, signalChan)
+
+					if err != nil {
+						log.Error("could not start ipvs", "err", err)
+						_ = wg.Down()
+						panic("could not start wireguard")
+					}
+
+					lb.AddBackend(nodeIP, 6443)
```
daniel-naegele: The IPVS is still not working. Maybe another IPVS forward type is needed, probably masquerade? `curl -k https://<ip>:6443/ -vvv` fails with connection refused.

daniel-naegele:
@thebsdbox so the Wireguard interface itself is properly working - with all required bells and whistles. However, I am struggling with the IPVS implementation, see the opened [comment](https://github.com/kube-vip/kube-vip/pull/1384#discussion_r2717813172). The workflow is:
- WireGuard creates and configures an interface.
- IPVS binds to that interface and forwards traffic to the proper backends

I think we do not need any nftable rules if the IPVS is working properly, right?

Also, I was wondering about the sysctls. The WireGuard feature needs one in particular `net.ipv4.conf.all.src_valid_mark=1`. However, we cannot set them ourselves due to the read-only FS in k8s pods. I now explicitly set them via the `securityContext` and had to enable certain unsafe sysctls in my kubelet. Thus, I did not implement the WireGuard feature to set that sysctl. However, the IVPS `masquerade` forwarding type also tries to set unsafe sysctls, e.g. `net.ipv4.vs.conntrack`. Can these sysctls even be set from inside the Pod? Even with elevated privileges, the `/proc/sys` is still read-only, if I am right. Should sysctl management be left to cluster administrators?

Cellebyte:
@daniel-naegele the sysctl can be set if the pod runs in host namespace and has privileged enabled.
Maybe we can crash if the mode requires it and it can’t be set to desired value? :) 

This would allow Admins who want to control it to still control it and Admins who don’t care to just use the feature ^^

Cellebyte:
@daniel-naegele as Kube-proxy is deprecating ipvs. We could maybe implement nftables in the future for everything currently used with ipvs.

daniel-naegele:
Ok so the nftables implementation is now working, making this PR ready (I think). My kubernetes control plane is now reachable via `curl -k https://<vpn-ip>:6443` from anywhere in the world!!

thebsdbox:
Awesome stuff! Thankyou ever so much for getting this across the line! (and thanks for being a kube-vip contributor!)

PR #1383, opened 2026-01-13
fix common lease fix from 1.0.1

slimm609:
Fix issue with hanging leader election and stalled lease

Test coverage specific to this issue
- TestManager_LeaderElectionRestartScenario - Simulates the bug from the issue

Additional tests for leases
- TestManager_Add_NewLease - Verifies new lease creation
- TestManager_Add_ExistingLease - Verifies returning existing lease
- TestManager_Delete_DecrementCounter - Verifies reference counting
- TestManager_Delete_CancelsContext - Verifies context cancellation on delete
- TestManager_Add_AfterDelete_CreatesNewLease - Critical test for the bug fix - verifies a new lease is created after delete
- TestManager_Add_DifferentServices - Verifies different services get different leases
- TestManager_Add_SameNameDifferentNamespace - Verifies namespace isolation
- TestManager_ConcurrentAccess - Verifies thread safety
- TestLease_StartedChannel - Verifies Started channel behavior
- TestGetName_WithoutAnnotation - Verifies default lease naming
- TestGetName_WithAnnotation - Verifies custom lease naming
- TestUsesCommon - Verifies common lease detection
- TestManager_CommonLeaseScenario - Verifies the common lease feature still works

  Affects issue #1377

PR #1386, opened 2026-01-14
Fixed leaderelection retry on error

### LINKED ISSUE #1377: kube-vip lost VIP due to stalled election, reproduced in v1.0.1 - v1.0.3
**Describe the bug**
kube-vip lost the VIP for the traefik LB service, so ingress was unavailable. 
I confirmed by 'ip addr show eth0' on all the k8s nodes; the VIP was not present on any node. It was working in v1.0.0 but broke in v1.0.1, and is not fixed in v1.0.3.

**To Reproduce**
1. Update helm from chart 0.9.0 to 0.9.4

**Expected behavior**
kube-vip pods restart and keep the VIP.

 - OS/Distro: Almalinux 9
 - Kubernetes Version: 1.29.10
 - Kube-vip Version: 1.0.2

**`Helm values`:**

```
nameOverride: kube-vip-lb
env:
  cp_enable: 'false'
  enable_node_labeling: true
  # Enable a separate election for each LB service, allowing them to be distributed across nodes. Required for 'externalTrafficPolicy: Local'.
  svc_election: true
envValueFrom:
  # Explicitly use the k8s node name (kubectl get node) as the identifier for kube-vip pods,
  # so they match endpoint slices when using per-service elections and ExternalTrafficPolicy: local
  # See https://github.com/kube-vip/kube-vip/pull/811
  vip_nodename:
    fieldRef:
      fieldPath: spec.nodeName
resources:
  requests:
    cpu: 5m
    memory: 30Mi
  limits:
    cpu: 50m
    memory: 100Mi
securityContext:
  allowPrivilegeEscalation: false
  privileged: false
  seccompProfile:
    type: RuntimeDefault
podMonitor:
  enabled: true
```

**Additional context**
After the upgrade, I restarted the daemonset, and then kube-vip pods restarted and recovered the VIP.
I could not reproduce it when doing the same upgrade on another cluster ...

p-strusiewiczsurmacki-mobica:
This PR should fix #1377 

It turned out that election and service context were not properly handled in case of election failure.

Additionally, this should fix an issue reported in the same thread about not electing a leader for service when endpoint moves from one node to another.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -157,7 +157,7 @@ e2e-tests129-bgp:
 
 e2e-tests129: e2e-tests129-arp e2e-tests129-rt e2e-tests129-bgp
 
-service-tests: 
+service-tests:
 	$(MAKE) -C testing/e2e/e2e dockerLocal
 	E2E_IMAGE_PATH=$(REPOSITORY)/$(TARGET):$(DOCKERTAG) go run ./testing/services -Services -simple -deployments -leaderActive -leaderFailover -localDeploy -egress -egressIPv6 -dualStack
 
--- a/pkg/endpoints/endpoints.go
+++ b/pkg/endpoints/endpoints.go
@@ -10,6 +10,7 @@ import (
 	"github.com/kube-vip/kube-vip/pkg/endpoints/providers"
 	"github.com/kube-vip/kube-vip/pkg/instance"
 	"github.com/kube-vip/kube-vip/pkg/kubevip"
+	"github.com/kube-vip/kube-vip/pkg/lease"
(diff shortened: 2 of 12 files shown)


Cellebyte:
Hey, I am in favor of merging #1383 to fix the issue with the stalled lease in #1377.

I would like if you could rebase your additional changes then on top of that. :)

p-strusiewiczsurmacki-mobica:
@Cellebyte done :)

This should fix the second reported issue with leader election with local policy when endpoint changes the node.

Cellebyte (approved):
LGTM. We should ask in #1377 if someone could test it. :)

PR #1387, opened 2026-01-15
Bumped e2e test kubernetes version to 1.35.0

p-strusiewiczsurmacki-mobica:
As per @Cellebyte request I've bumped the kubernetes version used in the e2e test to 1.35.0.

1.35.0 was selected arbitrarily as the most up-to-date version available, but we can change that of course. This can be also set 'on the go' with `K8S_VERSION` env variable
.
Deleted the targets `e2e-test129*` but kept the `V129` variable (and the according code) for now as true by default, if anyone would like to test with something older for any reason, but I believe that we might as well delete this as versions older than 1.32 has already reached theirs EOLs.

### DIFF
--- a/.github/workflows/ci-pull-request.yaml
+++ b/.github/workflows/ci-pull-request.yaml
@@ -25,17 +25,17 @@ jobs:
         run: make dockerx86Local
       - name: Run Manifest generation tests
         run: make manifest-test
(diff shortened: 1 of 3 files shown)


Cellebyte:
I think this needs a rebase :)

PR #1390, opened 2026-01-22
Fix IP refresh when using FQDN for VIP

### LINKED ISSUE #1389: BGP mode -
**Describe the bug**
Using bgp mode. 
if the Address env is set to an ipv4 address Peering is established and local arp table is populated.
ip addr show:
interface eth0
 inet x.x.x.x/32 scope global eth0

if the Address env is set to an fqdn  Peering is established for about 60s and fails.
ip addr show:
interface eth0
 inet x.x.x.x/32 scope global eth0
ip addr show:
  inet 10.35.230.2/32 scope global deprecated dynamic eth0
       valid_lft 60sec preferred_lft 0sec

Notice the  valid_lft  will decrease from 60 to zero, and entry will disappear.

**To Reproduce**
kube-vip yaml provided below tpo reproduce.

**Expected behavior**
Using Address as ipv4 or FQDN should not modify the behavior on how the arp table is populated.

 - OS/Distro: Centos9
 - Kubernetes Version: 1.33.5
 - Kube-vip Version: 0.9.2 - tried 1.0.1 as well

**`Kube-vip.yaml`:**
ApiVersion: v1
kind: Pod
metadata:
  name: kube-vip
  namespace: kube-system
spec:
  containers:
  - args:
    - manager
    env:
    - name: vip_arp
      value: "false"
    - name: vip_loglevel
      value: "10"
    - name: bgp_enable
      value: "true"
    - name: bgp_as
      value: "12345"
    - name: bgp_routerid
      value: "x.x.x.x"
    - name: bgp_interface
      value: "eth0"
    - name: bgp_sourceif
      value: "eth0"
    - name: bgp_multihop
      value: "true"
    - name: bgp_peers
      value: "1.1.1.1:12345:xxxxxxxxxxxxxxxxxx:True,"
    - name: port
      value: "6443"
    - name: cp_enable
      value: "true"
    - name: cp_namespace
      value: kube-system
    - name: address
      value: 1.2.3.4
    - name: dns_mode
      value: "ipv4"
    - name: prometheus_server
      value: :2112
    image: xxx/kube-vip/kube-vip:v0.9.2
    imagePullPolicy: IfNotPresent
    name: kube-vip
    resources: {}
    securityContext:
      capabilities:
        add:
        - NET_ADMIN
        - NET_RAW
    volumeMounts:
      name: kubeconfig
    - mountPath: /etc/hosts
      name: etchosts
  hostNetwork: true
  volumes:
  - hostPath:
      path: /etc/kubernetes/admin.conf
    name: kubeconfig
  - hostPath:
      path: /etc/kube-vip.hosts
      type: File
    name: etchosts

p-strusiewiczsurmacki-mobica:
This PR should fix #1389 

It seems that in case of CP without leader election in BGP mode the `vipService` function did not wait for the exit signal, so the DNS context was cancelled with defer of the parent function which resulted in IP updater being stopped.

Additionally, IP updater is refreshing the IP every 3 seconds, which might result in IP being deleted for a while (max 3 seconds) if lifetime expires. I have added a threshold to the `AddIP`, so if the valid lifetime is lower than the threshold the IP will be refreshed so there should be no situation when IP is deleted even for a while.

### DIFF
--- a/pkg/arp/arp.go
+++ b/pkg/arp/arp.go
@@ -169,7 +169,7 @@ func ensureIPAndSendGratuitous(instance *Instance) {
 			log.Info("deleted and recreating address with NODAD flag to skip DAD", "IP", ipString, "interface", iface)
 			// Re-add immediately without DAD check since we're recovering from DADFAILED
 			// The AddIP function will set IFA_F_NODAD flag for IPv6 addresses when skipDAD=true
-			if _, err := instance.network.AddIP(false, true, vip.NoLifetime); err != nil {
+			if _, err := instance.network.AddIP(false, true); err != nil {
 				log.Error("failed to recreate address after DADFAILED", "IP", ipString, "interface", iface, "err", err)
 			} else {
 				log.Info("successfully recreated address after DADFAILED recovery", "IP", ipString, "interface", iface)
@@ -180,7 +180,7 @@ func ensureIPAndSendGratuitous(instance *Instance) {
 	}
 
 	// Normal case: add IP with precheck and normal DAD process
-	if added, err := instance.network.AddIP(true, false, vip.NoLifetime); err != nil {
+	if added, err := instance.network.AddIP(true, false); err != nil {
 		log.Warn(err.Error())
 	} else if added {
 		log.Warn("Re-applied the VIP configuration", "ip", ipString, "interface", iface)
--- a/pkg/vip/address.go
+++ b/pkg/vip/address.go
@@ -37,7 +37,7 @@ const (
 
 // Network is an interface that enable managing operations for a given IP
 type Network interface {
-	AddIP(precheck bool, skipDAD bool, minLiftime int) (bool, error)
+	AddIP(precheck bool, skipDAD bool, minLifetime ...int) (bool, error)
 	AddRoute(precheck bool) error
 	DeleteIP() (bool, error)
 	DeleteRoute() error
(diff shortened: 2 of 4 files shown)


pkg/arp/arp.go
```diff
@@ -169,7 +169,7 @@ func ensureIPAndSendGratuitous(instance *Instance) {
 			log.Info("deleted and recreating address with NODAD flag to skip DAD", "IP", ipString, "interface", iface)
 			// Re-add immediately without DAD check since we're recovering from DADFAILED
 			// The AddIP function will set IFA_F_NODAD flag for IPv6 addresses when skipDAD=true
-			if _, err := instance.network.AddIP(false, true); err != nil {
+			if _, err := instance.network.AddIP(false, true, 0); err != nil {
```
Cellebyte: Why not a constant when it is not configurable?

pkg/vip/address.go
```diff
@@ -359,18 +359,18 @@ func (configurator *network) UpdateRoutes() (bool, error) {
 // AddIP - Add an IP address to the interface
 // precheck: if true, check if the IP already exists before adding
 // skipDAD: if true, set IFA_F_NODAD flag for IPv6 addresses to skip Duplicate Address Detection
-func (configurator *network) AddIP(precheck bool, skipDAD bool) (bool, error) {
+func (configurator *network) AddIP(precheck bool, skipDAD bool, minLiftime int) (bool, error) {
```
Cellebyte:
```suggestion
func (configurator *network) AddIP(precheck bool, skipDAD bool, minLifetime int) (bool, error) {
```
Lifetime

pkg/arp/arp.go
p-strusiewiczsurmacki-mobica: You're right, that would be much more readable :) Changed.

Issue #464, opened 2022-10-14
Is it possible to share ip for apiserver and other service


way-zer:
**Is your feature request related to a problem? Please describe.**
I just open a `LoadBalancer` that using the same ip as the `apiserver` VIP. It seems the VIP can't be shared.

**Describe the solution you'd like**
Is it possible to share ip for apiserver(using 6443) and other service(80/443). Or may add a warning in https://kube-vip.io/docs/usage/kubernetes-services/#multiple-services-on-the-same-ip

mruzicka:
Hi, we would also like this feature supported.

Originally we assumed this _was_ supported and configured the same IP for both the API server / control plane (by passing it in the `address` environment variable to kube-vip) and the Kubernetes services of the `LoadBalancer` type (by setting `spec.loadBalancerIP` on our ingress controller's service). Apparently it was not a good idea, as kube-vip was managing the addresses independently, so we often ended up with two nodes with the same VIP (one installed by one of the kube-vip pods for the control plane, and the other installed by another kub-vip pod for the ingress controller service) and an unstable cluster as a result.
Then we realized that since the VIP is already managed for the control plane, we can turn off the management of the VIP for the Kubernetes services (by passing `false` in the `svc_enable` environment variable to kube-vip) and only make sure that our service has its external IP provisioned (which is the other piece of functionality - besides managing the VIP installation itself - provided by kube-vip for the services of the `LoadBalancer` type, which we obviously lost when we turned it off). The provisioning of the external IP amounts to nothing more than merging the following to the service's status:
```yaml
status:
  loadBalancer:
    ingress:
      - ip: <external IP>
```
where:
*  `<external IP>` is the IP specified by `spec.loadBalancerIP` in the service definition

We implemented it by a helm post install job, which at its core does the following:
```sh
set -eo pipefail

# the namespace in which the service is defined
NAMESPACE="$1"
# the name of the service to provision the external IP for
SERVICE="$2"

# get the element which needs to be added to the .status.loadBalancer.ingress array
# (or empty string if nothing needs to be added)
INGRESS="$(kubectl --namespace "$NAMESPACE" get service "$SERVICE" --output json | jq --raw-output --compact-output '
  .spec.loadBalancerIP as $loadBalancerIP |
  if $loadBalancerIP == null then
    ""
  else
    .status.loadBalancer.ingress |
    if (. | type) == "array" and any(.ip == $loadBalancerIP) then
      ""
    else
[...]

Issue #1129, opened 2025-04-14
kube-vip modes and how they should behave (Static Manifest & DaemonSet)


Cellebyte:
# kube-vip modes

## Deploment Strategies

`kube-vip` currently has **two** Deployment Strategies which are used widely.

* as `Static Pod Manifest` (beside `kube-apiserver`) for control-plane nodes to remove the dependency on a separate loabalancer VM or Instance
  * installed under `/etc/kubernetes/manifests/kube-vip.yaml`
* as `DaemonSet` running on a all nodes which should advertise a specific IP for a `Service` to the world.
  * installed using  e.g. `manager manifest rbac | kubectl apply -f - && manager manifest daemonset | kubectl apply -f -`

Both of these deployment strategies can use different flags to enable a election mode and advertising mode.

## Election Modes

We can either have Leader Election enabled or disabled.

* Leader Election enabled
  * `Static Pod Manifest` uses the flag `vip_leaderelection=true`
  * `DaemonSet` can use the flags `vip_leaderelection=true, svc_election=false` (don't elect per service object)
  * `DaemonSet` can also use the flags `vip_leaderelection=true, svc_election=true` (elect per service object)
* Leader Election disabled
  * `Static Pod Manifest` uses the flag `vip_leaderelection=false`
  * `DaemonSet` uses the flag `vip_leaderelection=false,svc_election=false` (don't elect at all)
  * `DaemonSet` uses the flag `vip_leaderelection=false,svc_election=true` (elect per service probably)

The K8s `leases.coordination.k8s.io` API is used to elect a leader.

To configure the common parameters for the leader election we have different variables for the `leasename` but common variables for the lease timers.
* Lease Name
  * `Static Pod Manifest` for kube-apiserver `vip_leasename=plndr-cp-lock`
  * `DaemonSet` for Services LB `svc_leasename=plndr-svcs-lock`
* Common Lease Timers
  * `vip_leaseduration`
  * `vip_renewdeadline`
  * `vip_retryperiod`

## Advertising Modes
* BGP
* RoutingTable
* Wireguard
  * UPNP
* ARP/NDP
  * UPNP
  * DHCP: 0.0.0.0 (ipv4 only?)

## Service.Spec.ExternalTrafficPolicy: Cluster|Local

Services can have one of two `ExternalTrafficPolicies` configured either `Local` or `Cluster`.
[...]

PR #1395, opened 2026-01-27
Refactored manager's code for mode selection

p-strusiewiczsurmacki-mobica:
As I am working toward common global lease (as I call this for now, requested in #464) I decided that it will be easier to maintain all the modes if common functions for all modes (ARP, RT, BGP and wireguard) will be separated and reused (for example, I don't think we need to maintain 4 separate codepaths for leaderelection etc.)

Therefore new interface `manager/worker` was introduced in this PR.

Next step would be to make all the modes accept the same leaderelection/services election configurations.

Currently supported configurations:

- ARP
  - Control plane
    - With election
    - Without election
  - Services
    - Per service election
    - Global election
- BGP
  - Control plane
    - With election
    - Without election
  - Services
    - Per service election
    - Without election
- Routing table
  - Control plane
    - Without election
  - Services
    - Per service election
    - Global election
    - Without election
- Routing table
  - Control plane
    - Not supported
  - Services
    - Per service election
    - Global election

I see no reason why all the modes (maybe except wireguard, as I am not really familiar with that one) couldn't support all the CP/Services leaderelection configurations, but would love to hear others opinion as well.


Cellebyte:
@p-strusiewiczsurmacki-mobica not all of them make sense. That is why I created the issue #1129 .

no leader election would only work for anycast ready modes like BGP or RT.

Otherwise I agree with you.

thebsdbox:
One conflict here @p-strusiewiczsurmacki-mobica :-)

p-strusiewiczsurmacki-mobica:
@thebsdbox I have added the changes for wireguard. :)

Now, please be aware that with those changes all leader elections (both for services and control plane) will use common `election` package and `etcd` based leader election will be used for both control plane and services if set so (previously only control plane was able to use `etcd` and services were always using `kubernetes` type regardless of the config).

My question is, should it be possible to configure CP and services election types separately, so one could use `etcd` and the other one `kubernetes` election at the same time?

Cellebyte:
@p-strusiewiczsurmacki-mobica maybe we can die if a combination is not supported?

As we either support from single VIP perspective the following Traffic models.
 
* [Unicast](https://en.wikipedia.org/wiki/Unicast)
  * leader election
* [AnyCast](https://en.wikipedia.org/wiki/Anycast).
  * no leader election

And some modes only work with leader election but not without it.
e.g.

* ARP
  * UPnP
  * DDNS
  * DHCP
* WireGuard
* ...

I can update #1129 if something is missing or unclear :)

p-strusiewiczsurmacki-mobica:
@Cellebyte I've decided that I will leave this  as it is (at least for now), so supported modes are not changed in this PR. :)

Cellebyte:
Ok, so as an admin you can configure what you like. With or without leader election and if the mode you choose does not support e.g running in non leader election mode it still starts but won’t work.

p-strusiewiczsurmacki-mobica:
@Cellebyte Yes, same as it is in current `main`.

PR #1405, opened 2026-02-03
Improvement: global common lease

p-strusiewiczsurmacki-mobica:
This PR should help with #464 

It introduces common global lease (based on previously introduced common lease for services). Now both CP and services (global and per service leader election) can share leases.

Configuration (assuming `kube-vip` is running in `kube-system` namespace:

- When `svc_election: false`

```
- name: vip_leasename
  value: plndr-common-lock
- name: svc_leasename
  value: plndr-common-lock
```

If `vip_leasename` and `svc_leasename` are the same, both CP and Services will use the same lease (so would run on just one node).

- When `svc_election: true`

Service can use any lease known to kube-vip. Eg. let's assume that the configured lease is:

```
- name: vip_leasename
  value: plndr-cp-lock
```

if service is deployed in the same namespace as kube-vip, one can just provide the name of the lease to the service:
For example 
```
annotations:
  kube-vip.io/leaseName: "plndr-cp-lock"
```

will tie service in `kube-system` namespace to control plane lease (`plndr-cp-lock`) in the same namespace.

Additionally **cross-namespace** leases are supported. Example - service in namespace `default` can be tied to the CP lock in `kube-system` namespace with:

```
annotations:
   kube-vip.io/leaseName: "kube-system/plndr-cp-lock"
```

This is based on #1395

### DIFF
--- a/cmd/kube-vip.go
+++ b/cmd/kube-vip.go
@@ -9,6 +9,7 @@ import (
 	"slices"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	log "log/slog"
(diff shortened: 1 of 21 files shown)


Cellebyte:
I decided to go with the mode selection first.
As it makes the code cleaner.

Regarding the LeaseID I think we should make it a proper type instead of using sprintf everywhere.

Due to squashing your refactor code you unfortunately need to rebase from main again :)

p-strusiewiczsurmacki-mobica:
@Cellebyte I have closed this PR, will open a new one for this feature (can't change source branch here).

Issue #1381, opened 2026-01-11
State of Wireguard Feature


daniel-naegele:
I like the idea of the wireguard feature (although I am not 100% sure I understand it). However, the state of documentation is rather lacking and the code does not look configurable at all (e.g. hardcoded ports for peer endpoints). I would be willing to start contributing to this feature and improve it and the according docs.

My personal use case is, that I want to create a load balancer (for my control plane) using an external VPN server like ovpn.com (which uses non-default ports for its endpoints). The VPN server has a fixed public IP address and whichever node connects receives this address. Does my use case even fit the kube-vip wireguard goal? How could I help to further implement this?

Cellebyte:
Go for it ;)
Will review it together with @thebsdbox ^^

thebsdbox:
Hello, 

First of all apologies, the Wireguard feature was something I kind of dreamt up a long time ago and I didn't really have much of a use case other than someone suggested it. So it was thrown together in the hope someone might have a use for it. I would love to work together and help you turn it into something that is actually of use 😄

daniel-naegele:
I created a [pull request](https://github.com/kube-vip/kube-vip/pull/1384). Let's continue the discussion there.

guillaumejchauveau:
Hello,

I have a related use case: expose specific services (instead of control plane) through a VPN server. Additionnaly, kube-vip ARP mode should work in parallel for LAN access.
Let me know if you would like more detail or need a volunteer for testing.

daniel-naegele:
So a HA controlplane via WireGuard is now implemented. These next few steps are needed to make the WireGuard feature as a whole releasable:
- [x] add a mode for using WireGuard for services of type `LoadBalancer` as well
- [x] modify Helm chart a bit (for example make `hostNetwork` configurable)
- [x] add documentation

Also an implementation of the "other side" of the tunnel would be interesting for testing etc.

Regarding
> Additionnaly, kube-vip ARP mode should work in parallel for LAN access.

The WireGuard feature only spins up a wg interface and adds nftable rules for proper routing. In my opinion, that could work side by side with the ARP mode, but stuff like this looks more like day 2 operations. Besides, I have no idea on how to add this to the existing ARP/BGP implementation.

daniel-naegele:
I have created two more PRs for the website and the helm chart
https://github.com/kube-vip/website/pull/70
https://github.com/kube-vip/helm-charts/pull/102
These would make the Wireguard feature usable for people other than me.

[...]

PR #1414, opened 2026-02-12
WireGuard for services

daniel-naegele:
This PR implements the WireGuard capabilities for services of type `LoadBalancer`. The implementation has changed as such, that several independent tunnels can be added. If wanted, a VIP can be shared between the controlplane and a service (multiple services on one VIP has not been tested yet). This implementation still needs some more testing and the code a lot of clean-up.

This should make #1381 feature-complete for me - at least regarding the code and implementation. 

Disclaimer: Portions of the code have been written with AI, however followed my manual review and refactoring.


Cellebyte:
Hey, sorry to bother you, but I merged the PR for mode selection now and it means your manager code needs changes. Please keep that in mind :)

thebsdbox:
This all looks good to me, all the wireguard code is all wrapped behind feature flags too. Can you fix up the merge conflicts and we should be good to go.

daniel-naegele:
I rebased my branch and the implementation is working now for my use case (single VIP shared between control plane and services). Testing has not been really thorough yet, but with a broader user base and more diverse configurations this will be easier. Multiple tunnel configurations are supported, but due to the fact that each tunnel needs non-overlapping allowed IP ranges, reachability from internet is only supported for one VIP. 

I will open new PRs for the documentation and Helm charts, respectively.

pkg/nftables/nftables.go
```diff
@@ -561,6 +562,17 @@ func ApplyAPIServerDNAT(
 		return fmt.Errorf("invalid vip or target ip")
 	}
 
+	// Determine protocol number
+	var protoNum byte
+	switch protocol {
+	case "TCP":
+		protoNum = unix.IPPROTO_TCP
+	case "UDP":
```
Cellebyte, line 570: Can we safely use the Kubernetes types here?

pkg/utils/ip.go
```diff
@@ -56,3 +59,74 @@ func IsIPv6CIDR(cidr string) bool {
...
+	if idx := strings.Index(ip, "/"); idx >= 0 {
+		return ip[:idx]
+	}
+	return ip
+}
+
+// SanitizeServiceID sanitizes a service ID to be valid for nftables chain names.
+// Only alphanumeric characters and underscores are allowed; other characters are replaced with underscores.
+// The result is truncated to 50 characters to respect nftables name length limits.
+func SanitizeServiceID(id string) string {
```
Cellebyte, line 75: I think we can try to rely on hashing here? :)

pkg/nftables/nftables.go
Cellebyte: I used them here :)
https://github.com/telekom/multi-networkpolicy-nftables/blob/b4112317156b46c4babb8d528ded3329dca08a23/pkg/server/netfilterrules.go#L1572-L1584

pkg/utils/ip.go
Cellebyte: But it should be sufficient for now.

pkg/nftables/nftables.go
Cellebyte: Also not crucial. Let's merge this and update it later.

PR #1416, opened 2026-02-12
Reverted missing change from #623

p-strusiewiczsurmacki-mobica:
PR #623 introduced restart of kube-vip when VIP for control plane cannot be added. This was lost in the meantime, so I have reintroduced this behaviour.


Cellebyte:
Merging this without CI as GitHub seems to have lost our image in the CI :(

PR #1429, opened 2026-02-20
fix: align CLI lease flag defaults with Kubernetes client-go defaults (15/10/2)

k-jun:
## Summary

The CLI flag defaults for leader election lease parameters (`--leaseDuration`,
`--leaseRenewDuration`, `--leaseRetry`) are currently set to `5/3/1` seconds,
which is significantly more aggressive than the Kubernetes client-go defaults of
`15/10/2` seconds.

This causes VIP instability (flapping, split-brain) when the API server
experiences even minor latency spikes (e.g., slow etcd disk I/O, brief network
hiccups), because the leader cannot renew its lease within the tight 3-second
deadline.

## Changes

- `cmd/kube-vip.go`: Change CLI flag defaults from `5/3/1` to `15/10/2`

## Why 15/10/2?

These are the [Kubernetes client-go leader election defaults](https://github.com/kubernetes/client-go/blob/f0b431a6e0bfce3c7c1d10b223d46875df3d1c29/tools/leaderelection/leaderelection.go#L128),
and also match the values already used as fallback defaults in
`pkg/kubevip/config_generator.go` (with a TODO comment referencing this exact
client-go link).

The kube-vip documentation Flags page also states the defaults as `15/10/2`,
but the actual CLI defaults were `5/3/1` — this PR resolves that inconsistency.


Cellebyte:
@thebsdbox any reason beside fail over time that we went with these harsh timers?

thebsdbox:
Very odd, I can't actually remember where these values came from.. I was pretty sure that the values originally selected where the default values from upstream but apparently not. Happy to merge to the actual defaults.

PR #1432, opened 2026-02-23
Added context inheritance for services

p-strusiewiczsurmacki-mobica:
Currently `serviceContext` uses `context.Background()` as a parent context. This PR adds inheritance for `serviceContext` so the context for services is created as a child of either global context or election context if per-service election is enabled. This should lead to better context cancellation in case of e.g. kube-vip's shutdown.


Cellebyte:
@p-strusiewiczsurmacki-mobica do we have now a test covering the issue. So that we never introduce such regression again? :)

p-strusiewiczsurmacki-mobica:
@Cellebyte not yet, but this is only the first PR to fix this issue (I decided to split them for easier review). I'll add the tests in one of next PRs.

Cellebyte:
As long as you planned it ;)

PR #1433, opened 2026-02-26
fix(egress): prevent unnecessary SNAT updates and fix missed updates

### LINKED ISSUE #1428: Egress SNAT rule doesn't update on targeted pod IP change
**Describe the bug**
When using kube-vip egress functionality with services that have `kube-vip.io/egress: "true"` annotation, the SNAT rules are not updated when the active endpoint (pod IP) changes. After initial deployment, egress SNAT rules work correctly. However, when a pod is recreated and receives a new IP address, kube-vip detects the endpointslice change and triggers leader election, but fails to update the iptables/nftables SNAT rules to point to the new pod IP. This causes egress traffic to stop working.

The endpoint processor detects the pod IP change and updates the `kube-vip.io/active-endpoint` and `kube-vip.io/active-endpoint-ipv6` annotations, and the service watcher sees this annotation change. However, there was no logic in the service processor to recognize egress endpoint changes as requiring SNAT rule reconfiguration, so the old SNAT rules (pointing to the deleted pod's IP) remained in place.

**To Reproduce**
1. Deploy a LoadBalancer service with egress enabled:
   ```yaml
   apiVersion: v1
   kind: Service
   metadata:
     name: my-egress-service
     annotations:
       kube-vip.io/egress: "true"
       kube-vip.io/loadbalancerIPs: "192.168.1.100"
   spec:
     type: LoadBalancer
     selector:
       app: my-app
     ports:
     - port: 80
   ```

2. Deploy pods matching the service selector

3. Verify egress is working by testing outbound connections from the pod

4. Trigger a pod IP change by deleting and recreating the pod:
   ```bash
   kubectl delete pod -l app=my-app
   # Wait for new pod to start with new IP
   ```

5. Check kube-vip logs - you'll see endpointslice change detected and leader election triggered:
   ```
   [endpoints] endpointslice change detected
   [endpoints] triggering leader election
   ```

6. Test egress again - **egress traffic now fails** because SNAT rules still point to old (deleted) pod IP

7. Verify with iptables/nftables that SNAT rules contain the old pod IP, not the new one

**Expected behavior**
When the active endpoint (pod IP) changes for an egress-enabled service:
- kube-vip should detect the annotation change for `kube-vip.io/active-endpoint`
- The old SNAT rules (pointing to old pod IP) should be removed
- New SNAT rules (pointing to new pod IP) should be created
- Egress traffic should continue working seamlessly with the new pod IP
- Logs should show the egress reconfiguration:
  ```
  [service] update egress namespace=default name=my-egress-service
  [service] updating egress configuration old_ipv4=10.244.1.5 new_ipv4=10.244.2.8
  [service] egress configuration updated successfully
  ```

 - OS/Distro: Rocky Linux 9.7
 - Kubernetes Version: v1.31.13
 - Kube-vip Version: v1.0.4

**`Kube-vip.yaml`:**
```yaml
apiVersion: apps/v1
kind: DaemonSet
metadata:
  name: kube-vip-lb
  namespace: kube-system
  labels:
    app.kubernetes.io/name: kube-vip-lb
    app.kubernetes.io/version: v1.0.4
spec:
  selector:
    matchLabels:
      app.kubernetes.io/name: kube-vip-lb
  template:
    metadata:
      creationTimestamp: null
      labels:
        app.kubernetes.io/name: kube-vip-lb
        app.kubernetes.io/version: v1.0.4
      annotations: {}
    spec:
      containers:
        - name: kube-vip
          image: ghcr.io/kube-vip/kube-vip:v1.0.4
          args:
            - manager
            - '--log=-4'
          env:
            - name: vip_arp
              value: 'true'
            - name: port
              value: '6443'
            - name: vip_nodename
              valueFrom:
                fieldRef:
                  apiVersion: v1
                  fieldPath: spec.nodeName
            - name: vip_interface
              value: eth0
            - name: lb_class_only
              value: 'false'
            - name: egress_podcidr
              value: 10.255.127.192/26 # egress-no-nat ippool
            - name: egress_servicecidr
              value: 10.255.0.0/19
            - name: svc_enable
              value: 'true'
            - name: svc_election
              value: 'true'
            - name: svc_leasename
              value: plndr-svc-lock
            - name: vip_leaderelection
              value: 'true'
            - name: vip_leaseduration
              value: '5'
            - name: vip_renewdeadline
              value: '3'
            - name: vip_retryperiod
              value: '1'
            - name: enable_node_labeling
              value: 'true'
            - name: egress_clean
              value: 'true'
            - name: egress_withnftables
              value: 'true'
            - name: enable_service_security
              value: 'false'
            - name: HOST_IP
              valueFrom:
                fieldRef:
                  apiVersion: v1
                  fieldPath: status.hostIP
            - name: prometheus_server
              value: $(HOST_IP):2113
          resources: {}
          imagePullPolicy: IfNotPresent
          securityContext:
            capabilities:
              add:
                - NET_ADMIN
                - NET_RAW
                - SYS_TIME
              drop:
                - ALL
      restartPolicy: Always
      dnsPolicy: ClusterFirst
      serviceAccountName: kube-vip
      serviceAccount: kube-vip
      hostNetwork: true
      securityContext: {}
      tolerations:
        - operator: Exists
          effect: NoSchedule
        - operator: Exists
          effect: NoExecute
  updateStrategy:
    type: RollingUpdate
    rollingUpdate:
      maxUnavailable: 1
      maxSurge: 0
  revisionHistoryLimit: 10
```

**Additional context**
Using Calico as CNI

RnkeZ:
This PR resolves #1428

  ## Problem

  When egress-enabled pods are recreated, kube-vip was experiencing two issues:

  1. **Unnecessary SNAT chain updates**: Every endpoint event triggered a full
  delete/recreate of nftables SNAT chains, even when the endpoint IP hadn't changed. This
  caused unnecessary churn and potential brief traffic interruptions.

  2. **Missed updates on pod recreation**: When a pod was deleted and recreated, switching
   back to a previously-used endpoint IP (e.g., `10.255.127.227` → `10.255.127.224` →
  `10.255.127.227` → `10.255.127.224`), the egress configuration update was incorrectly
  skipped, leaving stale SNAT rules pointing to the wrong pod IP.

  ### Root Cause

  The endpoint processor was comparing the new endpoint against stale Kubernetes service
  annotations. When the annotation update failed (due to concurrent modifications), the
  K8s object retained the old value. On subsequent endpoint changes, comparing against
  this stale value caused false matches, incorrectly skipping necessary egress updates.

  ## Solution

  ### 1. Add endpoint change detection (pkg/services/services.go)
  - Compare old and new endpoint IPs before updating egress configuration
  - Skip delete/recreate cycle when endpoints are unchanged
  - Reduces unnecessary nftables operations

  ### 2. Use ServiceSnapshot for accurate state (pkg/endpoints/endpoints.go)
  - Read the current active endpoint from the in-memory ServiceSnapshot
  - ServiceSnapshot reflects actual egress configuration state
  - Fallback to K8s annotations only if instance not found
  - Ensures accurate change detection even when K8s annotation updates fail

  ### 3. Wire egress update callback (pkg/services/watch_endpoints.go)
  - Pass egress update function to endpoint processor
  - Enables direct egress reconfiguration from endpoint events
  - Required for fix 2. to trigger updates

### DIFF
--- a/pkg/endpoints/endpoints.go
+++ b/pkg/endpoints/endpoints.go
@@ -22,11 +22,11 @@ import (
 )
 
 type Processor struct {
-	config     *kubevip.Config
-	provider   providers.Provider
-	bgpServer  *bgp.Server
-	worker     endpointWorker
-	instances  *[]*instance.Instance
+	config    *kubevip.Config
+	provider  providers.Provider
+	bgpServer *bgp.Server
+	worker    endpointWorker
+	instances *[]*instance.Instance
 }
 
 func NewEndpointProcessor(config *kubevip.Config, provider providers.Provider, bgpServer *bgp.Server,


Cellebyte:
Hey @RnkeZ is this AI code or did you write this?
Additionally there is a lot of code I don't know if it deals with the resolvement of your issue.

RnkeZ:
Hey @Cellebyte, just for context, this was talked about in this [thread](https://kubernetes.slack.com/archives/C01RXPHDATB/p1768567763809919). I did this in a rush since this bug prevented us from rolling out this feature in production. Let me know which parts are most concerning to you, and I'll try to do something about it, thanks.

Cellebyte:
@RnkeZ are you sure this is still an issue with the latest changes in main, as we had issues with the endpoint tracking and leases.

RnkeZ:
@Cellebyte, I can check it out today. I'll test it again with the latest changes.

pkg/services/services.go
```diff
@@ -68,6 +69,11 @@ func (p *Processor) SyncServices(ctx *servicecontext.Context, svc *v1.Service) e
 		if err := p.nodeLabelManager.AddLabel(ctx.Ctx, svc); err != nil {
 			return fmt.Errorf("error adding label to node: %w", err)
 		}
+	case ActionUpdate:
```
Cellebyte: I don't get why we should catch Service Updates here. As the issue is with programming endpoints not the service itself.

RnkeZ:
I've built the latest image from main, and the egress service is still reporting the old IP in the `kube-vip.io/active-endpoint` annotation after targeted pod restart:
```yaml
apiVersion: v1
kind: Service
metadata:
  name: istio-egressgateway
  namespace: istio-system
  uid: bc584958-fa98-4309-b64e-03ba79eea2c8
  resourceVersion: '50320158'
  creationTimestamp: '2026-01-21T12:54:15Z'
  labels:
    app: istio-egressgateway
    app.kubernetes.io/instance: istio
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/name: istio-egressgateway
    app.kubernetes.io/part-of: istio
    app.kubernetes.io/version: 1.0.0
    helm.sh/chart: istio-egress-1.0.0
    install.operator.istio.io/owning-resource: istio-egress-gateway
    install.operator.istio.io/owning-resource-namespace: istio-system
    istio: egressgateway
    istio.io/rev: default
    operator.istio.io/component: EgressGateways
    operator.istio.io/managed: Reconcile
    operator.istio.io/version: 1.28.2
    release: istio
  annotations:
    kube-vip.io/active-endpoint: 10.255.127.227
    kube-vip.io/active-endpoint-ipv6: ''
    kube-vip.io/egress: 'true'
    kube-vip.io/egress-internal: 'true'
    kube-vip.io/vipHost: iot1-egress-mkern-1-worker-1-12-9mqnb-gl7pt
    metallb.io/ip-allocated-from-pool: default
```
Endpoints:
```yaml
apiVersion: v1
kind: Endpoints
metadata:
  name: istio-egressgateway
  namespace: istio-system
  uid: d6e9ffe6-52a2-4e42-9269-7d8967d29009
  resourceVersion: '50320775'
  creationTimestamp: '2026-01-21T12:54:15Z'
  labels:
    app: istio-egressgateway
    app.kubernetes.io/instance: istio
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/name: istio-egressgateway
    app.kubernetes.io/part-of: istio
    app.kubernetes.io/version: 1.0.0
    helm.sh/chart: istio-egress-1.0.0
    install.operator.istio.io/owning-resource: istio-egress-gateway
    install.operator.istio.io/owning-resource-namespace: istio-system
    istio: egressgateway
    istio.io/rev: default
    operator.istio.io/component: EgressGateways
    operator.istio.io/managed: Reconcile
    operator.istio.io/version: 1.28.2
    release: istio
  annotations:
    endpoints.kubernetes.io/last-change-trigger-time: '2026-03-06T13:51:12Z'
subsets:
  - addresses:
      - ip: 10.255.127.224
        nodeName: iot1-egress-mkern-1-worker-1-12-9mqnb-gl7pt
        targetRef:
          kind: Pod
          namespace: istio-system
          name: istio-egressgateway-64f959f94f-8jsnv
          uid: 160e3b6e-6531-45ec-b060-6c80402ab4d3
    ports:
      - name: http-8099
        port: 8099
        protocol: TCP
      - name: http2
        port: 8080
        protocol: TCP
      - name: https
        port: 8443
        protocol: TCP
```

Notice the .227 reported on the service annotation, while the actual pod is running with .224
Here are the logs [kube-vip.log](https://github.com/user-attachments/files/25796541/kube-vip.log)

pkg/services/services.go
```diff
@@ -86,6 +92,38 @@ func (p *Processor) getServiceInstanceAction(svc *v1.Service) ServiceInstanceAct
 
 	for _, instance := range p.ServiceInstances {
 		if instance != nil && instance.ServiceSnapshot.UID == svc.UID {
+			log.Debug("[DEBUG] found matching service instance", "service", svc.Name, "namespace", svc.Namespace, "uid", svc.UID)
```
Cellebyte, line 90: we could just call delete here again I don't think an update is required.

pkg/services/services.go
```diff
@@ -427,6 +465,101 @@ func (p *Processor) deleteService(ctx context.Context, uid types.UID) error {
 	return nil
 }
 
+func (p *Processor) updateEgressConfiguration(ctx context.Context, svc *v1.Service) error {
```
Cellebyte, line 448: this should be part of the endpoint processor not of the service itself.

RnkeZ: Working on it

pkg/services/services.go
RnkeZ: Removed the ActionUpdate completely

pkg/services/services.go
Cellebyte: I think @thebsdbox is also working on it.
Maybe you two sync first ^^

Cellebyte: Take a look at #1447

thebsdbox: The above PR is about cleaning dangling nftables rules on startup, so probably isn't relevant to this.

Cellebyte: Ok, then @RnkeZ you can proceed :)

RnkeZ: @Cellebyte, a question about moving the updateEgressConfiguration to the endpoints processor. Most of the egress functionality is in the services package. So if I were to move updateEgressConfiguration to the endpoints processor, many of those functions (e.g., configureEgress) should be accessible from the endpoints processor. Is that something we want to do?

Cellebyte: Hmm.
You are right.
It is a Service feature.

RnkeZ: Okay, can I then leave it as is?

pkg/services/watch_endpoints.go
```diff
@@ -23,29 +23,46 @@ func (p *Processor) watchEndpoint(svcCtx *servicecontext.Context, id string, ser
...
-	})
-
 	ch := rw.ResultChan()
 
-	epProcessor := endpoints.NewEndpointProcessor(p.config, provider, p.bgpServer, &p.ServiceInstances, p.leaseMgr, p.TunnelMgr)
+	egressUpdateFunc := func(ctx context.Context, svc *v1.Service) error {
+		return p.updateEgressConfiguration(ctx, svc)
+	}
+
+	epProcessor := endpoints.NewEndpointProcessor(p.config, provider, p.bgpServer, &p.ServiceInstances, p.leaseMgr, p.clientSet, egressUpdateFunc)
```
p-strusiewiczsurmacki-mobica: I think you can just pass `p.updateEgressConfiguration` instead of creating `egressUpdateFunc`

pkg/services/services.go
```diff
@@ -432,14 +424,103 @@ func (p *Processor) deleteService(ctx context.Context, uid types.UID) error {
...
+
+	if len(errList) > 0 {
+		return fmt.Errorf("errors configuring egress: %v", errList)
+	}
+
+	// Update the service snapshot to reflect the new state
+	// NOTE: Do NOT call UpdateServiceAnnotation here - the annotation was already updated
+	// by the endpoint processor, which is why we're being called in the first place.
+	// Calling it again would create an infinite loop of Modified events.
+	i.ServiceSnapshot = svc.DeepCopy()
```
p-strusiewiczsurmacki-mobica: Service is already passed as a copy from `updateAnnotations`, so I think this could be removed to not copy it twice, or, even better I think, original service could be passed alongside IPv4 and IPv6 endpoints, and copy could be done at the beginning of the function.

pkg/services/watch_endpoints.go
RnkeZ: Updated in the latest commit.

pkg/services/services.go
RnkeZ: True svc.DeepCopy was redundant, updated in the latest commit.

pkg/endpoints/endpoints.go
```diff
@@ -18,24 +18,30 @@ import (
...
-		config:    config,
-		provider:  provider,
-		bgpServer: bgpServer,
-		instances: instances,
-		worker:    newEndpointWorker(config, provider, bgpServer, instances, leaseMgr, tunnelMgr),
+		config:           config,
+		provider:         provider,
+		bgpServer:        bgpServer,
+		instances:        instances,
+		clientSet:        clientSet,
```
Cellebyte: why do we require another clientSet here?

RnkeZ: It's used to trigger ActiveEndpoint annotation update on the service in case the endpoint has changed: https://github.com/kube-vip/kube-vip/pull/1433/changes#diff-dc9255c9039453dfda93957ebde04b4a5f721feadfa30c0da45e91c41a9cf8b6R205

Cellebyte: Yeah, but please pass it via the function.
It is not static and only required for the updateAnnotation function.
Take a look at the serviceFunc parameter passing in `AddOrModify`

RnkeZ: Updated in the latest commit. Please check if it's okay now. Did a test in my environment, and everything works as expected on the egress pod recreation.

pkg/endpoints/endpoints.go
```diff
@@ -21,27 +21,31 @@ import (
...
-	provider  providers.Provider
-	bgpServer *bgp.Server
-	worker    endpointWorker
-	instances *[]*instance.Instance
+	config           *kubevip.Config
+	provider         providers.Provider
+	bgpServer        *bgp.Server
+	worker           endpointWorker
+	instances        *[]*instance.Instance
+	egressUpdateFunc func(context.Context, *v1.Service) error
```
Cellebyte: I don't think it is necessary to add the `egressUpdateFunc` into struct for the processor.
It can just be passed to the update function directly.

Cellebyte: I don't think it is necessary to add the `egressUpdateFunc` into struct for the processor.
It can just be passed to the update function directly.

RnkeZ: Updated in the latest commit

pkg/endpoints/endpoints.go
```diff
@@ -41,7 +41,9 @@ func NewEndpointProcessor(config *kubevip.Config, provider providers.Provider, b
 
 func (p *Processor) AddOrModify(svcCtx *servicecontext.Context, event watch.Event,
 	lastKnownGoodEndpoint *string, service *v1.Service, id string,
-	serviceFunc func(*servicecontext.Context, *v1.Service, *sync.WaitGroup) error, wg *sync.WaitGroup) (bool, error) {
+	serviceFunc func(*servicecontext.Context, *v1.Service, *sync.WaitGroup) error, wg *sync.WaitGroup,
+	updateAnnotationFunc func(context.Context, string, string, *v1.Service) error,
```
Cellebyte: why is this additional function needed if we already have this function on the processor?

pkg/endpoints/endpoints.go
```diff
@@ -147,15 +149,71 @@ func (p *Processor) updateLastKnownGoodEndpoint(svcCtx *servicecontext.Context,
 	}
 }
 
-func (p *Processor) updateAnnotations(service *v1.Service, lastKnownGoodEndpoint *string) {
+func (p *Processor) updateAnnotations(service *v1.Service, lastKnownGoodEndpoint *string, 
+	updateAnnotationFunc func(context.Context, string, string, *v1.Service) error,
```
Cellebyte: why is this passed as a parameter? :)

RnkeZ: Can you please check if it's okay now? It looks like I failed to add the latest changes to the commit.

pkg/endpoints/endpoints.go
RnkeZ: Could you check now? I failed to add the latest changes to the commit.

pkg/endpoints/endpoints.go
Cellebyte: much better.

Cellebyte:
@p-strusiewiczsurmacki-mobica can you also take a brief look on this PR? :)

pkg/endpoints/endpoints.go
Cellebyte: was done :)

Issue #1418, opened 2026-02-16
Multiple nodes allocate same IP address in ARP mode


ginkel:
**Describe the bug**
I am running kube-vip in ARP mode on RKE2 with networking provided by a Hetzner VSwitch.

From time to time, packet loss to the VIP occurs. The reason seems to be that multiple nodes have allocated the VIP in question and responding to ARP requests, which confuses the Hetzner VSwitch initiating frequent IP fail-overs.

**To Reproduce**
1. Deploy kube-vip. 
2. Wait for some time.
3. Observe packet loss and multiple IP allocations.

It might be possible to provoke this issue by rebooting the k8s nodes without draining them first, but the current occurrence of the issue just happened out of nowhere.

**Expected behavior**
The VIP is assigned to a single node.

**Screenshots**
n/a

**Environment (please complete the following information):**
 - OS/Distro: Ubuntu 24.04
 - Kubernetes Version: 1.35.0
 - Kube-vip Version: 1.0.4

**`Kube-vip.yaml`:**
```
apiVersion: apps/v1
kind: DaemonSet
metadata:
  creationTimestamp: null
  labels:
    app.kubernetes.io/name: kube-vip-ds
  name: kube-vip-ds
  namespace: kube-system
spec:
  selector:
    matchLabels:
      app.kubernetes.io/name: kube-vip-ds
  template:
    metadata:
      creationTimestamp: null
      labels:
        app.kubernetes.io/name: kube-vip-ds
    spec:
      affinity:
        nodeAffinity:
          requiredDuringSchedulingIgnoredDuringExecution:
            nodeSelectorTerms:
            - matchExpressions:
              - key: node-role.kubernetes.io/master
                operator: Exists
            - matchExpressions:
              - key: node-role.kubernetes.io/control-plane
                operator: Exists
      containers:
      - args:
        - manager
        env:
        - name: vip_arp
          value: "true"
        - name: vip_interface
          value: "enp35s0.4000"
        - name: port
          value: "6443"
        - name: vip_cidr
          value: "29"
        - name: cp_enable
          value: "true"
        - name: cp_namespace
          value: kube-system
        - name: vip_ddns
          value: "false"
        - name: enableUPNP
          value: "false"
        - name: svc_enable
          value: "True"
        - name: svc_election
          value: "True"
        - name: vip_leaderelection
          value: "true"
        - name: vip_leaseduration
          value: "5"
        - name: vip_renewdeadline
          value: "3"
[...]

PR #1434, opened 2026-02-26
Added waitgroups

### LINKED ISSUE #1413: VIP not released from host interface on Pod termination (Graceful Shutdown failure) [v1.0.4]
**Describe the bug**
We are running `kube-vip` v1.0.4 as a Deployment with `hostNetwork: true` with 2 replicas.
When the active leader pod is terminated (e.g., deleted or rolling update), the process exits immediately upon receiving `SIGTERM` without performing network cleanup.

As a result, the Virtual IP (VIP) **remains assigned to the host's interface (`eth0`)**. Since the pod is gone, but the IP remains, this causes an IP conflict when the new leader (on a different node) tries to advertise the same VIP.

**To Reproduce**

1. Deploy `kube-vip` v1.0.4 using a Deployment with `hostNetwork: true` and `vip_arp: "true"` (Manifest attached below).
2. Create a Service of type `LoadBalancer` so a VIP is assigned.
3. Identify the node where the active leader pod is running and verify the VIP is on `eth0`.
4. Delete the leader pod: `kubectl delete pod kube-vip-ipv4-xxxx`.
5. Check the `eth0` interface on the *original* node.
6. **Error:** The VIP is still present on the interface, even though the pod is gone.

**Expected behavior**
When the kube-vip pod receives a termination signal, it should:

1. Release the lease.
2. Remove the VIP from the host interface.
3. Send ARP/NDP updates if necessary.
4. Exit only after cleanup is complete.

**Logs**
The logs show the shutdown sequence initiating and finishing instantly in the same second, with **no messages** indicating VIP deletion or ARP cleanup.

```text
# Log snippet:
13:21:21 INFO leaderelection.go:272] "Successfully acquired lease" lock="ingress-nginx-protected-lb/kubevip-protected-lb-ipv4"
13:21:21 INFO (svcs) adding VIP ip=10.204.0.9 interface=eth0 namespace=ingress-nginx-protected-lb name=protected-lb-ipv4
...
# Termination starts:
13:27:33 INFO Received kube-vip termination, signaling shutdown
13:27:33 INFO (svcs) restartable service watcher done uid=...
13:27:33 WARN (svcs) restartable service watcher ending uid=...
13:27:33 WARN Stopping watching services for type: LoadBalancer in all namespaces
13:27:33 INFO Shutting down kube-Vip
# Process exits immediately here. The IP 10.204.0.9 remains on the node.

# MISSING logs like:
# INFO [ARP manager] removing ARP/NDP instance name=10.204.0.9/32-eth0
```

* OS/Distro: Linux / Ubuntu 22.04
* Kubernetes Version: v1.32.4
* Kube-vip Version: 1.0.4

**`Kube-vip.yaml`:**

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: kube-vip-ipv4
  namespace: kube-vip
spec:
  replicas: 2
  selector:
    matchLabels:
      name: kube-vip-ipv4
  template:
    metadata:
      labels:
        name: kube-vip-ipv4
    spec:
      hostNetwork: true
      serviceAccountName: kube-vip
      containers:
      - name: kube-vip
        image: ghcr.io/kube-vip/kube-vip:v1.0.4
        args:
        - manager
        env:
        - name: svc_election
          value: "true"
        - name: svc_enable
          value: "true"
        - name: vip_arp
          value: "true"
        - name: vip_interface
          value: eth0
        - name: vip_leaderelection
          value: "true"
        - name: lb_class_name
          value: kube-vip.io/ipv4
        securityContext:
          capabilities:
            add:
            - NET_ADMIN
            - NET_RAW

```

**Additional context**
It seems like a regression in the graceful shutdown logic or a race condition where the main context is canceled before the network cleanup routine has a chance to execute.
Since we use `hostNetwork: true`, the OS does not automatically clean up the IP when the container dies; the application must do it explicitly.

p-strusiewiczsurmacki-mobica:
This PR should fix #1413 and #1418 

Previously not all goroutines were synchronized when kube-vip was exiting, which, I believe, resulted in IP addresses not being removed properly. This PR adds `sync.WaitGroup` for every goroutine so we could ensure that all goroutines are finished before kub-vip's process exits.

I have added e2e tests for ARP, BGP and RT modes. However only in ARP mode the test tests for the IP address presence/deletion as it seems we do not clear IP addresses on exit in BGP and RT modes. For those modes the e2e test only checks if kube-vip exits within 10 seconds (which should mean it was not force-removed by k8s).

Discussion:

I'd also like to discuss the IP removal process for each mode. Should IP addresses and routes/bgp announcements be removed on exit in RT and BGP modes as well? Maybe this should be optional and something similar to `preserveVipOnLeadershipLoss` should be  introduced for other modes? Or maybe the addresses should not be deleted as RT and BGP modes can work fine without kube-vip's presence (and ARP cannot as kube-vip has to perform ARP/NDP advertisement). I'd just like to get some more guidance on the expected outcome for each mode.

### DIFF
--- a/cmd/kube-vip.go
+++ b/cmd/kube-vip.go
@@ -9,6 +9,7 @@ import (
 	"slices"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	log "log/slog"
@@ -302,14 +303,19 @@ var kubeVipManager = &cobra.Command{
 		// Welome messages
 		log.Info("kube-vip.io", "version", Release.Version, "build", Release.Build)
 
+		wg := sync.WaitGroup{}
+		defer wg.Wait()
+
 		// create main manager context
 		ctx, cancel := context.WithCancel(cmd.Context())
 		defer cancel()
 
 		// start prometheus server
 		if initConfig.PrometheusHTTPServer != "" {
-			go servePrometheusHTTPServer(ctx, PrometheusHTTPServerConfig{
-				Addr: initConfig.PrometheusHTTPServer,
+			wg.Go(func() {
+				servePrometheusHTTPServer(ctx, PrometheusHTTPServerConfig{
+					Addr: initConfig.PrometheusHTTPServer,
+				})
 			})
 		}
 
@@ -393,13 +399,13 @@ var kubeVipManager = &cobra.Command{
 				initConfig.Interface = defaultIF.Name
 				log.Info("kube-vip bind", "interface", initConfig.Interface)
 
-				go func() {
+				wg.Go(func() {
 					if err := vip.MonitorDefaultInterface(ctx, defaultIF); err != nil {
 
 						log.Error("interface monitor", "err", err)
 						return
 					}
-				}()
+				})
 			}
 		}
 		// Perform a check on the state of the interface
@@ -458,18 +464,18 @@ func servePrometheusHTTPServer(ctx context.Context, config PrometheusHTTPServerC
 		ReadHeaderTimeout: 2 * time.Second,
 	}
 
-	go func() {
+	wg := sync.WaitGroup{}
+
+	wg.Go(func() {
 		if err = srv.ListenAndServe(); err != nil && err != http.ErrServerClosed {
 			log.Error("prometheus HTTP server", "err", err)
 			return
 		}
-	}()
+	})
 
 	log.Info("prometheus HTTP server started")
 
 	<-ctx.Done()
-
-	log.Info("prometheus HTTP server stopped")
 
 	// create prometheus shutdown context (independent of other contexts)
 	ctxShutDown, cancel := context.WithTimeout(context.Background(), 5*time.Second)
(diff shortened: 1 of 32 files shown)


Cellebyte:
@p-strusiewiczsurmacki-mobica 
When we run with vip_leaderelection: true or svc_election: true.
I think we can safely remove the IP and count on a different leader to take over the IP.
I still see the issue with RollingUpdates when running as a DaemonSet and removing the IP as kube-vip is only a control-plane we should not interrupt the dataplane with it.

As kube-proxy also does not interrupt nftables/iptables rules while being updated.
Maybe we can add a one-time option to kube-vip to cleanup and exit similar to what kube-proxy does.
This allows for a safe uninstall of kube-vip and we don't interrupt traffic during normal runtime and update.

@thebsdbox I think this is something which needs a bit more discussion.

Cellebyte:
Adding @chdxD1 here, as we got issues when assignment of IPs happens in RT or BGP mode on an interface.

chdxD1:
Shouldn't conflict with #1442, no?

Cellebyte:
@chdxD1 it does not conflict but fixes the lease tracking issue.

PR #1441, opened 2026-03-02
Add skipIPOnInterface / vip_skipiponinterface

chdxD1:
In some deployments (e.g. RoutingTable mode or BGP) kube-vip works perfectly fine without adding the IP address to an interface.

For example, with node services listening on 0.0.0.0 / ::, adding the VIP would also expose the node services to the VIP, which might not be wanted.

To not break existing deployments, a flag vip-skipiponinterface is introduced, that can be used to disable this behaviour, including cleanup.

### DIFF
--- a/pkg/cluster/service.go
+++ b/pkg/cluster/service.go
@@ -82,9 +82,16 @@ func (cluster *Cluster) vipService(ctx context.Context, c *kubevip.Config, sm *e
 		}
 
 		if !c.EnableRoutingTable {
-			// Normal VIP addition, use skipDAD=false for normal DAD process
-			if _, err = network.AddIP(false, false); err != nil {
-				return fmt.Errorf("failed to add IP address %s: %w", network.IP(), err)
+			if c.SkipAddingVIP || c.EnableRoutingTable {
+				// Ensure VIP is not on the interface (cleanup from previous runs)
+				if _, err = network.DeleteIP(); err != nil {
+					log.Warn("attempted to clean existing VIP", "err", err)
+				}
+			} else {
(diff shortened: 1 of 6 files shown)


pkg/cluster/service.go
```diff
@@ -82,9 +82,16 @@ func (cluster *Cluster) vipService(ctx context.Context, c *kubevip.Config, sm *e
 		}
 
 		if !c.EnableRoutingTable {
-			// Normal VIP addition, use skipDAD=false for normal DAD process
-			if _, err = network.AddIP(false, false); err != nil {
-				return fmt.Errorf("failed to add IP address %s: %w", network.IP(), err)
+			if c.SkipAddingVIP || c.EnableRoutingTable {
```
Cellebyte, line 85: ?

Cellebyte: if !c.EnableRoutingTable { + if c.SkipAddingVIP || c.EnableRoutingTable { ?

This cancels each other out :D

pkg/kubevip/config_types.go
```diff
@@ -127,6 +127,9 @@ type Config struct {
 	// Clean routing table of redundant routes on start
 	CleanRoutingTable bool `yaml:"cleanRoutingTable"`
 
+	// SkipAddingVIP, if true, will not add the VIP to the network interface.
+	SkipAddingVIP bool `yaml:"skipAddingVIP"`
```
Cellebyte, line 131: please be consistent in naming here :)

chdxD1:
I am closing this PR because I don't think we should even guard this with a flag.

PR #1444, opened 2026-03-04
Cleanup of channels

p-strusiewiczsurmacki-mobica:
This PR is a followup to the series of context related PRs.

After adding the waitgroups in #1434 I believe that some of the channels that were used for synchronization are not relevant anymore, so I've changed those to use passed context instead.

Additionally changed the way `SIGINT` signal is sent - now using `sync.Once` so it would be not possible for a goroutine to hang on the channel operation.


pkg/cluster/clusterLeaderElection.go
```diff
@@ -19,7 +16,7 @@ import (
 
 // StartCluster - Begins a running instance of the Leader Election cluster
 func (cluster *Cluster) StartCluster(ctx context.Context, c *kubevip.Config,
-	em *election.Manager, bgpServer *bgp.Server, leaseMgr *lease.Manager) error {
+	em *election.Manager, bgpServer *bgp.Server, leaseMgr *lease.Manager, killFunc func()) error {
```
Cellebyte, line 19: @p-strusiewiczsurmacki-mobica where is this killfunc defined?
What happens if it is not defined?

p-strusiewiczsurmacki-mobica: It's defined [here](https://github.com/p-strusiewiczsurmacki-mobica/kube-vip/blob/f3ea8420b32d6f4c693d4154f99b02b757473538/pkg/manager/manager.go#L411) and then it's passed down. If `nil` will be passed (although it should be not) kube-vip will crash, so I can add a nil check for those calls.

Cellebyte: If it is always passed and it can never be `nil` I am fine with that. Was just curious what it is for and why we need to pass the function down when we already have context

p-strusiewiczsurmacki-mobica: It's used to send SIGINT signal and exit the kube-vip. Previously the `signalChan` was passed (and it was not even passed properly I beleive), but I've decided to pass a function instead to use `sync.Once` as multiple writes to `signalChan` could result in a deadlock.

Cellebyte: Thanks for the clarification 👍

PR #1446, opened 2026-03-05
fix: skip UPNP refresh logging when no service instances

ohauer:
Prevents log spam in control-plane-only deployments where ServiceInstances is always empty. The UPNP refresh goroutine still runs but no longer logs every 5 minutes when there's nothing to refresh.

Fixes noise in SIEM systems and log aggregators.


Cellebyte (approved):
LGTM

PR #1463, opened 2026-03-10
Refactored service handling

p-strusiewiczsurmacki-mobica:
>// Architecture walkthrough: (Had to do this as this code path is making my head hurt)
> \- pkg/services/processor.go, line 195

This PR aims to cleanup the service processor code. The major change is that in old version the `serviceFunc` was executed differently in every mode and also endpoint watcher's execution was dependent on the mode. I believe that service and endpoint handling should be similar in all modes, and only the actual action (e.g. adding BGP host in BGP mode or configuring routes in RT)  should depend on the mode. Therefore i propose to run `serviceFunc` and endpoint watcher always, regardless of the selected mode.

As kube-vip should only reconcile services that have endpoints (either local for local policy or any for cluster policy) synchronization channel was added to `servicecontext.Context` to achieve this.

I have also changed where the `Instance` for the service is being created which made possible to get rid of `FindServiceInstanceWithTimeout` function.

Additionally minor changes like removing `servicecontext.Context.IsActive` and `servicecontext.Context.HasEndpoints` fields were introduced as those seemed to be redundant now.

> **IMPORTANT**: This PR might be very, very risky, as it changes very important part of the service processing.

### DIFF
--- a/pkg/endpoints/endpoints_bgp.go
+++ b/pkg/endpoints/endpoints_bgp.go
@@ -102,7 +102,7 @@ func (b *BGP) clearBGPHosts(ctx context.Context, service *v1.Service) {
 	ClearBGPHosts(ctx, service, b.instances, b.bgpServer)
 }
 
-func (b *BGP) setInstanceEndpointsStatus(_ *v1.Service, _ []string) error {
+func (b *BGP) setInstanceEndpointsStatus(_ context.Context, _ *v1.Service, _ []string) error {
 	return nil
 }
 
--- a/pkg/endpoints/endpoints_routing_table.go
+++ b/pkg/endpoints/endpoints_routing_table.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"net"
+	"sync"
 
 	log "log/slog"
 
@@ -17,6 +18,7 @@ import (
 
 type RoutingTable struct {
 	generic
+	mtx      sync.Mutex
 	routeMgr *route.Manager
 }
 
@@ -27,19 +29,19 @@ func newRoutingTable(generic generic, routeMgr *route.Manager) endpointWorker {
 	}
 }
 
-func (rt *RoutingTable) processInstance(ctx *servicecontext.Context, service *v1.Service) error {
-	instance := instance.FindServiceInstance(service, *rt.instances)
-	if instance != nil {
-		for _, cluster := range instance.Clusters {
+func (rt *RoutingTable) processInstance(svcCtx *servicecontext.Context, service *v1.Service) error {
+	inst := instance.FindServiceInstance(service, *rt.instances)
+	if inst != nil {
+		for _, cluster := range inst.Clusters {
 			for i := range cluster.Network {
-				if !ctx.IsNetworkConfigured(cluster.Network[i].IP()) && cluster.Network[i].HasEndpoints() {
+				if !svcCtx.IsNetworkConfigured(cluster.Network[i].IP()) && cluster.Network[i].HasEndpoints() {
 					if err := rt.routeMgr.Add(lease.ServiceNamespacedName(service), cluster.Network[i], false, true); err != nil {
 						return fmt.Errorf("[%s] error adding route: %s", rt.provider.GetLabel(), err.Error())
 					} else {
 						log.Info("added route", "provider",
 							rt.provider.GetLabel(), "ip", cluster.Network[i].IP(), "service name", service.Name, "namespace",
 							service.Namespace, "interface", cluster.Network[i].Interface(), "tableID", rt.config.RoutingTableID)
-						ctx.ConfiguredNetworks.Store(cluster.Network[i].IP(), true)
+						svcCtx.ConfiguredNetworks.Store(cluster.Network[i].IP(), true)
 					}
 				}
 			}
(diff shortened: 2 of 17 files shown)


pkg/endpoints/endpoints_bgp.go
```diff
@@ -24,8 +24,8 @@ func newBGP(generic generic, bgpServer *bgp.Server) endpointWorker {
 }
 
 func (b *BGP) processInstance(svcCtx *servicecontext.Context, service *v1.Service) error {
-	if instance := instance.FindServiceInstance(service, *b.instances); instance != nil {
-		for _, cluster := range instance.Clusters {
+	if inst := instance.FindServiceInstance(service, *b.instances); inst != nil {
```
Cellebyte: please keep this lines as they were :)

pkg/endpoints/endpoints_routing_table.go
```diff
@@ -26,12 +28,12 @@ func newRoutingTable(generic generic) endpointWorker {
 	}
 }
 
-func (rt *RoutingTable) processInstance(ctx *servicecontext.Context, service *v1.Service) error {
-	instance := instance.FindServiceInstance(service, *rt.instances)
-	if instance != nil {
-		for _, cluster := range instance.Clusters {
+func (rt *RoutingTable) processInstance(svcCtx *servicecontext.Context, service *v1.Service) error {
+	inst := instance.FindServiceInstance(service, *rt.instances)
```
Cellebyte, line 33: any reason for the rename here?

p-strusiewiczsurmacki-mobica: Just to avoid confusion as we have a package called `instance`. But I can change that. :)

Cellebyte: No keep it :) I saw in later code paths you adapted it :)

p-strusiewiczsurmacki-mobica:
@Cellebyte @thebsdbox 

I have a question regarding this PR. Currently, for example in ARP mode, if `extarnalTrafficPolicy: Cluster` is used for a service, VIP will be added to the host even if there are no endpoints backing the service. Is there any specific reason why it should be that way?

Cellebyte:
@p-strusiewiczsurmacki-mobica I think that is a bug it should only be on one Node when ARP is used. Regardless if it is Cluster or Local.

p-strusiewiczsurmacki-mobica:
> @p-strusiewiczsurmacki-mobica I think that is a bug it should only be on one Node when ARP is used. Regardless if it is Cluster or Local.

That's true, but I mean something different. I use single node cluster for example. If I will now add a service that has no endpoints, if `extarnalTrafficPolicy: Local` is used, then kube-vip will not add the service VIP to the host's interface (which is correct, as there are no local endpoints). But if I'll use `extarnalTrafficPolicy: Cluster` even if there are no endpoints for the service (like, no pods in the entire cluster which means service's endpointslice/endpoints are empty), the VIP is added to the host.

Cellebyte:
Hmm. You are right.
Maybe it makes sense to have it behave the same way.

I also don't know if https://github.com/kube-vip/kube-vip/issues/1475 is somehow related to your discovered issue.

p-strusiewiczsurmacki-mobica:
@Cellebyte I think this is ready for a review

Cellebyte:
@p-strusiewiczsurmacki-mobica I did a brief review.
It looks much cleaner.
I think we should wait a bit before we merge it.

Maybe a test-build is beneficial to test it in some of our environments.

p-strusiewiczsurmacki-mobica:
> @p-strusiewiczsurmacki-mobica

> I think we should wait a bit before we merge it.

I agree, as other PRs might have to be rebased if this is merged.


> 
> Maybe a test-build is beneficial to test it in some of our environments.

That would be great!

thebsdbox:
I've just spend the last however long reading through this PR, it looks good to me. Are we considering an RC build or something to get it into peoples hands?

thebsdbox (approved):

Cellebyte:
Maybe we could add an rc pipeline for pre releases? I think this could help us to iterate faster on main

thebsdbox:
What is the general consensus to get this approved and merged in?

Cellebyte:
@thebsdbox we should introduce a pre-release pipeline.
But from my perspective we can merge it now and do a pre-release in the main branch.

Cellebyte (approved):
LGTM

p-strusiewiczsurmacki-mobica:
OK, I have just rebased this to the current main just to be sure.

Cellebyte:
@p-strusiewiczsurmacki-mobica the unnumbered test still failed.
```
2026/05/12 12:44:04 INFO kube-vip.io version=v1.1.2 build=cf642fa18f96e57fed0054be5a3b1e5fbb40c435
2026/05/12 12:44:04 INFO starting namespace=kube-system Mode=BGP "Control Plane"=true Services=true
2026/05/12 12:44:04 INFO using node name name=kube-vip-test-bgp-v6289574319-bgp-ds-v6-control-plane
2026/05/12 12:44:04 DEBUG Using external Kubernetes configuration from file path=/etc/kubernetes/admin.conf
2026/05/12 12:44:04 DEBUG No Node annotations to parse
2026/05/12 12:44:04 INFO starting Kube-vip Manager mode=BGP
2026/05/12 12:44:04 INFO Starting the BGP server to advertise VIP routes to BGP peers
2026/05/12 12:44:04 INFO Shutting down Kube-Vip
2026/05/12 12:44:04 ERROR start manager err="failed to configure BGP mode: starting BGP server: failed to get link-local address of interface eth0: no ipv6 link-local neighbor found"
```

p-strusiewiczsurmacki-mobica:
I think that it might be that neighbors are not yet populated when `GetIPv6LinkLocalNeighborAddress` is executed. I have added a retry for that in separate commit to this PR.

w13915984028:
Just FYI:

After this breaking change, Harvester loadbalancer or third-party placeholder serivces are all not working as usual. We will adpat to this change via https://github.com/harvester/harvester/issues/11574, thanks.

Issue #1466, opened 2026-03-12
v1.1.0: BGP peers added then immediately de-configured in control plane mode


cooperlees:
## Description

After upgrading from v1.0.4 to v1.1.0, BGP peers are added and then immediately deleted/de-configured, preventing BGP sessions from establishing. Rolling back to v1.0.4 resolves the issue — sessions establish within seconds.

## Environment

- **kube-vip mode**: Static pod, BGP control plane (`cp_enable=true`, `bgp_enable=true`)
- **Kubernetes**: 3 control plane nodes
- **BGP peers**: Two upstream FRRouting routers (IPv6, separate ASNs)
- **VIP**: IPv6 address on loopback interface
- **Previous version**: v1.0.4 (working)
- **Broken version**: v1.1.0

## kube-vip logs (v1.1.0)

```
2026/03/12 20:28:42 INFO kube-vip.io version=v1.1.0 build=64880b62ab6132a44b2d4e0b6114c39572a42fee
2026/03/12 20:28:42 INFO starting namespace=kube-system Mode=BGP "Control Plane"=true Services=false
2026/03/12 20:28:42 INFO starting Kube-vip Manager mode=BGP
2026/03/12 20:28:42 INFO Starting the BGP server to advertise VIP routes to BGP peers
time="2026-03-12T20:28:42Z" level=info msg="Add a peer configuration" Key="fd00:4::2" Topic=Peer
time="2026-03-12T20:28:42Z" level=info msg="Add a peer configuration" Key="fd00:4::3" Topic=Peer
time="2026-03-12T20:28:42Z" level=info msg="Delete a peer configuration" Key="fd00:4::2" Topic=Peer
time="2026-03-12T20:28:42Z" level=info msg="Delete a peer configuration" Key="fd00:4::3" Topic=Peer
2026/03/12 20:28:42 ERROR bgp server hasn't started yet
2026/03/12 20:28:42 INFO [BGP] peer="...session_state:IDLE..."
2026/03/12 20:28:42 INFO [BGP] peer="...session_state:ACTIVE..."
```

Note: peers are added then immediately deleted within the same second. No further log output after this — BGP never progresses beyond ACTIVE.

## FRRouting peer view (upstream router)

```
BGP neighbor is fd00:4::21, remote AS 65004, local AS 65069, external link
  BGP state = Active
  Connections established 10; dropped 10
  Last reset 00:04:11, Notification received (Cease/Peer De-configured)
```

[...]

PR #1467, opened 2026-03-13
fix: #1466 print bgp config + change order of AddPeer and GetPeer

Cellebyte:
* This tries to fix a race condition in the `Start` function of the BGP mode.


Cellebyte:
does not fix the issue.

PR #1469, opened 2026-03-15
Enhance WireGuard nftables and endpoint handling

daniel-naegele:
This PR improves a number of things in the current WireGuard implementation
- makes this mode suitable for running in hostNetwork mode (thus this [PR](https://github.com/kube-vip/helm-charts/pull/101) is no longer needed, but still this [PR](https://github.com/kube-vip/helm-charts/pull/102))
- does not masquerade source IPs for local endpoints (source IP preservation is important for example for mail servers)
- SCTP support
- load balancing for services with multiple endpoints
- utilizes the kubernetes service in the default namespace to implement kube api server functionality
- implements the following [improvement](https://github.com/kube-vip/kube-vip/pull/1414#discussion_r2873808476) 

What could be improved:
- [this suggestion](https://github.com/kube-vip/kube-vip/pull/1414#discussion_r2873820249), but idk which hashing algorithm should be used. also keeping the service name in the nftables makes debugging easier
- see the first comment in the nftable example, although this should be negligible 

These achievements were made possible by the new policy routing setup and nftables implementation. The nftables implementation now only uses one set of chains per tunnel and manages different services and endpoints via sets and rules. Also, the rules for routing packets back to the tunnel is no blanket rule anymore (i.e. "not wg fwmark -> our routing table -> our tunnel"). Instead, a conntrack mark is added for connections coming from the tunnel which is then used to route reply packets back to the tunnel. This results for example in the following nftables chains
```
table ip kube_vip_v4 {
	set kvip_accept_tunnel1 {
		type inet_service
		elements = { 80, 6443 }
	}

	set kvip_snat_tunnel1 {
		type inet_service
		elements = { 6443, 8000  }
	}

	set kvip_masq_ip_tunnel1 {
		type ipv4_addr
		elements = { 10.42.0.94, 100.64.0.1,
			     100.64.0.2 }
	}

	set kvip_masq_pt_tunnel1 {
		type inet_service
		elements = { 6443, 8000  }
	}

	map dnat_controlplane {
		type 0 : ipv4_addr
		elements = { 0 : 100.64.0.1,
			     1 : 100.64.0.2 }
	}

	# Not needed if the target is only one, should maybe only be created if more than one target exists
	map dnat_kube_system_traefik_p80 {
		type 0 : ipv4_addr
		elements = { 0 : 10.42.0.94 }
	}

	chain kvip_dnat_tunnel1 {
		type nat hook prerouting priority dstnat - 5; policy accept;
		iifname "tunnel1" tcp dport 6443 dnat to numgen random mod 2 map @dnat_controlplane:6443
		iifname "tunnel1" tcp dport 80 dnat to 10.42.0.94:8000
	}

	chain kvip_input_tunnel1 {
		type filter hook input priority filter; policy accept;
		iifname "tunnel1" th dport @kvip_accept_tunnel1 accept
	}

	chain kvip_post_tunnel1 {
		type nat hook postrouting priority srcnat; policy accept;
		oifname "tunnel1" th sport @kvip_snat_tunnel1 snat to VIP
		oifname != "tunnel1" ip daddr @kvip_masq_ip_tunnel1 th dport @kvip_masq_pt_tunnel1 masquerade
	}

	chain kvip_mpre_tunnel1 {
		type filter hook prerouting priority mangle; policy accept;
		# This mark is the wireguard listen port + an offset for distinction with the fwmark set by wireguard itself
		iifname "tunnel1" ct mark set 0x0001ca6c
		iifname != "tunnel1" ct mark 0x0001ca6c meta mark set ct mark
	}

	chain kvip_mout_tunnel1 {
		type route hook output priority mangle; policy accept;
		ct mark 0x0001ca6c meta mark set ct mark
	}
}
``` 

The policy routing setup also creates a rule to lookup our desired routing table if the packet mark matches the one we calculated (the ct mark). This should make it possible, to have several tunnels with overlapping `AllowedIPs` ranges and the `0.0.0.0/0` range itself, as only incoming connections from the tunnel interface get routed back (ensured by the ct mark stuff).

This makes https://github.com/kube-vip/kube-vip/issues/1381 complete.

Sorry for the large diff :(

### DIFF
--- a/pkg/nftables/nftables.go
+++ b/pkg/nftables/nftables.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"log/slog"
 	"net"
+	"slices"
 	"strconv"
 	"strings"
 
@@ -708,7 +709,7 @@ func EnsureTunnelInfrastructure(wgIf string, vipIP string, IPv6 bool, tunnelList
 		Chain: inputChain,
 		Exprs: []expr.Any{
 			&expr.Meta{Key: expr.MetaKeyIIFNAME, Register: 1},
-			&expr.Cmp{Op: expr.CmpOpEq, Register: 1, Data: append([]byte(wgIf), 0)},
+			&expr.Cmp{Op: expr.CmpOpEq, Register: 1, Data: ifname(wgIf)},
 			&expr.Payload{DestRegister: 1, Base: expr.PayloadBaseTransportHeader, Offset: 2, Len: 2},
 			&expr.Lookup{SourceRegister: 1, SetName: acceptSetName},
 			&expr.Verdict{Kind: expr.VerdictAccept},
@@ -722,7 +723,7 @@ func EnsureTunnelInfrastructure(wgIf string, vipIP string, IPv6 bool, tunnelList
 		Chain: postroutingChain,
 		Exprs: []expr.Any{
 			&expr.Meta{Key: expr.MetaKeyOIFNAME, Register: 1},
-			&expr.Cmp{Op: expr.CmpOpEq, Register: 1, Data: append([]byte(wgIf), 0)},
+			&expr.Cmp{Op: expr.CmpOpEq, Register: 1, Data: ifname(wgIf)},
 			&expr.Payload{DestRegister: 1, Base: expr.PayloadBaseTransportHeader, Offset: 0, Len: 2},
 			&expr.Lookup{SourceRegister: 1, SetName: snatSetName},
 			&expr.Immediate{Register: 1, Data: ipToBytes(vip, IPv6)},
@@ -744,7 +745,7 @@ func EnsureTunnelInfrastructure(wgIf string, vipIP string, IPv6 bool, tunnelList
 		Chain: postroutingChain,
 		Exprs: []expr.Any{
 			&expr.Meta{Key: expr.MetaKeyOIFNAME, Register: 1},
-			&expr.Cmp{Op: expr.CmpOpNeq, Register: 1, Data: append([]byte(wgIf), 0)},
+			&expr.Cmp{Op: expr.CmpOpNeq, Register: 1, Data: ifname(wgIf)},
 			&expr.Payload{DestRegister: 1, Base: expr.PayloadBaseNetworkHeader, Offset: ipOffset, Len: ipLen},
 			&expr.Lookup{SourceRegister: 1, SetName: masqIPSetName},
 			&expr.Payload{DestRegister: 1, Base: expr.PayloadBaseTransportHeader, Offset: 2, Len: 2},
@@ -762,7 +763,7 @@ func EnsureTunnelInfrastructure(wgIf string, vipIP string, IPv6 bool, tunnelList
 		Chain: manglePreChain,
 		Exprs: []expr.Any{
 			&expr.Meta{Key: expr.MetaKeyIIFNAME, Register: 1},
-			&expr.Cmp{Op: expr.CmpOpEq, Register: 1, Data: append([]byte(wgIf), 0)},
+			&expr.Cmp{Op: expr.CmpOpEq, Register: 1, Data: ifname(wgIf)},
 			&expr.Immediate{Register: 1, Data: binaryutil.NativeEndian.PutUint32(fwmark)},
 			&expr.Ct{Key: expr.CtKeyMARK, Register: 1, SourceRegister: true},
 		},
@@ -774,7 +775,7 @@ func EnsureTunnelInfrastructure(wgIf string, vipIP string, IPv6 bool, tunnelList
 		Chain: manglePreChain,
 		Exprs: []expr.Any{
 			&expr.Meta{Key: expr.MetaKeyIIFNAME, Register: 1},
-			&expr.Cmp{Op: expr.CmpOpNeq, Register: 1, Data: append([]byte(wgIf), 0)},
+			&expr.Cmp{Op: expr.CmpOpNeq, Register: 1, Data: ifname(wgIf)},
 			&expr.Ct{Key: expr.CtKeyMARK, Register: 1},
 			&expr.Cmp{Op: expr.CmpOpEq, Register: 1, Data: binaryutil.NativeEndian.PutUint32(fwmark)},
 			&expr.Meta{Key: expr.MetaKeyMARK, SourceRegister: true, Register: 1},
@@ -967,7 +968,7 @@ func ApplyDNAT(
 		// Multiple targets: use numgen + map for load balancing
 		dnatExprs = []expr.Any{
 			&expr.Meta{Key: expr.MetaKeyIIFNAME, Register: 1},
-			&expr.Cmp{Op: expr.CmpOpEq, Register: 1, Data: append([]byte(wgIf), 0)},
+			&expr.Cmp{Op: expr.CmpOpEq, Register: 1, Data: ifname(wgIf)},
 			&expr.Meta{Key: expr.MetaKeyL4PROTO, Register: 1},
 			&expr.Cmp{Op: expr.CmpOpEq, Register: 1, Data: []byte{protoNum}},
 			&expr.Payload{DestRegister: 1, Base: expr.PayloadBaseTransportHeader, Offset: 2, Len: 2},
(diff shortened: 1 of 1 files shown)


pkg/manager/worker/wireguard.go
```diff
@@ -127,34 +142,201 @@ func (w *WireGuard) OnStartedLeading(ctx context.Context) {
...
 		if err := w.svcProcessor.ServicesWatcher(ctx, w.svcProcessor.SyncServices); err != nil {
 			log.Error("failed to start services watcher", "err", err)
 		}
 	}
-	log.Info("nftables DNAT rule applied successfully")
+	log.Info("nftables DNAT rule applied successfully", "targetCount", len(targets))
+
+	// Start endpoint watcher for kubernetes service to handle API server changes
+	w.endpointWatcherCtx, w.endpointWatcherStop = context.WithCancel(ctx)
+	go w.watchKubernetesEndpoints(w.endpointWatcherCtx, tunnelConfig)
```
p-strusiewiczsurmacki-mobica: I think this goroutine should be synced (preferably using `sync.WaitGroup`) so it won't be shutted down amid DNAT update.

pkg/manager/worker/wireguard.go
```diff
@@ -127,34 +142,201 @@ func (w *WireGuard) OnStartedLeading(ctx context.Context) {
...
+			}
+		}
+	}
+
+	return targets, nil
+}
+
+// watchKubernetesEndpoints watches the kubernetes service EndpointSlices for changes
+// and updates the DNAT rules when API server endpoints change (e.g., when an API server goes down)
+func (w *WireGuard) watchKubernetesEndpoints(ctx context.Context, tunnelConfig *wireguard.TunnelConfig) {
```
p-strusiewiczsurmacki-mobica, line 156: As this is watching for service/endpoints, maybe we could reuse code from the `services` and `endpoints` packages?

pkg/manager/worker/wireguard.go
```diff
@@ -127,34 +142,201 @@ func (w *WireGuard) OnStartedLeading(ctx context.Context) {
...
+			case watch.Error:
+				log.Warn("kubernetes endpoint watch error", "event", event)
+			}
+		}
+	}
+}
+
+// updateControlPlaneDNAT updates the DNAT rules for the control plane with current kubernetes endpoints
+func (w *WireGuard) updateControlPlaneDNAT(ctx context.Context, tunnelConfig *wireguard.TunnelConfig) error {
+	targets, err := w.fetchKubernetesEndpoints(ctx)
```
p-strusiewiczsurmacki-mobica: I think we do not need to fetch for the endpoints - those should be obtainable from `watch.Event.Object`, similar to what's done in https://github.com/kube-vip/kube-vip/blob/main/pkg/endpoints/providers/endpointslices.go#L56

pkg/manager/worker/wireguard.go
```diff
@@ -127,34 +142,201 @@ func (w *WireGuard) OnStartedLeading(ctx context.Context) {
...
+	err = nftables.ApplyDNAT(
+		tunnelConfig.InterfaceName,
+		vipIP,
+		6443,
+		targets,
+		"controlplane",
+		false,
+		v1.ProtocolTCP,
+		false,
+		tunnelConfig.ListenPort,
```
p-strusiewiczsurmacki-mobica: As `IPv6` argument of `nftables.ApplyDNAT` is always false, what will happen for IPv6 enabled clusters? I see no safeguards for IP protocol check in `ApplyDNAT`. Same goes for `UpdateDNAT`.

pkg/manager/worker/wireguard.go
```diff
@@ -127,34 +142,201 @@ func (w *WireGuard) OnStartedLeading(ctx context.Context) {
 	// Strip CIDR notation from VIP if present
 	vipIP := utils.StripCIDR(w.config.VIP)
 
-	// Parse Kubernetes API port
-	kubeAPIPortInt, err := strconv.ParseUint(w.kubeAPIPort, 10, 16)
+	// Fetch the kubernetes API server endpoints from the "kubernetes" service in default namespace
+	targets, err := w.fetchKubernetesEndpoints(ctx)
```
p-strusiewiczsurmacki-mobica: Maybe, instead of fetching endpoints here, we could start watch for endpoints and then start the configuration as soon as the endpoints are available?

pkg/manager/worker/wireguard.go
daniel-naegele: The shutdown process is now synced to the DNAT update process.

pkg/manager/worker/wireguard.go
daniel-naegele: I refactored the code to reuse as much code from service/endpoints as possible (about -100 lines)

pkg/manager/worker/wireguard.go
daniel-naegele: As the provider for endpoints(lices) is reused, this should be implemented already

pkg/manager/worker/wireguard.go
daniel-naegele: The `ApplyDNAT` function now infers the protocol type and makes sure that the VIP and the target IPs are all of the same type.

pkg/manager/worker/wireguard.go
daniel-naegele: Should be fixed with the reuse of the endpoint(slice)s provider.

pkg/manager/worker/wireguard.go
Cellebyte: Are you using the inet table? :)

pkg/nftables/nftables.go
```diff
@@ -822,3 +1165,248 @@ func ipFamily(IPv6 bool) uint32 {
...
+		return fmt.Errorf("failed to get rules: %w", err)
+	}
+
+	hasIfaceRule := false
+	hasCtStateRule := false
+	for _, rule := range rules {
+		for _, e := range rule.Exprs {
+			// Check for interface-based rule
+			if cmp, ok := e.(*expr.Cmp); ok {
+				if string(cmp.Data) == wgIf+"\x00" {
```
Cellebyte: Do we have this null-byte thing on other strings as well?
Maybe we can have a interfaceName func which adds the null-byte :)
https://github.com/telekom/multi-networkpolicy-nftables/blob/b4112317156b46c4babb8d528ded3329dca08a23/pkg/server/netfilterrules.go#L841

pkg/nftables/nftables.go
```diff
@@ -822,3 +1165,248 @@ func ipFamily(IPv6 bool) uint32 {
...
+	if !hasIfaceRule {
+		ifaceRule := &nftables.Rule{
+			Table: mangleTable,
+			Chain: rpfilterChain,
+			Exprs: []expr.Any{
+				&expr.Meta{Key: expr.MetaKeyIIFNAME, Register: 1},
+				&expr.Cmp{
+					Op:       expr.CmpOpEq,
+					Register: 1,
+					Data:     append([]byte(wgIf), 0),
```
Cellebyte: same here :)

Cellebyte: I used in another project something like this -> https://github.com/telekom/multi-networkpolicy-nftables/blob/b4112317156b46c4babb8d528ded3329dca08a23/pkg/server/netfilterrules.go#L841

pkg/nftables/nftables.go
daniel-naegele: Good catch! I added that helper function and utilized it throughout the implementation to ensure predictable behavior.

daniel-naegele:
Does someone know why the service e2e tests are failing? The only difference between this and the last successful test run is this https://github.com/kube-vip/kube-vip/pull/1469/changes/a566c9ddedf4c7276b56e667cc584936eb3c740e

Cellebyte:
@daniel-naegele I don’t know.
I restarted them this afternoon and all of them were green in the other PRs. But it always fails at the same stage. Maybe it is some kind of timing issue.

Cellebyte:
@daniel-naegele could you rebase from latest main?

daniel-naegele:
> @daniel-naegele could you rebase from latest main?

I guess that did the trick

PR #1473, opened 2026-03-17
fix: Assign VIP to interface in BGP mode

MarijnRitzen:
PR https://github.com/kube-vip/kube-vip/pull/1442 re-introduced a guard for assigning the service IP to an interface - this time including a check for BGP being disabled.

There is a use-case where it's important that in BGP mode the IP is actually assigned to an interface: when using https://github.com/angeloxx/cilium-haegress-operator in BGP mode. 

Cilium docs say [here](https://docs.cilium.io/en/stable/network/egress-gateway/egress-gateway/):
> When Cilium is unable to select the Egress IP for an Egress Gateway policy (for example because the specified egressIP is not configured for any network interface on the gateway node), then the gateway node will drop traffic that matches the policy with the reason No Egress IP configured.

We need the ip to be assigned on an interface for the cilium egress gateway to work. 

Seeing as the original problem the contributor in https://github.com/kube-vip/kube-vip/pull/1442  was trying to solve was adding a guard for the routing table, I think this change should not affect their use case.

### DIFF
--- a/pkg/cluster/service.go
+++ b/pkg/cluster/service.go
@@ -312,9 +312,7 @@ func (cluster *Cluster) StartLoadBalancerService(ctx context.Context, c *kubevip
 			} else {
 				log.Info("successful add Route")
 			}
-		}
-
-		if !c.EnableRoutingTable && !c.EnableBGP {
+		} else if !c.EnableRoutingTable {
 			// Normal VIP addition, use skipDAD=false for normal DAD process
 			if _, err = network.AddIP(false, false); err != nil {
 				log.Warn(err.Error())


Cellebyte:
This will not be added.
As this makes the IP available on the node which can have unexpected side effects.

MarijnRitzen:
This was already present before?

MarijnRitzen:
Can you explain what unexpected side effects this would lead to? Can `vip_interface` help here? Setting it to `lo` is enough for the HA BGP egress use case to work, so maybe that can be used to alleviate these side effects? Otherwise I would also be open to adding a flag for this behavior.  Curious to hear what you think.

Cellebyte:
@MarijnRitzen if you have a socket listening on 0.0.0.0 or [::] it exposes unwanted node service. We could add an opt-in flag for Kube-vip or per-service.

MarijnRitzen:
Ah so if I understand you correctly this would be the problem:

1. Kubevip sets a bgp announced address (e.g. 1.2.3.4) on one of the interfaces
2. You have a socket listening on 0.0.0.0 on port 8000
3. The packet (if its directed at 1.2.3.4 port 8000) would not be NAT'ed by the CNI but would instead be sent to this socket

Is that the failure scenario? 

Btw I would be in favor of an opt-in flag like bgp_attach_ip_to_interface or similar

Cellebyte:
Then please create a new PR which allows to enable this behaviour via an opt-in flag on kube-vip as suggested :)

PR #1465, opened 2026-03-12
services: retry on 403 Forbidden in ServicesWatcher WatchFunc

sotoiwa:
## Summary

Fixes #1464

On Kubernetes 1.34, `admin.conf` uses `O = kubeadm:cluster-admins` instead of `O = system:masters`. On joining control plane nodes, kube-vip starts before etcd is promoted from learner to full member. The RBAC data is not yet available through the learner etcd, causing a transient 403 Forbidden or 401 Unauthorized on the first Watch call.

Previously `ServicesWatcher()` exited permanently on any error from the `WatchFunc`, breaking LoadBalancer VIP assignment and HA on joining CP nodes.

## Changes

- Replace the bare Watch call in `ServicesWatcher` with `wait.ExponentialBackoffWithContext` (from `k8s.io/apimachinery/pkg/util/wait`, already in `go.mod`) for idiomatic retry with exponential backoff (2s initial, 2x factor, 30s cap, 10 steps).
- Handle both 403 Forbidden and 401 Unauthorized, since `retrywatcher.go` treats both as permanent stops and the same etcd learner condition can produce either.
- Non-auth errors are returned immediately. Context cancellation stops the retry loop cleanly.
- Change `Processor.rwClientSet` from `*kubernetes.Clientset` to `kubernetes.Interface` to allow fake client injection in tests. A type assertion is added in `watchEndpoint()` for `CreateRetryWatcher` which still requires `*kubernetes.Clientset`.
- Add unit tests covering: 403 retry succeeds, non-403 error fails fast, context cancellation during retry.
- Add CHANGELOG entry.

## Testing

The root cause (transient 403 at startup due to etcd learner state) has been verified on an actual 3-node CP-only bare metal cluster (EKS Anywhere v0.24.4, Kubernetes v1.34.3) using kubeadm via Cluster API.

Unit tests for the retry logic have been added to `pkg/services/watch_services_test.go`. ~~End-to-end testing on actual hardware with K8s 1.34 joining CP nodes has **not yet been performed** — we would appreciate help from the maintainers or community to validate this fix in a real environment.~~

**Updated (2026-03-13):** End-to-end testing on actual hardware (EKS-A v0.24.4, K8s 1.34.3, Tinkerbell, 3-node CP-only bare metal) has been performed and the fix works. See comment below for full results and logs.

PR #1477, opened 2026-03-19
Fixed leader election restart issue

### LINKED ISSUE #1474: services aren't watched based upon externalTrafficPolicy
**Describe the bug**
A service with `externalTrafficPolicy=local` is never reconciled once the original pod is rec-cylced.. this is attributed to a change is v1.0.4.

**To Reproduce**

```
piVersion: v1
kind: Service
metadata:
  name: service3
  annotations:
    kube-vip.io/egress: "true"
    kube-vip.io/egress-internal: "true"
spec:
  type: LoadBalancer
  # "Local" preserves the client source IP and avoids a second hop for
  # LoadBalancer and NodePort
  externalTrafficPolicy: Local
  selector:
    env: demo3
  ports:
    - protocol: TCP
      port: 9000
      targetPort: 9000
```
```
apiVersion: v1
kind: Pod
metadata:
  name: pod-03
  labels:
    env: demo3
spec:
  containers:
    - name: pod-03
      image: thebsdbox/demo:v1
      env:
        - name: CLIENT
          value: "service4"
```

Delete the pod and recreate a few times and the watcher has stopped

**Expected behavior**
kube-vip continues watching for endpoints to a service..
**Screenshots**

 - OS/Distro: [e.g. Ubuntu 1804]
 - Kubernetes Version: [e.g. v.1.18]
 - Kube-vip Version: [e.g. 0.2.3]

**`Kube-vip.yaml`:**
If Possible add in your kube-vip manifest (please remove anything that is confidential)

**Additional context**
The `wg` and the `contexts` are too complicated to follow

p-strusiewiczsurmacki-mobica:
This should fix #1474 

The issue was, that context for lease and actual leader election was common. When there was no endpoints, service context was cancelled, which lead to lease deletion and leader election cancel and non of those were restarted.

### DIFF
--- a/pkg/endpoints/endpoints.go
+++ b/pkg/endpoints/endpoints.go
@@ -75,7 +75,14 @@ func (p *Processor) AddOrModify(svcCtx *servicecontext.Context, event watch.Even
 		// start leader election if it's enabled and not already started
 		if !svcCtx.IsActive && p.config.EnableServicesElection {
 			wg.Go(func() {
-				startLeaderElection(svcCtx, service, serviceFunc, wg)
+				for {
+					select {
+					case <-svcCtx.Ctx.Done():
+						return
+					default:
+						startLeaderElection(svcCtx, service, serviceFunc, wg)
+					}
+				}
 			})
 		}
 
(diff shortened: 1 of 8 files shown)


Cellebyte:
Could that also fix #1465, maybe?

Cellebyte:
LGTM grpc is fixed in another PR :)

p-strusiewiczsurmacki-mobica:
> Could that also fix #1465, maybe?

I don't think so. The ETCD issue seems to be different.

PR #1199, opened 2025-07-08
Add VIPs as deprecated

jplitza:
This causes IPv6 source address selection according to RFC 3484 to ignore VIPs when choosing which address the host should use for egress communication.

Fixes #1198

PR #1489, opened 2026-03-25
Fix preferred liftime settings for pre-existing addresses

### LINKED ISSUE #1475: DNS stops wotking when VIP is same as host IP in single node cluster.
**Describe the bug**
DNS stops working when VIP is same as host IP in single node cluster.
Due to the fix https://github.com/kube-vip/kube-vip/pull/1199, VIP is marked as deprecated and host stops using that IP for DNS resolve.

**To Reproduce**
Setup 1 node kubernetes cluster with VIP being same as host IP.  Host should have only one active external interface that is being used to communicate to DNS server.

**Expected behavior**
DNS should keep working when VIP is same as host IP.

**Screenshots**

 - OS/Distro: Ubuntu 2204
 - Kubernetes Version: NA
 - Kube-vip Version: 1.0.3

p-strusiewiczsurmacki-mobica:
This should fix #1475

After #1199 all addresses are added with `prefered_lft=0` - that results in address tagged as `deprecated` which might affect some things (e.g. DNS resolution, as in the describe issue). As this is not an issue for some VIPs, in the specific case of reusing the node IP as VIP this might not be optimal. With this PR, if kube-vip tries to reuse an existing address, `prefered_lft` of that address will be preserved instead of setting it as `0`.

### DIFF
--- a/pkg/vip/address.go
+++ b/pkg/vip/address.go
@@ -366,10 +366,13 @@ func (configurator *network) AddIP(precheck bool, skipDAD bool, minLifetime ...i
 	defer configurator.link.Lock.Unlock()
 	var existing *netlink.Addr
 	var err error
-	if precheck {
-		if existing, err = configurator.IsSet(); err != nil {
-			return false, errors.Wrap(err, "could not check if address exists")
-		}
+
+	if existing, err = configurator.IsSet(); err != nil {
+		return false, errors.Wrap(err, "could not check if address exists")
+	}
+
+	if existing != nil {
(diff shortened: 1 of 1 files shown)


Cellebyte:
@p-strusiewiczsurmacki-mobica let's put this to draft until the person in the raised issue responds on the question.
As this only solves IPv6 I think.

Cellebyte:
Closing this as feedback is still missing.
Sorry @p-strusiewiczsurmacki-mobica, but vip is a Virtual IP so it should collide with the nodes own IP ^^.

Issue #1145, opened 2025-04-29
KubeVIP won't stop advertising control plane address when api server is down


pchlupacek:
**Describe the bug**
When local kube api server fails, KubeVIP is not removing VIP address from BGP advertisements. 

**To Reproduce**
1. Start fresh kubernetes HA cluster
2. Setup kube-vip with BGP 
3. Remove kube-api manifest (to make api servet to fail)
4. Api server failure is not detected, address is still advertised 

**Expected behavior**
When api server is not reacheable locally, kube vip will stop advertise its VIP

**Environment (please complete the following information):**
 - OS/Distro: Flatcar linux 
 - Kubernetes Version: 1.32
 - Kube-vip Version: v0.8.10

**`Kube-vip.yaml`:**

apiVersion: v1
kind: Pod
metadata:
  name: kube-vip
  namespace: kube-system
spec:
  containers:
    - args:
        - manager
      env:
        - name: vip_arp
          value: "false"
        - name: kubernetes_addr
          value: "127.0.0.1:6443"
        - name: port
          value: "6443"
        - name: vip_nodename
          valueFrom:
            fieldRef:
              fieldPath: spec.nodeName
        - name: vip_interface
          value: lo
        - name: vip_cidr
          value: "32"
        - name: dns_mode
          value: first
        - name: cp_enable
          value: "true"
        - name: cp_namespace
          value: kube-system
        - name: bgp_enable
          value: "true"
        - name: bgp_routerid
          value: "{{ kube_vip.bgp_routerid }}"
        - name: bgp_sourceip
          value: "{{ kube_vip.bgp_sourceip }}"
        - name: bgp_as
          value: "{{ kube_vip.bgp_as }}"
        - name: bgp_peeraddress
        - name: bgp_peerpass
        - name: bgp_peeras
          value: "65000"
        - name: bgp_peers
          value: "{{ kube_vip.bgp_peers }}"
        - name: address
          value: "{{ kube_vip.address }}"
        - name: prometheus_server
          value: :2112
      image: "ghcr.io/kube-vip/kube-vip:{{ kube_vip.version }}"
      imagePullPolicy: IfNotPresent
      name: kube-vip
      resources: {}
      securityContext:
        capabilities:
          add:
            - NET_ADMIN
            - NET_RAW
          drop:
            - ALL
      volumeMounts:
        - mountPath: /etc/kubernetes/admin.conf
          name: kubeconfig
  hostAliases:
    - hostnames:
[...]

Issue #1236, opened 2025-08-15
DeathSwitch for kube-vip when running without leader-election


Cellebyte:
**Is your feature request related to a problem? Please describe.**

We have the awesome leader-election mechanism for single-path modes or to elect one leader for multi-path modes.
This allows us to drawback all announcements when the API Server becomes unreachable.

A similar feature should be implemented when kube-vip does not elect a leader but rather runs in single-instance or multiple single instances. So when `vip_leaderelection: false` and `svc_election: false`.

**Describe the solution you'd like**

Instead of electing the leader it should use the lease API as a hearbeat to the kube-api.
We could either reuse the existing `node` lease or instantiate our own per kube-vip instance.
This feature should be opt-out.

**Additional context**

This allows a cluster to cut all its ties to the world in case it is not possible to schedule or do anything else on the api server.

todaygood:
I can't argee any more,  I recently experienced one accident occured in the production envionment:  the vip jump from one to another node  when the kube-apiserver is down, the kube-vip pod  on the master node is crashed,  so 
kubevip shoudn't depend on kube-apiserver.

Cellebyte:
@todaygood maybe I add a Out Of Scope section here.
As this was mainly for the `svc_enable` flag and not for the `cp_enable`.

p-strusiewiczsurmacki-mobica:
@Cellebyte I'll be implementing this if that's OK

Cellebyte:
Ignore the cp mode here entirely. We should not make the entrypoint to the apiserver dependent on the apiserver or etcd.

PR #1498, opened 2026-04-02
Add configurable control-plane health check for BGP mode

g-gaston:
In BGP mode without leader election, every kube-vip instance announces the same VIP and tipically an upstream router distributes traffic via ECMP. If a node's kube-apiserver goes down while kubelet keeps running, kube-vip continues advertising the route — sending a fraction of API traffic into a black hole.

This PR adds an optional health check loop that polls a configurable HTTP(S) endpoint (e.g. `https://localhost:6443/livez`). When the endpoint fails consecutively, the BGP route is withdrawn. When it recovers, the route is re-announced. On SIGTERM, the route is gracefully withdrawn before exit.

The feature is disabled by default. Set `control_plane_health_check_address` to enable it.

For more info, I wrote a draft for the docs: https://github.com/kube-vip/website/pull/77

I think this addresses https://github.com/kube-vip/kube-vip/issues/1145

## Design decisions

**HTTP health check instead of Kubernetes client**

A plain HTTP GET is used instead of a Kubernetes client:

- **Lightweight**: no REST config parsing, clientset construction, or retry logic on each tick.
- **Decoupled from kubeconfig and kube-apiserver**: kube-vip doesn't require a working kubeconfig.
- **Flexible**: the operator can point it at any HTTP endpoint (`/livez`, `/healthz`, a sidecar, a custom aggregator), not just the kube-apiserver.

**Opt-in only (no default endpoint)**

The health check is entirely opt-in via `control_plane_health_check_address`. When unset, behavior is unchanged: routes are announced immediately on startup. When set, routes are only announced after the first successful probe. This is a behavioral change that should be explicit. It also avoids assuming the service behind the VIP is always a kube-apiserver on port 6443.

**Consecutive failure threshold before route withdrawal**

A single failed probe does not withdraw the route. The route is only withdrawn after `control_plane_health_check_failure_threshold` consecutive failures (default 3). This avoids route flapping on transient errors (e.g. a brief API server restart or a slow response under load).

**BGP-only for now, generic config names**

This initial implementation targets BGP control-plane mode only, to keep the scope manageable. However, the configuration fields (`ControlPlaneHealthCheck.*`) and the health check logic are not BGP-specific. The same mechanism could be extended to other modes in the future without renaming or breaking the config surface.

## Configuration

| Environment variable | Default | Description |
|---|---|---|
| `control_plane_health_check_address` | `""` (disabled) | URL to poll |
| `control_plane_health_check_period_seconds` | `5` | Interval between checks |
| `control_plane_health_check_timeout_seconds` | `3` | Per-request timeout |
| `control_plane_health_check_failure_threshold` | `3` | Consecutive failures before withdrawal |
| `control_plane_health_check_ca_path` | `""` | CA cert for HTTPS verification |

Equivalent `--controlPlaneHealthCheck*` CLI flags are also available.

## Testing

- [x] Unit tests for HTTP client initialization, TLS CA loading, and config validation
- [x] White-box tests for the health check loop (announce/withdraw/recovery/shutdown)
- [x] E2e test on a 2-node Kind cluster that:
  - Verifies routes are announced for all healthy nodes
  - Stops the apiserver on one node and verifies route withdrawal
  - Restores the apiserver and verifies re-announcement
  - Stops kube-vip itself and verifies graceful route withdrawal
  - Restores kube-vip and verifies re-announcement


Cellebyte:
Hey I like the idea of this feature. It would be awesome if your code would solve it for all modes. As this is not only an issue with the BGP mode. 

You correctly mentioned #1145.
Additionally there is the idea to use leases for this kind of issue, maybe take a look at #1236.

p-strusiewiczsurmacki-mobica:
We already have some health-checking code in `pkg/loadbalancer/ipvs.go` - maybe that could be abstracted somehow and reused?

g-gaston:
> Hey I like the idea of this feature. It would be awesome if your code would solve it for all modes. As this is not only an issue with the BGP mode.

Yeah supporting other modes would make sense. That's why I didn't name the config env vars/flags after BGP, so they can be re-used for other modes. However, I decided to start with BGP only to reduce the scope for the first iteration.

What other configuration combinations would you like to see support for and what behavior would you expect from the healthcheck? I'll take a look to those parts of the codebase and see what the lift would be.

That said, would you/others be opposed to start with BGP and extend to other modes in follow-up PRs? This one is already quite big.

> You correctly mentioned #1145. Additionally there is the idea to use leases for this kind of issue, maybe take a look at #1236.

On the topic of leases: that's an interesting idea, what benefit do you see to use leases over a health endpoint?

I went this route because I believe a health endpoint is more flexible. Most people will use the kube-api sever one (in fact there are 2 to choose from) but you could have your own that groups the kube-api server one with something else or even be independent from the api server. Also, I find it simpler to reason about and it's way lighter on the api server.

g-gaston:
> We already have some health-checking code in `pkg/loadbalancer/ipvs.go` - maybe that could be abstracted somehow and reused?

Thanks for the pointer, good catch! After reading the code in `pkg/backend` and `pkg/loadbalancer/ipvs.go`, a few things that might make reuse tricky:

- Probe mechanism: `backend.Entry.Check()` builds a full Kubernetes clientset on every call and hits `ServerVersion()`. For this health-check I wanted a plain HTTP GET to a configurable endpoint (e.g. `/livez`). Using a HTTP call lets us avoid the overhead of creating a k8s client (and requiring a kubeconfig), and it gives operators flexibility to point at any health endpoint — not just the k8s API discovery path.
- Single target vs. multi-backend: IPVS iterates a map of backends and checks each one individually. The BGP case only monitors one local endpoint to decide whether to announce/withdraw a single route.
- Failure handling: IPVS reacts immediately on a single failed check (remove backend). The BGP health check uses a consecutive failure threshold before withdrawing the route, which avoids flapping on transient blips.

The one generic piece is `backend.Watch()` (the ticker loop), I could reuse that. That said I'm not sure that saves more than a few lines.

I think a shared "health checker" abstraction *could* work for both. We would take a `Checker` interface (HTTP probe, k8s probe, whatever else), configurable failure thresholds, and `onHealthy`/`onUnhealthy` callbacks. And we would need to support N tuples of checker+callbacks to serve the ipvs use-case.

I just feel like we'd end up writing more code to build that abstraction than we're saving by reusing logic.

Is that the kind of thing you had in mind, or were you thinking of something else?

Cellebyte:
@g-gaston I think for the controlplane mode it makes sense to go with your suggested solutions as the apiserver is the endpoint you are interested in. So I think progressing here with livez would be a good idea.

I would just like to implement the solution in such a way it could be used with any mode.

g-gaston:
> @g-gaston I think for the controlplane mode it makes sense to go with your suggested solutions as the apiserver is the endpoint you are interested in. So I think progressing here with livez would be a good idea.
> 
> I would just like to implement the solution in such a way it could be used with any mode.

@Cellebyte Got you. Happy to explore this!

What do you have in mind? Do you think the other modes could also use a health endpoint? That would make sense to me, since it's probably the lowest common denominator.

Maybe we can have a reusable package that takes the config struct and runs the healthcheck in aloop, taking a couple hooks (one for healthy and one for unhealthy)?

Just trying to understand your vision so I focus on the right things 🙂

thebsdbox:
This is a bit of a funny one for the control plane behaviour, only BGP is/was designed to work without requiring leaderElection as ARP would result in duplicate nodes with the IP address. Also the control plane behaviour is that these kube-vip instances run on the control plane nodes and connect to localhost. If the API server is down for whatever reason then the instance of kube-vip will `panic()` and a leader Election would take place (no point attracting the traffic to a node where the API server is dead).  ¯\_(ツ)_/¯

g-gaston:
> This is a bit of a funny one for the control plane behaviour, only BGP is/was designed to work without requiring leaderElection as ARP would result in duplicate nodes with the IP address. Also the control plane behaviour is that these kube-vip instances run on the control plane nodes and connect to localhost. If the API server is down for whatever reason then the instance of kube-vip will `panic()` and a leader Election would take place (no point attracting the traffic to a node where the API server is dead). ¯_(ツ)_/¯

@thebsdbox to be honest, I'm only familiar with ARP and BGP for control plane mode. I knew this feature wasn't necessary for ARP. My current issue is with BGP without leader election: my kube-vip kept announcing the route even if the API server was down. So that's why I started with BGP only. But I don't know anything about other modes and if/how this could be re-used. 

I am little bit confused. Are you suggesting this feature is not necessary in control plane mode even when not using leader election? Or that this feature only makes sense for BGP mode and doesn't need to be generic?

Cellebyte:
@g-gaston I think @thebsdbox wanted to say we never really anticipated the use-case for multipath on the control-plane. I know that the ipvs code already checks if the local backend is available. So as a workaround you could enable ipvs which would redirect traffic to a healthy node, but traffic would still arrive at the broken node.

Cellebyte:
But I also think supporting multipath in the control-plane mode should be doable.

g-gaston:
@Cellebyte Got you, thanks for the clarification :)

So would you and @thebsdbox be open to support multipath for control plane? And if so, is the approach in this PR appropriate or do you think I should explore other ideas?

thebsdbox:
@g-gaston I'd be happy with that 😄

g-gaston:
> @g-gaston I'd be happy with that 😄

@thebsdbox awesome. Rebased and moving PR out of draft mode, ptal

I think one outstanding question based in the conversation in this PR is:

Do we want to keep the "generic" naming in the new config fields/flags for control plane? Or given that BGP is the only mode that is designed to work without leader election, should the config naming be BGP specific?

`control_plane_health_check_address` -> `bgp_control_plane_health_check_address`

I opted for generic in case we ever wanted to sue this for other control plane modes, but I don't have a strong opinion.

thebsdbox:
We will keep it as generic for now as I think there is scope for other control planes in the future 😄

PR #1511, opened 2026-04-09
Fixed restart on node watcher error

p-strusiewiczsurmacki-mobica:
This should fix #1510 I think

On Node watcher error kube-vip does not either restart the watch itself nor restarts the kube-vip's pod. This PR levereges `watchWithAuthRetry` function added in #1465 to add watch restart on `403 forbidden` error. Addtionally, if watch will not be properly restarted after 30s, kube-vip pod will be killed and recreated.

### DIFF
--- a/pkg/loadbalancer/ipvs.go
+++ b/pkg/loadbalancer/ipvs.go
@@ -51,15 +51,13 @@ type IPVSLoadBalancer struct {
 	backendMap          backend.Map
 	interval            int
 	lock                sync.Mutex
-	stop                chan struct{}
-	networkInterface    string
 	killFunc            func()
 	address             string
 	family              ipvs.AddressFamily
 }
 
 func NewIPVSLB(ctx context.Context, address string, port uint16, forwardingMethod string, backendHealthCheckInterval int,
-	networkInterface string, killFunc func(), wg *sync.WaitGroup) (*IPVSLoadBalancer, error) {
+	killFunc func(), wg *sync.WaitGroup) (*IPVSLoadBalancer, error) {
 	log.Info("Starting IPVS LoadBalancer", "address", address)
 
 	// Create IPVS client
@@ -140,7 +138,6 @@ func NewIPVSLB(ctx context.Context, address string, port uint16, forwardingMetho
 		forwardingMethod:    m,
 		interval:            backendHealthCheckInterval,
 		backendMap:          make(backend.Map),
-		networkInterface:    networkInterface,
 		killFunc:            killFunc,
 		address:             address,
 		family:              family,
@@ -167,7 +164,6 @@ func enableProcSys(path, name string) error {
 
 func (lb *IPVSLoadBalancer) RemoveIPVSLB() error {
 	log.Info("Stopping IPVS LoadBalancer", "address", lb.address)
-	close(lb.stop)
 	err := lb.client.RemoveService(lb.loadBalancerService)
 	if err != nil {
 		return fmt.Errorf("error removing existing IPVS service: %v", err)
@@ -190,6 +186,7 @@ func (lb *IPVSLoadBalancer) AddBackend(address string, port uint16) error {
 		if err != nil {
 			log.Error("checking if backend is local", "err", err)
 		}
+		log.Info("checked if backend is local", "addr", address, "local", isLocal)
 	}
 
 	backend := backend.Entry{Addr: address, Port: port, IsLocal: isLocal}
@@ -346,6 +343,7 @@ func (lb *IPVSLoadBalancer) healthCheck(ctx context.Context) {
 				}
 				if lb.forwardingMethod == ipvs.Local && !lb.localBackendExists() {
 					if lb.killFunc != nil {
+						log.Error("no local backends available, restarting kube-vip")
 						lb.killFunc()
 					}
 				}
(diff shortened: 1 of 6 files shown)


pkg/loadbalancer/ipvs.go
```diff
@@ -355,29 +353,32 @@ func (lb *IPVSLoadBalancer) healthCheck(ctx context.Context) {
...
 func (lb *IPVSLoadBalancer) isLocal(address string) (bool, error) {
-	link, err := netlink.LinkByName(lb.networkInterface)
-	if err != nil {
-		return false, fmt.Errorf("getting link '%s': %w", lb.networkInterface, err)
-	}
-
-	family := netlink.FAMILY_V6
-	if utils.IsIPv4(address) {
-		family = netlink.FAMILY_V4
-	}
```
p-strusiewiczsurmacki-mobica, line 366: @Cellebyte this is the change I mentioned in the #1510 

Previously we only searched for the presence of the backend IP on the interface kube-vip is configured to use. However, in case someone uses for example `lo` this might not be so great, so I think checking all interfaces should be fine.

pkg/loadbalancer/ipvs.go
```diff
@@ -355,29 +353,32 @@ func (lb *IPVSLoadBalancer) healthCheck(ctx context.Context) {
...
-
 	target := net.ParseIP(address)
 	if target == nil {
-		return false, fmt.Errorf("address '%s' is not a valid IP address", address)
+		return false, fmt.Errorf("unable to parse IP address %s", address)
 	}
 
-	addrs, err := netlink.AddrList(link, family)
+	links, err := netlink.LinkList()
```
p-strusiewiczsurmacki-mobica, line 361: @Cellebyte Here, instead of using the interface, list all the interfaces and check all of them for the address presence.

Cellebyte: Is it possible to filter here?
This can get very long when you have a lot of interfaces + addresses.

Cellebyte: Maybe we can search for the address using netlink?

p-strusiewiczsurmacki-mobica: > Maybe we can search for the address using netlink?

I don't think that's possible.

I think what we can do is to either

1. Add env variable like `secondary_interfaces: "eth1,eth3,eth5"` where user could provide list of interfaces that should be scanned (maybe that could be used later for something else as well).

or

2. Eventually we could filter some interfaces, for example removing those that have a master interface defined to remove veths from the list so we wouldn't iterate over their addresses, and remove the addresses with scope other than `global` to not perform the `addr.IP.Equal` check for the link-local addresses.

Cellebyte: I think the second option is easier to use. So let’s go with a constant set of ignored cases and document them :)

Cellebyte: @p-strusiewiczsurmacki-mobica is this code-path only used when ipvs is enabled?

p-strusiewiczsurmacki-mobica: Yes

p-strusiewiczsurmacki-mobica: @Cellebyte I have added a check to filter `veth` interfaces and non-global IP addresses. Also added info about this in https://github.com/kube-vip/website/pull/81

Cellebyte: What is with IPv4 addresses do they also have this scope?

pkg/loadbalancer/ipvs.go
```diff
@@ -355,29 +353,39 @@ func (lb *IPVSLoadBalancer) healthCheck(ctx context.Context) {
...
+	family := netlink.FAMILY_V4
+	if utils.IsIPv6(address) {
+		family = netlink.FAMILY_V6
 	}
 
-	for _, addr := range addrs {
-		if addr.IP.Equal(target) {
-			return true, nil
+	for _, link := range links {
+		if link.Type() == "veth" {
```
Cellebyte, line 372: Is there a constant for this type?

pkg/loadbalancer/ipvs.go
```diff
@@ -355,29 +353,39 @@ func (lb *IPVSLoadBalancer) healthCheck(ctx context.Context) {
...
+		}
+
+		addrs, err := netlink.AddrList(link, family)
+		if err != nil {
+			log.Error("listing addresses", "link", link.Attrs().Name, "error", err.Error())
+			continue
+		}
+
+		for _, addr := range addrs {
+			if addr.Scope != int(netlink.SCOPE_UNIVERSE) {
```
Cellebyte, line 383: If scope is a bit mask maybe a bit compare is more efficient.

pkg/loadbalancer/ipvs.go
p-strusiewiczsurmacki-mobica: There's not. Even in `netlink`, string is being used for that:
https://github.com/vishvananda/netlink/blob/main/link_linux.go#L2197

pkg/loadbalancer/ipvs.go
p-strusiewiczsurmacki-mobica: I've tried to create a small benchmark to simulate this and compare simple `==` with some bit operators, but I see no difference here really and `==` is much more readable I think.

Cellebyte: Ok :)

PR #1208, opened 2025-07-21
Draft: Support BGP unnumbered peers

excepshenal:
Add support for BGP unnumbered peers

https://github.com/kube-vip/kube-vip/issues/1207

Still needs integration test

Issue #1207, opened 2025-07-21
Support for BGP unnumbered


excepshenal:
**Is your feature request related to a problem? Please describe.**
Support BGP unnumbered for kube-vip.

**Describe the solution you'd like**
GoBGP supports BGP unnumbered, so this entails adding support for parsing BGP unnumbered peer(s) from the config. From there, it's minimal work to pass the values in `AddPeer` to GoBGP.

I am thinking of only adding support in the `bgp_peers` env var, to reduce parsing logic sprawl across the other BGP env vars that assume one peer, which I assume are really kept for backward compatibility.

**Describe alternatives you've considered**
Can also add support in the other BGP env vars, but this will complicate the logic slightly. However, this makes it a bit more flexible in that users can use either approach to define an (unnumbered) peer.

I assume kube-vip is moving toward a CRD approach in the future which will improve the growing parsing logic sprawl.

**Additional context**
I'm probably going to work on a PoC.

thebsdbox:
Thanks and looking forward to seeing your experiments. Kube-vip is in a bit of an odd position where due to the control plane side of things it exists before CRDs (hence ENV vars). For the services side of things then yes we aim to move to CRDs 😀 

If you need any help or pointers please let me know.

mattcarp12:
@Cellebyte this can be closed now.

Cellebyte:
Sorry to reopen it, but the tests seem to fail.
I reverted the feature again and would like you to investigate the issue @mattcarp12

mattcarp12:
@Cellebyte Close?

Cellebyte:
Yes :) thanks I thought it was already linked ^^

PR #1512, opened 2026-04-12
feat: add support for BGP unnumbered peers

mattcarp12:
### Title: 
feat: Add BGP Unnumbered peering support (Supersedes #1208)

### Related Issues
* **Fixes:** #1207 
* **Supersedes:** #1208 

Note: This PR revives and continues the fantastic foundational work started by @excepshenal in #1208, bringing it up to date with the current `main` branch.

### Validation & Testing
* Added unit tests to explicitly verify that `ParseBGPPeerConfig` correctly parses the `unnumbered:` prefix while maintaining strict backwards compatibility for standard IPv4 and IPv6 BGP configurations.

### DIFF
--- a/pkg/bgp/peers.go
+++ b/pkg/bgp/peers.go
@@ -13,6 +13,7 @@ import (
 	api "github.com/osrg/gobgp/v3/api"
 
 	"github.com/kube-vip/kube-vip/pkg/utils"
+	"github.com/osrg/gobgp/v3/pkg/config/oc"
 	"github.com/osrg/gobgp/v3/pkg/server"
 	"google.golang.org/protobuf/types/known/anypb"
 )
@@ -21,9 +22,10 @@ import (
 func (b *Server) AddPeer(ctx context.Context, peer kubevip.BGPPeer) (err error) {
 	p := &api.Peer{
 		Conf: &api.PeerConf{
-			NeighborAddress: peer.Address,
-			PeerAsn:         peer.AS,
-			AuthPassword:    peer.Password,
+			NeighborAddress:   peer.Address,
+			PeerAsn:           peer.AS,
+			NeighborInterface: peer.Interface,
+			AuthPassword:      peer.Password,
 		},
 
 		Timers: &api.Timers{
@@ -45,6 +47,17 @@ func (b *Server) AddPeer(ctx context.Context, peer kubevip.BGPPeer) (err error) 
 			RemoteAddress: peer.Address,
 			RemotePort:    uint32(179),
 		},
+	}
+
+	if peer.Interface != "" {
+		neighborAddress, err := oc.GetIPv6LinkLocalNeighborAddress(peer.Interface)
+		if err != nil {
+			return fmt.Errorf("failed to get link-local address of interface %s: %w", peer.Interface, err)
+		}
+
+		p.State = &api.PeerState{
+			NeighborAddress: neighborAddress,
+		}
 	}
 
 	if b.c.MpbgpNexthop != "" {
--- a/pkg/kubevip/config_bgp.go
+++ b/pkg/kubevip/config_bgp.go
@@ -16,6 +16,7 @@ import (
 type BGPPeer struct {
 	Address      string
 	Port         uint16
+	Interface    string
 	AS           uint32
 	Password     string
 	MultiHop     bool
@@ -65,6 +66,7 @@ func ParseBGPPeerConfig(config string) (bgpPeers []BGPPeer, err error) {
 			continue
 		}
 		isV6Peer := peerStr[0] == '['
+		isUnnumberedPeer := strings.HasPrefix(peerStr, "unnumbered:")
 
 		address := ""
 		if isV6Peer {
@@ -74,20 +76,29 @@ func ParseBGPPeerConfig(config string) (bgpPeers []BGPPeer, err error) {
 			}
 			address = peerStr[1:addressEndPos]
 			peerStr = peerStr[addressEndPos+1:]
+		} else if isUnnumberedPeer {
+			unnumberedEndPos := strings.IndexByte(peerStr, ':')
+			peerStr = peerStr[unnumberedEndPos+1:]
 		}
 
 		peer := strings.Split(peerStr, ":")
-		if len(peer) < 2 {
+		if len(peer) < 2 && !isUnnumberedPeer {
 			return nil, fmt.Errorf("mandatory peering params <host>:<AS> incomplete")
 		}
 
-		if !isV6Peer {
+		iface := ""
+		if isUnnumberedPeer {
+			iface = peer[0]
+		} else if !isV6Peer {
 			address = peer[0]
 		}
 
-		ASNumber, err := strconv.ParseUint(peer[1], 10, 32)
-		if err != nil {
-			return nil, fmt.Errorf("BGP Peer AS format error [%s]", peer[1])
+		var ASNumber uint64
+		if len(peer) >= 2 {
+			ASNumber, err = strconv.ParseUint(peer[1], 10, 32)
+			if err != nil {
+				return nil, fmt.Errorf("BGP Peer AS format error [%s]", peer[1])
+			}
 		}
 
 		password := ""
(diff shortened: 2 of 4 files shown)


mattcarp12:
@Cellebyte - I see all the e2e tests passed. Can you tell me how to reproduce the error from here - https://github.com/kube-vip/kube-vip/pull/1507#issuecomment-4207591079

Cellebyte:
@mattcarp12 I am unsure, why it detected two `fe80::/64` link local addresses for the unnumbered peer. But it happens from time to time. Maybe I restart CI?

mattcarp12:
@Cellebyte 

The tests are passing for me locally. Yes, maybe try running the tests again in CI.

Otherwise, I think `ERROR start manager err="failed to configure BGP mode: starting BGP server: failed to get link-local address of interface eth0: found 2 link-local neighbors. only support p2p link"` is the expected behavior.

thebsdbox (approved):
Nice, this looks great!

Cellebyte:
@thebsdbox we had a weird issue in CI with go BGP code finding two link local addresses.
That’s why it was reverted. 
I think we can merge it again an then take a look, why CI fails?

thebsdbox:
I'll force CI a few times and see if I can get it to trip up.

thebsdbox:
@Cellebyte / @mattcarp12 we got the same error again, what should we be doing next?

Cellebyte:
I think the best thing would implement some logging which prints the interface link local addresses of the interface used for the unnumbered session.

Cellebyte:
@thebsdbox I asked Copilot about it and it seems to have a point here:

# Copilot Response

Looking at the failing test and the code reference you've provided, I can see the issue:Now I can see the issue. Let me look at how unnumbered BGP peers are handled:Now I see the issue! Let me check how the unnumbered peer is being used:Now I can see the issue clearly. Looking at the code in `pkg/bgp/peers.go` at lines 52-61, when an unnumbered BGP peer is configured (with `peer.Interface != ""`), the code calls `oc.GetIPv6LinkLocalNeighborAddress()` from GoBGP to discover the link-local IPv6 address of the peer.

The problem you're describing is that `GetIPv6LinkLocalNeighborAddress()` is discovering **2 link-local IPv6 addresses** instead of the expected 1, causing it to fail with the error: "found 2 link-local neighbors. only support p2p link".

## Root Cause

The issue is in the GoBGP function logic. When querying neighbors on an interface, it's likely picking up:
1. The actual peer's link-local address
2. A duplicate or additional neighbor entry (possibly from gratuitous ARP/NDP or timing issues during test cluster setup/teardown)

In a test environment where clusters are rapidly created and destroyed, network state can be inconsistent, leading to stale neighbor table entries.

## Solution

Add a filter to the neighbor discovery to select the most recent/active neighbor entry:

```go
// In pkg/bgp/peers.go, modify the unnumbered peer handling (lines 52-61)

if peer.Interface != "" {
	neighborAddress, err := oc.GetIPv6LinkLocalNeighborAddress(peer.Interface)
	if err != nil {
		// Log and retry or implement fallback
		return fmt.Errorf("failed to get link-local address of interface %s: %w", peer.Interface, err)
	}

	p.State = &api.PeerState{
		NeighborAddress: neighborAddress,
	}
}
```

However, the real fix should be in the GoBGP library itself. The `GetIPv6LinkLocalNeighborAddress` function should be modified to filter neighbors by:
1. Reachability state (exclude `NUD_FAILED`)
2. Recency (prefer most recently updated entries)
3. Return the single best match instead of failing on count > 1

**Better approach - create a wrapper function in kube-vip:**

```go
// Add this to pkg/bgp/peers.go

func getUnnumberedPeerAddress(ifname string) (string, error) {
	ifi, err := net.InterfaceByName(ifname)
	if err != nil {
		return "", fmt.Errorf("interface not found: %w", err)
	}

	neighs, err := netlink.NeighList(ifi.Index, netlink.FAMILY_V6)
	if err != nil {
		return "", fmt.Errorf("failed to list neighbors: %w", err)
	}

	var validNeighbors []netlink.Neigh
	for _, neigh := range neighs {
		// Only consider reachable neighbors
		if neigh.State&netlink.NUD_REACHABLE != 0 && neigh.IP.IsLinkLocalUnicast() {
			validNeighbors = append(validNeighbors, neigh)
		}
	}

	if len(validNeighbors) == 0 {
		return "", fmt.Errorf("no reachable link-local IPv6 neighbors found on %s", ifname)
	}

	// Use the first reachable neighbor
	addr := validNeighbors[0].IP
	return fmt.Sprintf("%s%%%s", addr, ifname), nil
}
```

Then replace the call in `AddPeer`:

```go
if peer.Interface != "" {
	neighborAddress, err := getUnnumberedPeerAddress(peer.Interface)
	if err != nil {
		return fmt.Errorf("failed to get link-local address of interface %s: %w", peer.Interface, err)
	}

	p.State = &api.PeerState{
		NeighborAddress: neighborAddress,
	}
}
```

This filters to only `NUD_REACHABLE` neighbors, which eliminates stale entries and handles the test environment race condition.

thebsdbox:
OK, well I'm actually pretty impressed with AI for working that out 😄

Cellebyte:
@thebsdbox Keep the suggestion still with a grain of salt. I don’t checked it yet.

p-strusiewiczsurmacki-mobica:
@thebsdbox @Cellebyte I've spent some time to debug this and I think this is more of an issue with the tests than the PR itself.

I was able to reproduce both `found 2 link-local neighbors` and `no ipv6 link-local neighbor found` errors locally.

- **no neighbor case** - it seems that kube-vip starts before neighbor is populated.
- **multiple neighbors** - I am able to reproduce this sometimes when using `E2E_PRESERVE_CLUSTER=true`. In such case when two kind clusters are running using single network bridge, one neighbor is populated by the IP address from the docker bridge on the host and the other one is taken from the other kind container.

Now, the AI solution might work, but I am not sure if it is worth to duplicate parts of code from `gobgp` to solve an issue with those tests + selecting just a random address from the list seems to be a bit odd to me. I think that this test could be deleted for now and implemented later if we will have an idea on how to fix this properly (I have none yet).

mattcarp12:
@Cellebyte @thebsdbox I've been looking into this more. 

If you look at the two failing tests runs - [here](https://github.com/kube-vip/kube-vip/actions/runs/24096611687/job/70296721320) and [here](https://github.com/kube-vip/kube-vip/actions/runs/24318111025/job/71456304205) - you'll see that the "2 link-local IPv6 addresses" error is not actually happening. 

Instead, both of these test runs failed because `checkIPAddress` failed [here](https://github.com/mattcarp12/kube-vip/blob/f4e57e599110b14ebe6521cf9e9e0a46cd87af1d/testing/e2e/e2e_ds_test.go#L531) after timing out after 180 seconds.

Do you think this could be timing out due to DAD? What do you think?

Cellebyte:
@p-strusiewiczsurmacki-mobica I also don’t think we should duplicate code but rather contribute a fix for goBGP or fix our tests :D

p-strusiewiczsurmacki-mobica:
@mattcarp12 The first link you've provided indeed fails on IP check for control-plane, but it seems to use IPv4 kind cluster, so I am not sure if that's relevant. It is also seems to be based on a bit older branch, as it does not incorporate the logs for deamonset based tests.

The second one fails with `2 link-local IPv6 addresses` according to the logs, so maybe you've pasted a wrong link?

mattcarp12:
@p-strusiewiczsurmacki-mobica You are correct. Sorry I'm new to this project and still learning how these tests work 😅. I'm still working on debugging this on my branch.

p-strusiewiczsurmacki-mobica:
@mattcarp12 @Cellebyte I have found that `kind` can use env variable `KIND_EXPERIMENTAL_DOCKER_NETWORK` to [create a cluster that uses specified docker network](https://github.com/kubernetes-sigs/kind/blob/main/pkg/cluster/internal/providers/docker/provider.go#L72) instead of default `kind` network. I think we could use this to create separate network for each daemonset-based e2e test cluster and therefore get rid of this bug. The creation has to be guarded by a mutex, so the env will not be overridden by concurrent tests.

I have created a commit that incorporates those changes - you can take a look at the last commit here - https://github.com/p-strusiewiczsurmacki-mobica/kube-vip/pull/63

Now, if this solution will be accepted then either @mattcarp12 can pull the commit into theirs PR, or I can create a separate PR for that.

Cellebyte:
@p-strusiewiczsurmacki-mobica If you have a solution I am fine to merge a separate MR after this one :)

p-strusiewiczsurmacki-mobica:
@Cellebyte I have create a PR. The unnumbered BGP tests are commented there, so after it will be merged @mattcarp12 can rebase this branch and uncomment the tests.

Cellebyte:
@mattcarp12 could you rebase your PR on top of current main?

mattcarp12:
@Cellebyte I have rebased onto main. Thank you for your help with this.

Cellebyte:
@mattcarp12 CI is already running :)

Cellebyte (approved):
LGTM

Thanks to @p-strusiewiczsurmacki-mobica  and @mattcarp12  :)

PR #1516, opened 2026-04-14
fix: make `get-gobgp` a dependency of `e2e-tests-bgp` in Makefile

mattcarp12:
My `make e2e-tests-bgp` was failing locally because of the missing `gobgpd`. So I decided to make this a dependency in the Makefile.


thebsdbox (approved):
Thanks, simple fix. This looks good to me.

PR #1518, opened 2026-04-15
Track BGP paths

p-strusiewiczsurmacki-mobica:
This PR introduces a tacker for BGP paths similar to what `ARP manager` does in ARP mode - this should make it much easier to manage BGP paths in case of service/endpoint creation/deletion, as decision whether path should be deleted or created is contained within the BGP server object.

My next step will be to add similar thing for routes in RT mode to get rid of the code that searches for routes common for multiple services, and to get similar workflow for each mode.

### DIFF
--- a/pkg/cluster/service.go
+++ b/pkg/cluster/service.go
@@ -41,6 +41,13 @@ func (cluster *Cluster) StartVipService(ctx context.Context, c *kubevip.Config, 
 		killFunc()
 	})
 
+	nodename := ""
+	if c.NodeName != "" {
+		nodename = c.NodeName
+	} else {
+		nodename = os.Getenv("HOSTNAME")
+	}
+
 	loadbalancers := []*loadbalancer.IPVSLoadBalancer{}
 
 	for i := range cluster.Network {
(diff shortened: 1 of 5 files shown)


pkg/cluster/service.go
```diff
@@ -75,7 +75,7 @@ func (cluster *Cluster) StartVipService(ctx context.Context, c *kubevip.Config,
 		if c.EnableBGP {
 			// Lets advertise the VIP over BGP, the host needs to be passed using CIDR notation
 			log.Debug("Attempting to advertise over BGP", "address", network.CIDR())
-			err = bgpServer.AddHost(ctx, network.CIDR())
+			err = bgpServer.AddHost(ctx, network.CIDR(), "kube-vip-control-plane")
```
Cellebyte: I hope this never conflicts with a service and this should be a constant :)

p-strusiewiczsurmacki-mobica: Should not, as services use namespaced name. But I have changed that to use nodename at first and then fallback to the `const` if nodename was not found.

pkg/cluster/service.go
```diff
@@ -41,6 +41,13 @@ func (cluster *Cluster) StartVipService(ctx context.Context, c *kubevip.Config,
 		killFunc()
 	})
 
+	nodename := ""
+	if c.NodeName != "" {
+		nodename = c.NodeName
+	} else {
+		nodename = os.Getenv("HOSTNAME")
```
Cellebyte: should we not use the nodename function here?

p-strusiewiczsurmacki-mobica: I have changed that part. I think that, as the `c.NodeName` is already processed when a new `Manager` is being created, there is no need at all to process it here again.

Cellebyte (approved):
LGTM

PR #1520, opened 2026-04-16
Stop service status update on context cancel

p-strusiewiczsurmacki-mobica:
Minor bugfix

With this change the `statusUpdate` function will exit if context was cancelled as there's no point on to retrying service status update with cancelled context.


Cellebyte (approved):
LGTM

PR #1527, opened 2026-04-23
Create separate Docker network for each kind cluster in daemonset-bas…

p-strusiewiczsurmacki-mobica:
…ed e2e tests

This should fix issues with the tests in PR #1512 .

This uses env variable `KIND_EXPERIMENTAL_DOCKER_NETWORK` so different docker network can be used for each daemonset-based e2e test cluster instead of the default `kind` network.  The creation of the cluster is guarded by a mutex, so the env will not be overridden by concurrent tests.


thebsdbox:
Nice! I love anything that will make the tests more stable 😄

PR #1529, opened 2026-04-25
docs: document SELinux IPVS module setup

### LINKED ISSUE #506: Documentation: Kernel modules (SELinux)
When running kube-vip on nodes with SELinux enabled it fails to load the kernel modules if not SELinux boolean `domain_kernel_load_modules` is set to true. I seen this being configured in other projects. Dan Walsh is arguing against this [here](https://danwalsh.livejournal.com/77415.html) and I agree with him - that we don't want to let any container load whatever kernel module it wants.

It might be a good idea to add a section to the documentation about making sure to load the modules `ip_vs` and `ip_vs_rr` before deploying kube-vip. I'm not sure if this should be a "Troubleshooting SELinux" section or a general recommendation/requirement.

---
Below are some symptoms for anyone finding this issue trying to get kube-vip running and having no idea what is going on (same as me from the beginning). If this is true, try loading the kernel modules `ip_vs` and `ip_vs_rr`.
- kube-vip pod in Error/CrashLoopBackOff
- kube-vip pod logs show
  ```
  time="2023-01-25T07:22:03Z" level=info msg="Starting IPVS LoadBalancer"
  time="2023-01-25T07:22:03Z" level=error msg="ensure IPVS kernel modules are loaded"
  time="2023-01-25T07:22:03Z" level=fatal msg="Error starting IPVS [netlink receive: no such file or directory]"
  ```
- Linux audit log show
  ```
  type=AVC msg=audit(1674633883.119:874): avc:  denied  { module_request } for  pid=15328 comm="kube-vip" kmod="net-pf-16-proto-16-family-IPVS" scontext=system_u:system_r:container_t:s0:c139,c162 tcontext=system_u:system_r:kernel_t:s0 tclass=system permissive=0
  ```

Thanks!

AkashKumar7902:
Closes #506.

## Summary
- Add a README troubleshooting note for SELinux-enforcing nodes that block kube-vip from requesting IPVS kernel modules.
- Document the common symptoms and the required `ip_vs` and `ip_vs_rr` modules.
- Recommend preloading only those modules instead of enabling broad container module loading.

## Validation
- `go test ./pkg/loadbalancer -run Test_ipAndFamily -count=1`
- `git diff --check`

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,33 @@ All of these would require a separate level of configuration and in some infrast
 
 ## Troubleshooting and Feedback
 
+### SELinux and IPVS kernel modules
+
+When using IPVS load balancing on nodes with SELinux enforcing, kube-vip may be
+blocked from requesting kernel modules from inside the container. Symptoms can
+include the kube-vip pod entering `Error` or `CrashLoopBackOff`, logs that show
+`ensure IPVS kernel modules are loaded`, or audit denials for `module_request`
+from `container_t`.
+
+Load the required IPVS modules on every node that can run kube-vip before
+deploying it:
+
+```shell
+sudo modprobe ip_vs
+sudo modprobe ip_vs_rr
+```
+
+To persist this across reboots, add the modules to a file such as
+`/etc/modules-load.d/kube-vip-ipvs.conf`:
+
+```text
+ip_vs
+ip_vs_rr
+```
+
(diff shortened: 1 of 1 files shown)


Cellebyte (approved):
LGTM, maybe a PR for the website would be good as well

PR #1534, opened 2026-04-28
Improve logging in election.go and set defaultBGPPort constant

mattcarp12:
**Changes:**

- Logging: Refactored election.go to use structured logging and error wrapping (%w) for better trace debugging.
- BGP: Introduced a defaultBGPPort constant (179) to replace magic numbers in peer configuration.
- Cleanup: Simplified conditional logic in interfaces.go

### DIFF
--- a/pkg/election/election.go
+++ b/pkg/election/election.go
@@ -198,7 +198,7 @@ func (em *Manager) NodeWatcher(ctx context.Context, lb *loadbalancer.IPVSLoadBal
 					if checkIfNodeIsReady(node) {
 						err = lb.AddBackend(node.Status.Addresses[x].Address, port)
 						if err != nil {
-							log.Error("failed to add node to load balancer", "node", node.Name, "ip", node.Status.Addresses[x].Address, "err", err)
+							log.Error("adding node to load balancer", "node", node.Name, "ip", node.Status.Addresses[x].Address, "err", err)
 							if errors.Is(err, &utils.PanicError{}) {
 								return fmt.Errorf("add IPVS backend: %w", err)
 							}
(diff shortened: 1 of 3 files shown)


pkg/election/election.go
```diff
@@ -198,15 +198,15 @@ func (em *Manager) NodeWatcher(ctx context.Context, lb *loadbalancer.IPVSLoadBal
 					if checkIfNodeIsReady(node) {
 						err = lb.AddBackend(node.Status.Addresses[x].Address, port)
 						if err != nil {
-							log.Error("add IPVS backend", "err", err)
+							log.Error("failed to add node to load balancer", "node", node.Name, "ip", node.Status.Addresses[x].Address, "err", err)
```
Cellebyte: failed to add should be replaced with adding, as it is already an error log.

pkg/election/election.go
```diff
@@ -198,15 +198,15 @@ func (em *Manager) NodeWatcher(ctx context.Context, lb *loadbalancer.IPVSLoadBal
...
 							if errors.Is(err, &utils.PanicError{}) {
 								return fmt.Errorf("add IPVS backend: %w", err)
 							}
 						}
 					} else {
 						err = lb.RemoveBackend(node.Status.Addresses[x].Address, port)
 						if err != nil {
-							log.Error("remove IPVS backend", "err", err)
+							log.Error("failed to remove node from load balancer", "node", node.Name, "ip", node.Status.Addresses[x].Address, "err", err)
```
Cellebyte: Same here

pkg/election/election.go
```diff
@@ -222,7 +222,7 @@ func (em *Manager) NodeWatcher(ctx context.Context, lb *loadbalancer.IPVSLoadBal
 				if node.Status.Addresses[x].Type == v1.NodeInternalIP {
 					err = lb.RemoveBackend(node.Status.Addresses[x].Address, port)
 					if err != nil {
-						log.Error("Del IPVS backend", "err", err)
+						log.Error("failed to remove node from load balancer", "node", node.Name, "ip", node.Status.Addresses[x].Address, "err", err)
```
Cellebyte: And here

Cellebyte (changes requested):
Don’t use failed in an error log as it is already error :)

pkg/election/election.go
mattcarp12: Done.

pkg/election/election.go
mattcarp12: Done.

pkg/election/election.go
mattcarp12: Done.

PR #1535, opened 2026-04-28
ARP Rate Validation and Ticker

mattcarp12:
Updates:
- `StartAdvertisement` checks `ArpBroadcastRate < 500` after every broadcast. But once it resets the value to 3000, the condition can never be true again — so it's only useful on the first iteration. This PR validates the rate once in `NewManager`.
- The time.Sleep in the loop means a ctx cancellation during the sleep is undetected for up to 3 seconds.  This PR replaces the sleep-loop with a ticker that respects context cancellation.

### DIFF
--- a/pkg/arp/arp.go
+++ b/pkg/arp/arp.go
@@ -25,6 +25,10 @@ type Instance struct {
 }
 
 func NewManager(config *kubevip.Config) *Manager {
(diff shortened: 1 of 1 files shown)


p-strusiewiczsurmacki-mobica:
LGTM

PR #1536, opened 2026-04-29
Added route tracker

p-strusiewiczsurmacki-mobica:
This is a continuation of #1518 

It adds similar tracking object as it was implemented for BGP and ARP. This way routes can be deleted without tracking separately if they are shared (all the logic is contained within the `route.Manager`).


Cellebyte (approved):
LGTM :)

PR #1542, opened 2026-05-06
Configuration of the deprecated endpoints in the tests fixed

p-strusiewiczsurmacki-mobica:
PR #1215 set `EndpointSlices` as a default instead of the deprecated `Endpoints`. However this change was not reflected in the tests, and therefore all the tests, even those configured explicitly to use `Endpoints`, used `EndpointSlices` despite the settings. As the support for the deprecated `Endpoints` was not yet removed, I think this should be fixed.


Cellebyte (approved):
LGTM I think we will remove it when Kubernetes will also remove it ^^

PR #1543, opened 2026-05-07
Changed logs gathering path in the CI

p-strusiewiczsurmacki-mobica:
While trying to figure out another CI error I have discovered that not all logs are gathered as not all paths follow the same convention with `kube-vip-test` directory name prefix. This should be fixed with this PR.


Cellebyte (approved):
LGTM

PR #1553, opened 2026-05-15
Fixed ARP mode with cluster policy in 1.2.0-rc.0

p-strusiewiczsurmacki-mobica:
It turned out that after #1463 the search for all endpoints was missing in ARP mode so the VIP was not added for services  with `externalClusterPolicy: Cluster` (previously only local endpoints were used, and VIP was always added to the host if Cluster policy was deployed).

This PR should fix this issue.

There is also an issue with the deletion of the VIP in such case, but I'll fix this in a separate PR.


Cellebyte:
How does this deal with leader election? 
So it would advertise only from the elected node, or?

Cellebyte (approved):
LGTM

PR #1554, opened 2026-05-16
feat(svc): add vlan support for services

### LINKED ISSUE #1517: Add Vlan support to kube-vip
When defining a service be able to specify that it should also be part on an interface that is tagged with a vlan header.

An annotation in the service that specified both the interface and the vlan tag, kube-vip will then create a sub-interface with the correct vlan tag and apply the service address to this sub-interface.

N/A

**Additional context**
[A simple example of doing this natively with the Go netlink library](https://github.com/vishvananda/netlink/blob/c822ed716ea170073a6fa4239082c2512be1fe99/link_test.go#L918)

thebhdn:
Closes https://github.com/kube-vip/kube-vip/issues/1517

Added annotation `kube-vip.io/serviceVLAN` to a service that specifies both the parent interface and the vlan tag. Then kube-vip will create subinterface with the vlan tag and apply the service address to this subinterface.

### DIFF
--- a/pkg/instance/instance.go
+++ b/pkg/instance/instance.go
@@ -44,6 +44,10 @@ type Instance struct {
 	DHCPv4Client        vip.DHCPClient
 	DHCPv6Client        vip.DHCPClient
 
+	// Service use Vlan
+	IsVLAN        bool
+	VLANInterface string
+
 	// External Gateway IP the service is forwarded from
 	UPNPGatewayIPs []string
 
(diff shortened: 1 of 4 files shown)


pkg/instance/instance.go
```diff
@@ -432,6 +465,62 @@ func getAutoInterfaceName(link netlink.Link, defaultInterface string) string {
 	return link.Attrs().Name
 }
 
+func (i *Instance) addVLAN(parentInterface string, tag int) error {
+	var parent netlink.Link
+
+	interfaceName := fmt.Sprintf("%s.%d", parentInterface, tag)
+	_, err := net.InterfaceByName(interfaceName)
```
p-strusiewiczsurmacki-mobica: NIT: I'd use `netlink.LinkByName` here and reuse this value in the `else` instead of looking for it twice.

p-strusiewiczsurmacki-mobica (approved):
LGTM

thebsdbox (approved):
Can you also update the docs in the website repo. But this looks great.

thebhdn:
yep, will do! And thanks btw 😎

thebsdbox:
> yep, will do! And thanks btw 😎 

The thanks is to you, it's your hard work adding new features! 

So thanks 🙏🏻

PR #1561, opened 2026-05-19
Fixed ARP and RT modes with global leader election for services in 1.2.0-rc.0

p-strusiewiczsurmacki-mobica:
This is a followup to #1553 

This PR fixes the VIP deletion and re-addition error I observed in ARP mode when using global leader election.

E2E tests for such case will be added in a separate PR.

EDIT: I have found similar issue in RT mode (it's the same code path) - this is fixed in this PR as well.


Cellebyte (approved):
LGTM

PR #1562, opened 2026-05-21
Added debouncer for watch events

p-strusiewiczsurmacki-mobica:
This PR is a proposal for an event debouncer for Endpoints and services events.

At the moment kube-vip is processing events as those arrive which means in example, that if there is a service that kube-vip is reconciling endpoints for, if, let's say 20 endpoints will be created, and each one will came up within 1 second of another, kube-vip might try to run reconciliation code up to 20 times.

This PR introduces a debouncer that will make kube-vip reconcile only the last request. This should simplify the logs generated by kube-vip and as a result could make maintenance a little bit more approachable.

The debouncer waits for the event for the service/endpoints/endpointslices object to arrive, then waits some `time` from the event. If new event arrives, it overrides the previous event, debouncer waits for the same `time` for another event and so on. If the `time` passes with no new events, the event is sent to the output channel and reconciliation is carried as usual.

The `time` debouncer waits for the event can be configured with a CLI flag `debounceTime`, or with an ENV variable `debounce_time`. Those can be set with a `string` value that can be parsed with [time.ParseDuration()](https://pkg.go.dev/time#ParseDuration) (e.g. "5s", "500ms", etc.). The minimal value is "200ms", and the default one is "2s", however those values were selected arbitrarily and this can be discussed.

Question: If this will be accepted, should it be possible to disable the debouncer, for example with `debounce_time` set to `0s`?

### DIFF
--- a/cmd/kube-vip.go
+++ b/cmd/kube-vip.go
@@ -19,6 +19,7 @@ import (
 	"github.com/vishvananda/netlink"
 	"golang.org/x/sys/unix"
 
+	"github.com/kube-vip/kube-vip/pkg/debouncer"
 	"github.com/kube-vip/kube-vip/pkg/kubevip"
 	"github.com/kube-vip/kube-vip/pkg/manager"
 	"github.com/kube-vip/kube-vip/pkg/metrics"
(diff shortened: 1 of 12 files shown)


thebsdbox:
I'd accept this. But the question for me is the other way around, should this be opt-in as it changes the default behaviour.

p-strusiewiczsurmacki-mobica:
> I'd accept this. But the question for me is the other way around, should this be opt-in as it changes the default behaviour.

Sure, I am open to this. :)

thebsdbox:
This looks good to me, would it be possible to have this behind a flag then to move this forward? Also `/pkg/debouncer/*` has a mis-spelt file.

p-strusiewiczsurmacki-mobica:
@thebsdbox This was rebased and updated. Debouncer will now be only enabled if `debounce_time` will be explicitly set to value >= 200ms and the default value is 0s, which means it's disabled by default.

thebsdbox:
Awesome, thanks for adding this 😄

PR #1566, opened 2026-05-24
fix: refactor labeler interface, enable labeling with cp_enabled=true

### LINKED ISSUE #1131: KubeVIP node labeling does not work
**Describe the bug**
When adding the environment variable `enable_node_labeling` to the KubeVIP Pod, the leader node is never labeled.

**To Reproduce**
1. Deploy KubeVIP Pods on "_node1_", "_node2_" and "_node3_" with env variable `enable_node_labeling` set to `true`
2. Find the leader node (eg: "_node2_")
3. The label **kube-vip.io/has-ip=\<VIP address>** does not exists on "_node2_"

**Expected behavior**
I shoud see the label **kube-vip.io/has-ip=\<VIP address>** on "_node2_"

**Environment:**
 - OS/Distro: **AlmaLinux 8.8**
 - Kubernetes Version (k3s): **v1.30.6+k3s1**
 - Kube-vip Version: **v0.8.2**

**`Kube-vip.yaml`:**
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: kube-vip
  namespace: kube-system
spec:
  containers:
  - args:
    - manager
    env:
    - name: svc_election
      value: "false"
    - name: enable_node_labeling
      value: "true"
    - name: vip_arp
      value: "true"
    - name: port
      value: "6443"
    - name: vip_interface
      value: "ens192"
    - name: vip_cidr
      value: "32"
    - name: cp_enable
      value: "true"
    - name: cp_namespace
      value: kube-system
    - name: vip_ddns
      value: "false"
    - name: svc_enable
      value: "false"
    - name: vip_leasename
      value: "vip-ops"
    - name: vip_leaderelection
      value: "true"
    - name: vip_leaseduration
      value: "5"
    - name: vip_loglevel
      value: "5"
    - name: vip_renewdeadline
      value: "3"
    - name: vip_retryperiod
      value: "1"
    - name: address
      value: "192.168.10.10"
    - name: my_ip
      valueFrom:
        fieldRef:
          fieldPath: status.podIP
    - name: prometheus_server
      value: "$(my_ip):2113"
    image: ghcr.io/kube-vip/kube-vip:v0.8.2
    imagePullPolicy: IfNotPresent
    name: kube-vip
    resources:
      requests:
        cpu: 30m
        memory: 32Mi
      limits:
        cpu: "1"
        memory: 1Gi
    securityContext:
      capabilities:
        add:
        - NET_ADMIN
        - NET_RAW
        - SYS_TIME
    volumeMounts:
      name: kubeconfig
  hostAliases:
  - hostnames:
    - kubernetes
    ip: 127.0.0.1
  hostNetwork: true
  volumes:
  - hostPath:
      path: /etc/rancher/k3s/k3s.yaml

```

**Additional context**
Here are the logs of one of my KubeVIP pod: 
```
time="2025-04-18T08:17:55Z" level=info msg="Starting kube-vip.io [v0.8.2]"
time="2025-04-18T08:17:55Z" level=debug msg="Build kube-vip.io [04ce471366c21d4586fb2d683cd166f0dc4e18ce]"
time="2025-04-18T08:17:55Z" level=info msg="namespace [kube-system], Mode: [ARP], Features(s): Control Plane:[true], Services:[false]"
time="2025-04-18T08:17:55Z" level=info msg="prometheus HTTP server started"
time="2025-04-18T08:17:55Z" level=warning msg="Node name is missing from the config, fall back to hostname"
time="2025-04-18T08:17:55Z" level=info msg="Using node name [dlc2]"
time="2025-04-18T08:17:55Z" level=debug msg="[k8s client] we try the incluster first, this error [open /var/run/secrets/kubernetes.io/serviceaccount/token: no such file or directory] can saf
time="2025-04-18T08:17:55Z" level=debug msg="Using external Kubernetes configuration from file [/etc/kubernetes/admin.conf]"
time="2025-04-18T08:17:55Z" level=info msg="Starting Kube-vip Manager with the ARP engine"
time="2025-04-18T08:17:55Z" level=debug msg="init enable service security: false"
time="2025-04-18T08:17:55Z" level=info msg="Beginning cluster membership, namespace [kube-system], lock name [plndr-cp-lock], id [dlc2]"
I0418 08:17:55.531269       1 leaderelection.go:250] attempting to acquire leader lease kube-system/plndr-cp-lock...
time="2025-04-18T08:17:55Z" level=info msg="Node [dlc1] is assuming leadership of the cluster"
```

As you can see, i am using ARP Mode.

I read the source code of KubeVIP. I found that the `applyNodeLabel(...)` function is called only one time and it is in the `startARP(...)` method of the Manager. 
However, it looks like the `applyNodeLabel()` call is never reached as i can't see any of the logs above the call of `applyNodeLabel()`. 

This is the part of the code i'm talking about: 
```go
	// Start a services watcher (all kube-vip pods will watch services), upon a new service
	// a lock based upon that service is created that they will all leaderElection on
	if sm.config.EnableServicesElection {
		log.Infof("beginning watching services, leaderelection will happen for every service")
		err = sm.startServicesWatchForLeaderElection(ctx)
		if err != nil {
			return err
		}
	} else {

		log.Infof("beginning services leadership, namespace [%s], lock name [%s], id [%s]", ns, sm.config.ServicesLeaseName, id)
		// we use the Lease lock type since edits to Leases are less common
		// and fewer objects in the cluster watch "all Leases".
		lock := &resourcelock.LeaseLock{
			LeaseMeta: metav1.ObjectMeta{
				Name:      sm.config.ServicesLeaseName,
				Namespace: ns,
			},
			Client: sm.clientSet.CoordinationV1(),
			LockConfig: resourcelock.ResourceLockConfig{
				Identity: id,
			},
		}

		// start the leader election code loop
		leaderelection.RunOrDie(ctx, leaderelection.LeaderElectionConfig{
			Lock: lock,
			// IMPORTANT: you MUST ensure that any code you have that
			// is protected by the lease must terminate **before**
			// you call cancel. Otherwise, you could have a background
			// loop still running and another process could
			// get elected before your background loop finished, violating
			// the stated goal of the lease.
			ReleaseOnCancel: true,
			LeaseDuration:   time.Duration(sm.config.LeaseDuration) * time.Second,
			RenewDeadline:   time.Duration(sm.config.RenewDeadline) * time.Second,
			RetryPeriod:     time.Duration(sm.config.RetryPeriod) * time.Second,
			Callbacks: leaderelection.LeaderCallbacks{
				OnStartedLeading: func(ctx context.Context) {
					err = sm.servicesWatcher(ctx, sm.syncServices)
					if err != nil {
						log.Fatal(err)
					}
				},
				OnStoppedLeading: func() {
					// we can do cleanup here
					log.Infof("leader lost: %s", id)
					for _, instance := range sm.serviceInstances {
						for _, cluster := range instance.clusters {
							cluster.Stop()
						}
					}

					log.Fatal("lost leadership, restarting kube-vip")
				},
				OnNewLeader: func(identity string) {
					// we're notified when new leader elected
					if sm.config.EnableNodeLabeling {
						applyNodeLabel(sm.clientSet, sm.config.Address, id, identity)
					}
					if identity == id {
						// I just got the lock
						return
					}
					log.Infof("new leader elected: %s", identity)
				},
			},
		})
	}
```  

These are the lines 114 to 182 of the `pkg/manager/manager_arp.go`.

I am wondering what am i missing. Why does the labeling of the node is not even triggered ? 

Thank you

thebhdn:
Potentially closes https://github.com/kube-vip/kube-vip/issues/1131

Hi folks,

This PR is a proposal to refactor node labeling and enable it for control plane mode.

Currently node labeling only works for services, and only for add action because during election fight it directly calls `deleteService`,  whereas logic for removal exists in `SyncServices`. So the only way labels can be purged - cleanup during restart/exit of kubevip.

The main idea is to make node label interface more generic and avoid tying it to a service object. This should make the implementation easier to reuse in places where only the required labels need to be applied.


Cellebyte (approved):
LGTM

PR #1573, opened 2026-05-26
Tidy - Extract duplicate code to remove VIP

mattcarp12:
There was somewhat duplicated code in `pkg/cluster/clusterLeaderElection.go - OnStoppedLeading()` and `pkg/cluster/service.go - StartLoadBalancerService()`. 

This PR extracts the duplicate code into a separate function that is called by both.


mattcarp12:
@Cellebyte can you run the tests again? I think there was an issue with github. All the tests and checks are running fine for me locally.

thebsdbox:
This looks like an compiler issue, as there is a pkg being imported that isn't being used.

mattcarp12:
@thebsdbox I fixed the issue.

thebsdbox:
Awesome, Nothing better than removing code and improving it at the same time 😄

Issue #1078, opened 2025-02-17
Support for shared DHCP addresses


erickuiper:
**Is your feature request related to a problem? Please describe.**
I am currently working on a large edge deployment and we are on track to successfully implement Kube-VIP as part of this. Everything works as expected, but we have a specific challenge. Our cluster is connected to a unmanaged network that has DHCP. We need to expose multiple services on this network with the same IP as we have systems that need to connect to exposed services on a single IP (this address is advertised using mDNS).

**Describe the solution you'd like**
Kube-vip to support sharing an IP address that was requested using DHCP.

Possible direction, in addition of the 0.0.0.0 address. Add 0.0.0.0/32 as an option which then instructs kube-vip to re-use the same DHCP IP address for all services. 

**Describe alternatives you've considered**
- Tested the setup with 0.0.0.0 address, this works perfectly but it leases a new address for each service.
- The setup works using a static /32 address, but this is an undesired setup with static IPs in a DHCP network.

thebsdbox:
Hmm, this is an interesting one as there would be a ton of required logic to ensure that when a lease expires that all of the various services sharing the DHCP address are updated

Cellebyte:
@thebsdbox @erickuiper maybe #1235 is a start to that.

Cellebyte:
@p-strusiewiczsurmacki-mobica do you think this is now possible with the shared lease annotation?

p-strusiewiczsurmacki-mobica:
@Cellebyte I don't think so as kube-vip is creating separate macvlan interface for each service when DHCP is being used But I think we could easily add an annotation that would override the name of the macvlan interface so all services would be pointed to the same interface, and then all the services could be tied to a single host with the common lease annotation. I'll try to make a POC of that to test if I am right.

p-strusiewiczsurmacki-mobica:
@Cellebyte I have added proposed PR #1574. As far as I've tested this in ARP mode it should work OK and should allow to share DHCP IP address among multiple services.

PR #1574, opened 2026-05-27
Added macvlan annotation for DHCP

p-strusiewiczsurmacki-mobica:
This PR introduces changes that could potentially satisfy request in #1078 

It introduces new service annotation `kube-vip.io/macvlanName` which can be used to force multiple services to use the same macvlan interface. In such case, those services will use the same DHCP IP address.

Example configuration:

```
kube-vip.io/loadbalancerIPs: "0.0.0.0"  # this enables DHCP for the service
kube-vip.io/macvlanName: "test-macvlan" # this ties service to specified macvlan interface
```

When using `extarnalClusterPolicy: Cluster` this can be used alongside `kube-vip.io/leaseName` annotation to tie multiple services with DHCP to single node with single IP address.

The annotation `kube-vip.io/hwaddr` can be used to assign MAC address to the macvlan interface, and therefore preserve the DHCP IP addresses between e.g. kube-vip's restarts.

### DIFF
--- a/pkg/instance/instance.go
+++ b/pkg/instance/instance.go
@@ -44,6 +44,7 @@ type Instance struct {
 	DHCPHostname        string
 	DHCPv4Client        vip.DHCPClient
 	DHCPv6Client        vip.DHCPClient
+	macvlanName         string
 
 	// Service use Vlan
 	IsVLAN        bool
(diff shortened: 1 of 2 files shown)


thebsdbox (approved):
Looks good

thebsdbox:
I'll merge this, is there a corresponding docs update?

p-strusiewiczsurmacki-mobica:
@thebsdbox Not yet, but I'll add that.

PR #1576, opened 2026-05-29
Add dhcp-broadcast annotation

### LINKED ISSUE #1577: LoadBalancer not working with kube-vip DHCP and Cilium KPR
**Describe the bug**
DHCP address assignment fails for `LoadBalancer` services when a DHCP is configured to unicast replies to the VIP address. The reply packets are intercepted by Cilium kube-proxy replacement before the macvlan (`vip-*`) interface can receive them, so the DHCP handshake never completes.

**To Reproduce**

1. Use a network setup where the DHCP replies with unicast by default
2. Deploy kube-vip with service mode enabled
3. Deploy Cilium with kube-proxy replacement enabled
4. Configure a `LoadBalancer` service with DHCP (`kube-vip.io/loadbalancerIPs: "0.0.0.0"`)
5. Observe that the macvlan interface (`vip-*`) has no address and traffic to the VIP is not delivered

**Expected behavior**
The DHCP handshake completes and the macvlan interface (`vip-*`) receives an IP address from the DHCP server.

 - OS/Distro: SLES 15.6
 - Kubernetes Version: 1.34.x (RKE2)
 - Kube-vip Version: v0.9.2

**`Kube-vip.yaml`:**
```yaml
...
  containers:
  - args:
    - manager
    env:
    - name: svc_enable
      value: "true"
    - name: svc_election
      value: "false"
    - name: vip_arp
      value: "true"
    - name: cp_enable
      value: "false"
    - name: lb_enable
      value: "true"
    - name: enable_service_security
      value: "true"
    image: registry.rancher.com/rancher/mirrored-kube-vip-kube-vip-iptables:v0.9.2
    securityContext:
      capabilities:
        add: [NET_ADMIN, NET_RAW]
        drop: [ALL]
  hostNetwork: true
```

DavideRutigliano:
This PR introduces a new service annotation `kube-vip.io/dhcp-broadcast` to address DHCP failures in environments where the relay sends unicasts replies directly to the VIP. In those setups the packet can be intercepted by the Cilium kube-proxy replacement before the macvlan interface receives it.

Setting the BROADCAST flag in DISCOVER/REQUEST packets instructs the DHCP to broadcast the reply instead, bypassing the interception.

Example configuration:

```
kube-vip.io/loadbalancerIPs: "0.0.0.0"    # enables DHCP for the service
kube-vip.io/dhcp-broadcast: "true"        # sets BROADCAST flag in DHCP packets
```

Resolves #1577

### DIFF
--- a/pkg/instance/instance.go
+++ b/pkg/instance/instance.go
@@ -45,6 +45,7 @@ type Instance struct {
 	DHCPv4Client        vip.DHCPClient
 	DHCPv6Client        vip.DHCPClient
 	macvlanName         string
+	dhcpBroadcast       bool
 
 	// Service use Vlan
 	IsVLAN        bool
@@ -303,6 +304,7 @@ func NewInstance(ctx context.Context, svc *v1.Service, config *kubevip.Config,
 		}
 		instance.DHCPHostname = svc.Annotations[kubevip.LoadbalancerHostname]
 		instance.macvlanName = svc.Annotations[kubevip.MacvlanName]
+		instance.dhcpBroadcast = svc.Annotations[kubevip.DHCPBroadcast] == "true"
 	}
 
 	configPorts := make([]kubevip.Port, 0)
@@ -614,7 +616,7 @@ func (i *Instance) startDHCP(ctx context.Context, index int, backoffAttempts uin
 			initRebootFlag = true
 		}
 
-		client = vip.NewDHCPv4Client(iface, initRebootFlag, i.DHCPInterfaceIPv4, backoffAttempts)
+		client = vip.NewDHCPv4Client(iface, initRebootFlag, i.DHCPInterfaceIPv4, backoffAttempts, i.dhcpBroadcast)
 
 		// Add the client so that we can call it to stop function
 		i.DHCPv4Client = client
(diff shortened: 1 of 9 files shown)


Cellebyte:
This sounds like a bug in ciliums ebpf code. Maybe they can add a bypass for dhcp packets?

DavideRutigliano:
A Cilium fix might actually solve but would probably be a complex one to cover this specific problem. However DHCP RFC explicitly provides the broadcast flag for cases like this, and the annotation pattern makes it opt-in so there's no impact on working deployments.

Cellebyte:
@thebsdbox please review :)

thebsdbox:
I'm happy with this, given it is part of the RFC and is opt-in. I'll tag in some of the Cilium maintainers, just in case they have any input or thoughts if any changes are needed in KPR. cc/ @brb & @joestringer 😄

thebsdbox (approved):
/lgtm

brb:
Mind opening an issue in https://github.com/cilium/cilium/issues, so that others are aware of this?

thebsdbox:
@DavideRutigliano I was aiming to create the issue for you, however the issue template requires a ton of detail from your system (kernel version/cilium version etc..) Would you be able to raise the issue?

DavideRutigliano:
@thebsdbox I already created issue [#1577](https://github.com/kube-vip/kube-vip/issues/1577). Shall we use that one? I can add something else there if needed

thebsdbox:
haha, I think Martynas was referring to creating the issue on the Cilium repo so that the devs can track it, as mentioned I tried but it requires a bunch of details that only you know.

PR #1578, opened 2026-05-29
Add endpointless reconciliation after service refactor 

Nabsku:
## Summary

 This PR adds an explicit opt-in for reconciling LoadBalancer Services that intentionally have no endpoints (common with some Gateway API controllers like Cilium), while preserving existing default behavior.

We had an "incident" after updating to v1.2.0, since the Gateway services where not being reconciled anymore. Removing the gateway showed it being in pending state and not getting an IP assigned. Reverting to 1.1.X worked. 

 ## What changed

 - Added new Service annotation:
     - kube-vip.io/allow-reconcile-without-endpoints: "true"
 - Narrow scope:
     - applies only to externalTrafficPolicy: Cluster
     - has no effect for externalTrafficPolicy: Local
     - default behavior remains unchanged for services without this annotation
 - Updated endpoint handling flow so opted-in, endpointless Cluster Services still start service handling paths where required.
 - Added/updated unit tests for:
     - annotation gating behavior
     - zero-endpoint AddOrModify behavior (opt-in Cluster vs non-opt-in/Local)
 - Updated CONTRIBUTING.md to reflect actual make targets 
 
 ## Motivation

 After service handling refactors in #1463, kube-vip waits for endpoint readiness before reconciling.
 That is correct by default, but it blocks valid endpointless LoadBalancer patterns used by some Gateway API implementations.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - Non-disruptive: Process continues running after configuration dump
   - Added comprehensive unit tests for all dump methods
   - Added E2E tests for signal handling
+- Opt-in support for endpointless `LoadBalancer` services with `externalTrafficPolicy: Cluster`
+  - Annotation: `kube-vip.io/allow-reconcile-without-endpoints: "true"`
+  - Starts service handling path for opted-in endpointless Cluster services while preserving default endpoint-gated behavior for non-opt-in services and `Local` policy
+  - Added endpoint behavior tests and README usage documentation
 
 ### Changed
 - Updated signal handlers in manager_arp.go, manager_bgp.go, manager_wireguard.go, and manager_table.go to use switch statement pattern for handling multiple signals (SIGUSR1, SIGINT, SIGTERM)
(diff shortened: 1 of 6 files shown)


p-strusiewiczsurmacki-mobica:
> applies only to externalTrafficPolicy: Cluster

What's the reason for such distinction?

Nabsku:
> > applies only to externalTrafficPolicy: Cluster
> 
> What's the reason for such distinction?

I limited it to `externalTrafficPolicy: Cluster` because this happened with Cilium Gateway services for us specifically.

Those services can intentionally have no Kubernetes Endpoints, but with Cluster policy that is still workable: kube-vip’s job is only to make the VIP present/advertised, while Cilium handles where traffic goes internally.

I did not apply this to `Local` because that felt riskier. With `Local`, advertising a VIP from a node with no local backend can blackhole traffic (AFAIU).

That said, this is based on our Cilium Gateway use case. If there are valid Local-policy cases with no endpoints, or if another implementation needs different behavior, I’m open to discuss and learn.

thebsdbox:
More hacks around Cilium! Someone should have a word with them 😄 This looks good to me, happy to merge this once it passes the tests.

PR #1580, opened 2026-05-29
Remove duplicate IPv6 check functions

mattcarp12:
The functions `pkg/cluster/service.go - isV6()` and `pkg/utils/ip.go - IsIPv6()` were duplicated.

This PR replaces calls to `isV6()` with `IsIPv6()` which is more widely used in the code base.

### DIFF
--- a/pkg/cluster/service.go
+++ b/pkg/cluster/service.go
@@ -5,7 +5,6 @@ import (
 	"errors"
 	"fmt"
 	"io/fs"
-	"net"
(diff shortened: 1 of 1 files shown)


p-strusiewiczsurmacki-mobica (approved):
LGTM

thebsdbox:
Oh wow, I worry if there is other duplication in the code base too. Thanks for this.

thebsdbox (approved):

Issue #1550, opened 2026-05-12
Remove the last requirements for iptables/nftables (shim)


thebsdbox:
**Is your feature request related to a problem? Please describe.**
We have largely replicated most of the needed functionality in the nftables-netlink library to install the various rules for egress in the kernel without external dependancies in the container image.

**Describe the solution you'd like**
To remove the last few call outs where we still need to "shell" out to iptables. 

**Describe alternatives you've considered**
N/A

**Additional context**
Areas that remain:

- [ ] `egress/egress.go` - still has code to create rules in iptables
- [ ] `bip/address.go` - still has code to create firewall rules around the control plane VIP
- [ ] `services/egress.go` - still references the original iptables pkg

p-strusiewiczsurmacki-mobica:
@thebsdbox @Cellebyte  I'll take a look into this.

p-strusiewiczsurmacki-mobica:
@thebsdbox @Cellebyte I have added a PR for `vip/address.go` but I have a few questions, so if you could take a look at it it would be great :)

p-strusiewiczsurmacki-mobica:
@thebsdbox @Cellebyte So, I've spend some time translating iptables code in `services/egress.go` into nftables, only to realize (after I was done) that the `internal egress` feature already implements that. 😅
 I have moved a few changes into already existing PR #1592, and I think all the dependencies on the `iptables` should be now redundant. 

I have set the use of non-iptables implemetation to default for now, but I can also remove all the iptables-related code if you want, but just wanted to be sure, if you want to delete it already, or to keep iptables support for a while as a fallback for the users just in case?

thebsdbox:
Lets leave it in for a little while (3 months) or so

PR #1592, opened 2026-06-09
Added nftables for vip/address.go

p-strusiewiczsurmacki-mobica:
This PR introduces support for `google/nftables` in `vip/address.go` as requested in #1550

**Additional changes and questions**

1. I have decided for now that nftables will be enabled with the already existing `EgressWithNftables` flag (and added that flag as a CLI parameter,). Is this OK, or should we add separate flag that will enable `nftables` use (maybe it could override `EgressWithNftables` when true)? Or maybe `EgressWithNftables` should be completly removed (that might be a breaking change).

~~1. I  am yet unable to figure out how to use `google/nftables` to replace IPVS masquerade rules. The `nft` tool does not support `IPVS` matcher and therefore I can't take a look at the netlink commands that could do that. While investigating this, I've seen some info that this is not really possible, and it might be that we would have to ditch IPVS for internal implementation of loadbalncer with nftables (I believe this is what k8s did), but I am not yet sure. I have omitted this part of code for now, so it still uses `iptables-nft`.~~

~~UPDATE: It seems that instead of using IPVS matcher it should be posible to use IPVS to mark packets with fwmark and then use this mark to match using nftables. Will try that.~~

UPDATE 2: I have implemented the IPVS with `masquerade` forwarding using fwmark and it seems to be OK :)

```
IP Virtual Server version 1.2.1 (size=4096)
Prot LocalAddress:Port               Conns   InPkts  OutPkts  InBytes OutBytes
  -> RemoteAddress:Port
FWM  10001                               7      104       77    11407    21413
  -> kind-control-plane3:6443            3       44       33     4859     9178
  -> kind-control-plane.kind:6443        2       30       22     3274     6117
  -> kind-control-plane2.kind:644        2       30       22     3274     6118
```

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -53,7 +53,7 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v7
       - name: Build image with iptables
-        run: make dockerx86ActionIPTables
+        run: make dockerx86Action
       - name: Run Trivy vulnerability scanner
         uses: aquasecurity/trivy-action@master
         with:
--- a/Makefile
+++ b/Makefile
@@ -150,7 +150,7 @@ e2e-tests: e2e-tests-arp e2e-tests-rt e2e-tests-bgp
 
 service-tests:
 	$(MAKE) -C testing/e2e/e2e dockerLocal
-	E2E_IMAGE_PATH=$(REPOSITORY)/$(TARGET):$(DOCKERTAG) go run ./testing/services -Services -simple -deployments -leaderActive -leaderFailover -localDeploy -egress -egressIPv6 -dualStack
+	E2E_IMAGE_PATH=$(REPOSITORY)/$(TARGET):$(DOCKERTAG) go run ./testing/services -Services -simple -deployments -leaderActive -leaderFailover -localDeploy -egress -egressIPv6 -dualStack -egressInternal
 
 trivy: dockerx86ActionIPTables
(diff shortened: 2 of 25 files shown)


thebsdbox:
We don't need to remove IPVS, the main goal to remove iptables is that there isn't a stable kernel interface so we need to bundle the tooling (and associated dependancies). 😄

p-strusiewiczsurmacki-mobica:
> We don't need to remove IPVS, the main goal to remove iptables is that there isn't a stable kernel interface so we need to bundle the tooling (and associated dependancies). 😄

Yes, I have implemented this with fwmark and it seems to work just fine

```
IP Virtual Server version 1.2.1 (size=4096)
Prot LocalAddress:Port               Conns   InPkts  OutPkts  InBytes OutBytes
  -> RemoteAddress:Port
FWM  10001                               7      104       77    11407    21413
  -> kind-control-plane3:6443            3       44       33     4859     9178
  -> kind-control-plane.kind:6443        2       30       22     3274     6117
  -> kind-control-plane2.kind:644        2       30       22     3274     6118
```

thebsdbox:
Amazing stuff ! 🎉

Cellebyte:
If we remove ipvs as well we would comply with kube-proxy in kubernetes as they deprecated ipvs as well.

thebsdbox:
This also looks good, this will need rebasing due to the gobgp dependancy.

p-strusiewiczsurmacki-mobica:
@thebsdbox rebased :)

Issue #1590, opened 2026-06-09
StartVipService: defer wg.Wait() deadlock when AddIP or NewIPVSLB returns error


zhulinwei:
Describe the bug

StartVipService has a goroutine deadlock. When network.AddIP() or loadbalancer.NewIPVSLB() returns an error, defer wg.Wait() blocks forever waiting on a goroutine that only exits on ctx.Done() — which hasn't been cancelled yet. The kube-vip process becomes unresponsive (no new logs, no metrics updates) without crashing.

Introduced by commit 88efcc7 (PR #1434, "Added waitgroups"), present in v1.1.0 through current main. v1.0.4 is unaffected.

To Reproduce

1. Deploy kube-vip v1.2.0 as a static pod with ARP mode + IPVS masquerade load balancing (see manifest below)
2. Watch logs: kubectl -n kube-system logs kube-vip-<node> -f
3. Logs stop after leader election succeeds, with no IPVS creation or error messages:
```
I0609 11:25:55.863999 leaderelection.go:272] "Successfully acquired lease"
2026/06/09 11:25:55 INFO New leader leader=10-100-200-111
```
The process hangs here — no further output.

Root cause

In pkg/cluster/service.go, AddIP returns an error because the container image doesn't include an iptables binary, yet AddIP → configureIPTables() unconditionally calls iptables --version. v1.0.4 only logged this error and continued; v1.1.0+ returns it, which triggers defer wg.Wait() before the context-cancellation goroutine can exit.

Code pattern in v1.1.0+:
```golang
var wg sync.WaitGroup
defer wg.Wait()           // (1) runs on every return

wg.Go(func() {            // (2) only exits when ctx.Done()
    <-ctx.Done()
    killFunc()
})

if _, err = network.AddIP(false, false); err != nil {
    return fmt.Errorf(...) // (3) triggers (1) → waits forever on (2) → deadlock
}

Code in v1.0.4 (no deadlock):
if _, err = network.AddIP(false, false); err != nil {
    log.Error(err.Error())  // logs only, continues to IPVS creation
}
```
Expected behavior

AddIP's iptables error should not be fatal. IPVS masquerade performs SNAT in the kernel netfilter framework and does not require the iptables userspace binary. The function should log the error and proceed to create the IPVS service — as v1.0.4 did.

Environment

- OS: Ubuntu 20.04.3 LTS
- Kubernetes Server: v1.32.0
[...]

PR #1594, opened 2026-06-11
Fixed deadlock on AddIP() call in StartVipService

p-strusiewiczsurmacki-mobica:
This PR sgould fix deadlock on AddIP() call in StartVipService as reported in #1590 

I have confirmed that the issue occurs. The error return was changed to log (As it is for other `AddIP()` calls. I have additionally added calls to `killFunc` in case of other errors that can occur in this function, this should result in context being cancelled in case any error would occur.

### DIFF
--- a/pkg/cluster/service.go
+++ b/pkg/cluster/service.go
@@ -58,6 +58,7 @@ func (cluster *Cluster) StartVipService(ctx context.Context, c *kubevip.Config, 
 		}
 
 		if err := network.SetMask(c.VIPSubnet); err != nil {
+			killFunc()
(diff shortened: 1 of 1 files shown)


thebsdbox:
Looks good, this will need rebasing due to the gobgp dependancy.

p-strusiewiczsurmacki-mobica:
@thebsdbox rebased :)

Cellebyte (approved):
LGTM
Is it somehow possible to not do a conditional defer?

p-strusiewiczsurmacki-mobica:
> Is it somehow possible to not do a conditional defer?

Yeah, you're right, defer is redundant here. Deleted that.

PR #1616, opened 2026-07-01
Fixed BGP metrics

### LINKED ISSUE #1614: Metrics broken after gobgp v3 -> v4 upgrade
**Describe the bug**
The Metrcis reported by kubevip are shifted by one? and thus are not correct anymore.

I think the problem is because of this: https://github.com/osrg/gobgp/blob/8c4802dadc416a6c78e5b244c1b270c2b8665e31/pkg/server/grpc_server.go#L443

and the comparison kubevip is doing here: https://github.com/kube-vip/kube-vip/blob/v1.2.1/pkg/manager/worker/bgp.go#L62-L65

`p.Peer.State.SessionState` now seems to come from the internal GoBGP FSM state.

But `api.PeerState_SessionState_value` is the public API enum used for labels like:

the api adds +1 to the sessionsState and kube-vip is comparing fsm state against the api enum which is shfited by 1. 

So internally:

`BGP_FSM_ESTABLISHED = 5`

But in the API enum:

```
SESSION_STATE_OPENCONFIRM = 5
SESSION_STATE_ESTABLISHED = 6
```

Because of that, kube-vip marks OPENCONFIRM as active, even though the real state is ESTABLISHED.

**To Reproduce**

with kube-vip 1.2.0 installed, check metrics reported:

````yaml
kube_vip_manager_bgp_session_info{state="ESTABLISHED"} = 1
kube_vip_manager_bgp_session_info{state="OPENCONFIRM"} = 0
````

Update to kube-vip 1.2.1 and check again:

````yaml
kube_vip_manager_bgp_session_info{state="SESSION_STATE_ESTABLISHED"} = 0
kube_vip_manager_bgp_session_info{state="SESSION_STATE_OPENCONFIRM"} = 1
````

kube-vip works normally, metrics are wrong.

**Expected behavior**
kube-vip reporting the metrics and the state correctly so alerting works as expected.

 - Kubernetes Version: [e.g. v.1.36.2]
 - Kube-vip Version: [e.g. 1.2.1]

p-strusiewiczsurmacki-mobica:
This should fix #1614 

There is a mismatch in the internal state values and API state values by one, as the API also has the value for `unspecified` state.

### DIFF
--- a/pkg/manager/worker/bgp.go
+++ b/pkg/manager/worker/bgp.go
@@ -59,7 +59,7 @@ func (b *BGP) Configure(ctx context.Context, _ *sync.WaitGroup) error {
 
 		for stateName, stateValue := range api.PeerState_SessionState_value {
 			metricValue := 0.0
-			if int(p.Peer.State.SessionState) == int(stateValue) {
+			if int(p.Peer.State.SessionState) == int(stateValue)-1 {
 
 				metricValue = 1
 			}


Cellebyte (approved):
LGTM

PR #1627, opened 2026-07-09
Honor the explicitly configured kubeconfig path in manager and backend health checks

mkowalski:
`kubeConfigPath` from `--k8sConfigPath` is now ignored in manager.go and backend.go; we have noticed it when running a POC with kube-vip in routing table mode deployed as a static pod with kubeconfig in a custom path `/etc/kubernetes/kubeconfig`

The patch has been tested using out downstream fork of kube-vip.

### DIFF
--- a/pkg/backend/backend.go
+++ b/pkg/backend/backend.go
@@ -21,6 +21,16 @@ type Entry struct {
 
 type Map map[Entry]bool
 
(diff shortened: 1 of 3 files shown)


thebsdbox:
Merging this as we know the issue with is with the test itself

PR #1635, opened 2026-07-14
fix(egress): isolate nftables tables by instance

### LINKED ISSUE #1634: Allow per-deployment nftables egress table names
## Is your feature request related to a problem? Please describe.

When two kube-vip deployments run on the same cluster nodes with the internal nftables egress implementation and `egress_clean` enabled, both deployments use the fixed `kube_vip_v4` and `kube_vip_v6` tables. On startup, either deployment can delete and recreate the shared tables, removing the other deployment's egress rules.

Kubernetes namespaces do not isolate nftables state because the tables live in the host network namespace.

## Describe the solution you'd like

Add an optional, per-deployment nftables egress table base name. kube-vip should append `_v4` or `_v6` and use the resulting table consistently for rule creation, service teardown, and startup cleanup. The empty/default value should retain the current `kube_vip_v4` and `kube_vip_v6` names for backward compatibility.

For example, a base name of `kube_vip_release_a` should create `kube_vip_release_a_v4` and `kube_vip_release_a_v6`.

## Describe alternatives you've considered

Using unique chain names alone is insufficient because `egress_clean` deletes the complete table. Disabling cleanup avoids that deletion but leaves stale egress state and does not provide ownership isolation.

## Additional context

This follows the centralized internal nftables cleanup introduced in #1447.

GabboPenna:
## What this PR does

Adds an optional, generic identity for a kube-vip instance. In this PR it is used to namespace the internal nftables egress tables, preventing multiple kube-vip deployments on the same nodes from deleting or modifying each other's nftables state when `egress_clean` is enabled.

The instance name can be configured with:

- environment variable: `instance_name`
- CLI flag: `--instanceName`
- config file field: `instanceName`

For example, `release_a` produces `kube_vip_release_a_v4` and `kube_vip_release_a_v6`. An empty value preserves the existing `kube_vip_v4` and `kube_vip_v6` names.

`InstanceName` lives in the general configuration so the same identity can be reused to address other host-state conflicts in the future. This PR currently applies it only to nftables egress state; it does not claim that every kube-vip mode is already isolated.

## Validation

The instance name:

- allows only ASCII letters, digits, `.`, `-`, and `_`;
- is limited to 243 bytes, reserving space for the `kube_vip_` prefix and `_v4` / `_v6` suffix within nftables' 255-byte limit;
- remains optional for backward compatibility.

Invalid values fail configuration validation instead of being silently sanitized or truncated. Generated Pod and DaemonSet manifests propagate `instance_name` to the kube-vip container.

## Upgrade and ownership handling

After successfully applying egress rules, kube-vip records the effective table base name in the Service annotation:

    kube-vip.io/egress-nftables-table: kube_vip_release_a

During an upgrade or instance-name change:

1. kube-vip replaces the Service UID chain in the target table, making retries idempotent;
2. it applies the new SNAT rules;
3. it discovers the exact `kube_vip_snat_<Service UID>` chain locally and removes it from all other tables in the same address family;
4. it persists the target table annotation only after rules were actually applied.

The annotation is cluster-wide but nftables state is node-local. Every kube-vip pod that observes the annotated Service therefore performs the same Service-UID-scoped stale-chain cleanup on its own node, after the existing namespace/load-balancer-class filters. Teardown checks both address families and does not depend on an active endpoint or a potentially stale annotation.

A stale chain belonging to a Service deleted before upgrading cannot be attributed through Service metadata and remains an operational cleanup case.

## Additional implementation details

- startup `egress_clean` clears only the tables derived for the current instance;
- the IPv4/IPv6 lookup used by table cleanup is corrected and the cleanup transaction is flushed;
- legacy nftables entry points retain default-table wrappers;
- the instance and effective table base are included in the runtime configuration dump;
- tests cover environment/YAML/manifest propagation, validation boundaries, default/custom names, and exact Service UID cleanup;
- the unreleased changelog is updated.

## Testing

- `gofmt` on changed Go files
- `git diff --check`
- Linux cross-compilation of test binaries for all affected packages
- `GOOS=linux GOARCH=amd64 go vet ./...`

Full runtime unit and e2e tests are left to Linux CI because nftables/netlink dependencies do not build natively on Windows.

Closes #1634

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Fixed
+- Add a configurable kube-vip instance name and use it to isolate internal nftables egress tables, persist table ownership on Services, and migrate per-Service chains without affecting other deployments. Fixes #1634.
 - Retry on 403 Forbidden and 401 Unauthorized in `ServicesWatcher` at startup with exponential backoff. Fixes #1464.
 - Reintroduce BGP config via node annotations. Fixes #1488.
 - Fail fast in runtime `manager` and `service` paths when legacy `vip_address` is used without `vip_subnet` in control-plane ARP, BGP, or Routing Table mode.
--- a/cmd/kube-vip.go
+++ b/cmd/kube-vip.go
@@ -167,6 +167,7 @@ func init() {
 
 	// Configuration file flag
 	kubeVipCmd.PersistentFlags().StringVar(&initConfig.ConfigFile, "config-file", "", "Path to a JSON/YAML configuration file to load settings from")
+	kubeVipCmd.PersistentFlags().StringVar(&initConfig.InstanceName, "instanceName", "", "Unique name for this kube-vip instance (currently used to isolate nftables egress tables)")
 
 	kubeVipCmd.PersistentFlags().BoolVar(&initConfig.EgressWithNftables, "egressWithNftables", true, "Use nftables-based egress implementation")
 
--- a/pkg/egress/egress.go
+++ b/pkg/egress/egress.go
@@ -19,15 +19,13 @@ func Teardown(podIP, vipIP, namespace, serviceUUID string, annotations map[strin
 	internalEgress := annotations[kubevip.EgressInternal]
 
 	protocol := iptables.ProtocolIPv4
-	IPv6 := false
 	if utils.IsIPv6(podIP) {
 		protocol = iptables.ProtocolIPv6
-		IPv6 = true
 	}
 
 	// Use the internal egress implementation
 	if internalEgress != "" || useNftables {
-		return nftables.DeleteSNAT(IPv6, serviceUUID)
+		return nftables.DeleteSNATFromAllTables(serviceUUID)
 	}
 
 	i, err := vip.CreateIptablesClient(useNftables, namespace, protocol)
--- a/pkg/kubevip/annotations.go
+++ b/pkg/kubevip/annotations.go
@@ -37,6 +37,9 @@ const (
 	// The current active endpoint(pod) for the Egress VIP (v6)
 	ActiveEndpointIPv6 = "kube-vip.io/active-endpoint-ipv6"
 
+	// The nftables egress table base name that owns this Service's SNAT chain
+	EgressNftablesTable = "kube-vip.io/egress-nftables-table"
+
 	// Flush the conntrack rules (remove existing sessions) once Egress is configured
 	FlushContrack = "kube-vip.io/flush-conntrack"
 
--- a/pkg/kubevip/config_environment.go
+++ b/pkg/kubevip/config_environment.go
@@ -32,6 +32,13 @@ func ParseEnvironment(c *Config) error {
 		c.Logging = int32(logLevel)
 	}
 
+	if env = os.Getenv(instanceName); env == "" {
+		env = os.Getenv(strings.ToUpper(instanceName))
+	}
+	if env != "" {
+		c.InstanceName = env
+	}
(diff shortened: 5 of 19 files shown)


GabboPenna:
Updated based on the upgrade concern raised by @thebsdbox.

The PR now persists the effective base name in `kube-vip.io/egress-nftables-table`. Migration is scoped to the exact `kube_vip_snat_<Service UID>` chain and uses local chain discovery, so each node can remove an old chain even when another node has already updated the cluster-wide annotation. The annotation is written only after rules were successfully applied, and teardown checks both address families.

This keeps legacy-to-custom and custom-to-custom rolling upgrades isolated without deleting a shared legacy table.

thebsdbox:
Thanks @GabboPenna the issues raIsed in the github actions is not related to this PR and I will fix afterwards.

Cellebyte:
Hey, if we allow configuring the deployment-name as a environment variable or cli flag, we would need to ensure that either the input cannot be longer then what is left from `255 chars - prefix length - suffix length` check https://github.com/telekom/multi-networkpolicy-nftables/blob/nftables/pkg/server/netfilterrules_sanitize_test.go .

Cellebyte:
@GabboPenna I also think it is now conflicting for egress but could also happen on other host based configurations as well. 
So it would require to be a more generic configuration name like `instance_name`.

environment variable: `instance_name`
CLI flag: `--instanceName`
config file field: `instanceName`

This would also require a validation if it even works for multiple instances in other configurations not only `egress`.
Maybe it would require a rethink and would make more sense to move the instance somewhere else and enable kube-vip to handle your use-case for which you require 2 separate deployments.

GabboPenna:
Thanks @Cellebyte, good points.

I can add validation for the nftables name length and allowed characters.

Regarding `instance_name`, I agree that it would suggest broader isolation than this PR currently provides. Our use case is specifically two routing-table deployments with separate LB classes and routing-table IDs, where the remaining collision is the egress nftables table.

Would you prefer keeping this change egress-specific and handling a generic instance identity separately, or should we rethink this PR around `instance_name`?

GabboPenna:
I've added the nftables name validation.

The configured base name now:
- allows only ASCII letters, digits, `.`, `-`, and `_`;
- is limited to 252 bytes, reserving 3 bytes for the `_v4` / `_v6` suffix;
- keeps an empty value valid for the legacy `kube_vip` default.

I chose to reject invalid configured values rather than silently sanitize or truncate them, since this value is also persisted as the Service's table identity. Boundary and invalid-character tests are included.

Cellebyte:
I would prefer to have the general instanceName flag this would allow to retrofit similar conflicts with the same input. @thebsdbox what do you think?

GabboPenna:
Thanks @Cellebyte - I went ahead with the generic `instanceName` approach.

The PR now exposes `instance_name`, `--instanceName`, and `instanceName`, and derives the egress tables from it. For example, `release_a` becomes `kube_vip_release_a_v4` / `kube_vip_release_a_v6`; leaving it empty keeps the legacy names.

I also updated the validation to account for the full prefix and suffix (243-byte maximum, with the same ASCII character set), and made sure generated manifests pass `instance_name` to the container.

For now the value is only consumed by nftables egress, but it lives in the general config so the same instance identity can be reused for other host-state conflicts later.

thebsdbox:
If this is ready to merge, i'll pull it in and then fix the image vulnerability issue. All the License issues have been fixed in FOSSA.

GabboPenna:
Thanks @thebsdbox ... for me its ready

thebsdbox:
One additional ask, is to update the docs in the `/website` repository.

GabboPenna:
@thebsdbox  Sure, thanks a lot Dan! I opened kube-vip/website#86.

PR #1636, opened 2026-07-15
routing table mode: re-assert the VIP route on every healthy check cycle

mkowalski:
In order to workaround https://github.com/FRRouting/frr/issues/22654 where zebra misses updates if multiple are bundled in the same netlink event, we are introducing a new mechanism.

On every cycle we are re-asserting the route with an idempotent RouteReplace. Because for a no-op replace kernel emits no notification, we are alternating the realm attribute on every cycle. Thanks to this, every re-asstertion is a real change which emits the notification.

Wire impact of this change is none. The prefix, nexthop, metric never change so BGP daemons do not send UPDATE to peers.

Another option was delete+add but that one would have a real wire impact if performed too slow.


thebsdbox:
Will merge this and then bump golang to fix vulnerability.

PR #1639, opened 2026-07-21
Add an opt-in flag/env var to allow kube-vip to start when interface exists but is not up 

gabrielstedman:
Add an opt-in flag/env var ( `--allowInterfaceNotUp`  /  `vip_allow_interface_not_up` ) that lets kube-vip continue operating even when its bind interface exists but is not yet UP .

Currently, kube-vip fails immediately if the configured interface is down at startup. This can be problematic on single-node, edge, and bare-metal clusters where network interfaces may become operational only after kube-vip starts. In some cases this may be a transient condition that resolves on its own, so failing immediately isn't always desirable.

With this change, when `allowInterfaceNotUp` is enabled, kube-vip logs a warning and continues instead of erroring out.

This is disabled by default and introduces no change to existing behavior. Users who want the more tolerant startup behavior must explicitly opt in via the new flag or environment variable.

### DIFF
--- a/cmd/kube-vip.go
+++ b/cmd/kube-vip.go
@@ -60,6 +60,7 @@ func init() {
 	// Basic flags
 	kubeVipCmd.PersistentFlags().StringVar(&initConfig.Interface, "interface", "", "Name of the interface to bind to")
 	kubeVipCmd.PersistentFlags().StringVar(&initConfig.ServicesInterface, "serviceInterface", "", "Name of the interface to bind to (for services)")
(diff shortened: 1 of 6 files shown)


Cellebyte:
@gabrielstedman is this not something the correct ordering on the node could fix?

gabrielstedman:
@Cellebyte Generally speaking, yes this would be the standard approach in most cases.
This helps with edge scenarios where network may be unreliable and other workloads are resilient to network outages.
By using this option, the kubelet can successfully communicate with the VIP address and allow these other pods to come up.

thebsdbox:
This seems like a valid use case, would you be able to update the documentation too?

gabrielstedman:
> This seems like a valid use case, would you be able to update the documentation too?

Certainly, could you point me to the documentation that needs updating and I'll open a PR for it?

Issue #1640, opened 2026-07-22
A crash occurs when routing table mode is used


yangyangai520:
**Describe the bug**
New cluster with 3 nodes, after installing kube-vip as a DaemonSet, the pods keep crashing right after startup with a nil pointer dereference. 

```
2026/07/22 10:21:34 INFO kube-vip.io version=v1.2.0 build=e2a0e815fe3e0727896464ed184f649785f5dbb7
2026/07/22 10:21:34 INFO starting namespace=kube-system Mode="" "Control Plane"=true Services=true
2026/07/22 10:21:34 INFO prometheus HTTP server started
2026/07/22 10:21:34 INFO using node name name=ecs-test-001
2026/07/22 10:21:34 INFO prometheus HTTP server stopped
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x1e6399b]

goroutine 1 [running]:
github.com/kube-vip/kube-vip/pkg/manager.(*Manager).startMode.func1()
        /src/pkg/manager/manager.go:345 +0x3b
panic({0x219ee20?, 0x3e423d0?})
        /usr/local/go/src/runtime/panic.go:860 +0x13a
github.com/kube-vip/kube-vip/pkg/manager.(*Manager).startMode(0x2355d2068000, {0x2749688, 0x2355d24e6000})
        /src/pkg/manager/manager.go:350 +0x1d6
github.com/kube-vip/kube-vip/pkg/manager.(*Manager).Start(0x2355d2068000, {0x2749688, 0x2355d24e6000})
        /src/pkg/manager/manager.go:326 +0x2bc
github.com/kube-vip/kube-vip/cmd.init.func11(0x3e66d40, {0x25c43ac?, 0x4?, 0x25c43b0?})
        /src/cmd/kube-vip.go:465 +0xfea
github.com/spf13/cobra.(*Command).execute(0x3e66d40, {0x3ea1540, 0x0, 0x0})
        /go/pkg/mod/github.com/spf13/cobra@v1.10.2/command.go:1019 +0xafb
github.com/spf13/cobra.(*Command).ExecuteC(0x3e661c0)
        /go/pkg/mod/github.com/spf13/cobra@v1.10.2/command.go:1148 +0x465
github.com/spf13/cobra.(*Command).Execute(...)
        /go/pkg/mod/github.com/spf13/cobra@v1.10.2/command.go:1071
[...]

PR #1641, opened 2026-07-24
fix(cmd): throw an error when no valid mode detected

thebhdn:
Hello gentlemen,

Follow-up https://github.com/kube-vip/kube-vip/issues/1640

Small fix to throw an error when no mode defined.  Also added check if someone went full yolo and enabled more that 1 mode - [kube-vip modes](https://kube-vip.io/docs/modes/)

cc @Cellebyte @p-strusiewiczsurmacki-mobica

### DIFF
--- a/cmd/kube-vip.go
+++ b/cmd/kube-vip.go
@@ -359,21 +359,38 @@ var kubeVipManager = &cobra.Command{
 		}
 
 		// Determine the kube-vip mode
-		var mode string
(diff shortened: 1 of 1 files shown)


Cellebyte (commented):
LGTM I am not sure if we have people relying on more than one mode.

p-strusiewiczsurmacki-mobica (approved):

PR #1664, opened 2026-08-04
services: recreate the lease when the service context was cancelled

### LINKED ISSUE #1663: Service leader election permanently wedges after the service watcher is rebuilt: "no existing lease found for service" (v1.2.1, v1.2.2)
**Describe the bug**

kube-vip permanently stops advertising a LoadBalancer VIP after its service watcher is torn down and rebuilt — for example after an API-server restart, an etcd `storage is (re)initializing` episode, or the loss of a control-plane node. Once it happens, the process logs this on every subsequent service event, forever:

```
ERROR no existing lease found for service "<svc>" with UID "<uid>"
```

The Lease object's `renewTime` stops advancing, no node holds the address, and only restarting the pod clears it.

**Mechanism**

`Processor.AddOrModify` only creates a lease inside its `if svcCtx == nil` branch:

```go
if svcCtx == nil {
    ns, name := lease.ServiceName(svc)
    leaseID := lease.NewID(p.config.LeaderElectionType, ns, name)
    lease := p.leaseMgr.Add(ctx, leaseID)
    svcCtx = servicecontext.New(lease.Ctx)
    p.svcMap.Store(svc.UID, svcCtx)
}
```

but the in-memory lease is removed independently, by the cleanup goroutine in `StartServicesLeaderElection`:

```go
go func() {
    <-svcCtx.Ctx.Done()
    p.leaseMgr.Delete(id, objectName)
}()
```

and `Manager.Delete` drops the entry once its refcount reaches zero.

Several paths cancel the service context **without** also removing it from `p.svcMap` — for example the deferred `close(stopChan)` in `watchEndpoint` (which triggers `svcCtx.Cancel()`), and the `utils.PanicError` branch in `AddOrModify`.

Afterwards `p.svcMap` still holds a **cancelled** context for that UID. Every later watch event therefore takes the `svcCtx != nil` path, skips `leaseMgr.Add`, and `p.leaseMgr.Get(id)` returns `nil` in `StartServicesLeaderElection`:

```go
svcLease := p.leaseMgr.Get(id)
if svcLease == nil {
    return fmt.Errorf("no existing lease found for service %q with UID %q", service.Name, service.UID)
}
```

Leader election never restarts, so the address is never re-assigned.

**This is not fixed by #1650.** That change detached the cleanup goroutine from the `WaitGroup`, removing the `wg.Wait()` deadlock, but the `svcMap` ↔ `leaseMgr` desync above is a separate defect and still reproduces. As of 2026-08-03, `main` carries only `68123d3` (discard invalid BGP metrics events) and `c3ba4a8` (default kubeconfig path handling) beyond `v1.2.2`, neither of which touches this.

**To Reproduce**

1. Run kube-vip as a DaemonSet with per-service leader election (`svc_enable=true`, `svc_election=true`) managing a `LoadBalancer` Service.
2. Disrupt the API server / etcd, or lose a control-plane node, so the service watcher is torn down and rebuilt.
3. Observe the Lease object's `renewTime` freeze, the VIP present on no node (`ip addr`), and `no existing lease found for service ... with UID ...` repeating in the logs.
4. It never self-heals; restarting the pod does.

**Expected behavior**

After the watcher is rebuilt, leader election restarts for the service and the VIP is re-advertised without operator intervention.

**Environment**

- Kubernetes Version: v1.36
- Kube-vip Version: **v1.2.1 and v1.2.2** (both affected)
- Mode: DaemonSet, services enabled, per-service leader election, ARP/NDP

**Impact observed**

A management cluster's ingress LoadBalancer VIP was unadvertised for roughly 18 days before the cause was found, and then re-wedged within hours of upgrading to v1.2.2 during a control-plane disruption. Downstream clusters that depend on that VIP for DNS failed closed each time, so the blast radius is considerably wider than the one Service that holds the address.

**Proposed fix**

Treat a cancelled service context as absent, so the existing code path creates a fresh context and lease and the invariant "a service context in `svcMap` always has a matching lease in the lease manager" is restored. Draft PR to follow.

An alternative — having `StartServicesLeaderElection` call `leaseMgr.Add` when `Get` returns `nil` — was considered and rejected, because it would attach the lease lifetime to the service context rather than the watcher context, inverting the intended ownership model.

MaxRink:
Fixes #1663.

`AddOrModify` only calls `leaseMgr.Add` inside its `if svcCtx == nil` branch, but the in-memory lease is removed independently: the cleanup goroutine started by `StartServicesLeaderElection` calls `leaseMgr.Delete` once `svcCtx.Ctx` is done, and `Manager.Delete` drops the lease once its last object goes away.

Several paths cancel the service context **without** also removing it from `svcMap` — the deferred `close(stopChan)` in `watchEndpoint`, and the `utils.PanicError` branch in `AddOrModify`. Afterwards `svcMap` still holds a cancelled context for that UID, so every later watch event reuses it, skips `leaseMgr.Add`, and `StartServicesLeaderElection` fails on

```
no existing lease found for service %q with UID %q
```

for the lifetime of the process. Leader election never restarts and the address is never re-advertised; only restarting the pod clears it.

This is distinct from #1650, which removed the `wg.Wait()` deadlock but not this `svcMap` ↔ `leaseMgr` desync. Both **v1.2.1 and v1.2.2** are affected.

## Change

`dropCancelledServiceContext` discards a service context whose context has already been cancelled, removing it from `svcMap` and returning `nil`, so the existing `if svcCtx == nil` branch creates a fresh context and a fresh lease. That restores the invariant that a service context in `svcMap` always has a matching lease in the lease manager.

It is a separate small method rather than an inline `if` specifically so the behaviour is directly unit-testable.

### Rejected alternative

Having `StartServicesLeaderElection` call `leaseMgr.Add` when `Get` returns `nil` would also stop the error, but it attaches the lease lifetime to the *service* context rather than the *watcher* context, inverting the intended ownership model. The `svcMap` purge keeps the existing ownership intact.

## Tests

`pkg/services/processor_lease_test.go`:

- `TestDropCancelledServiceContext` — a cancelled context is dropped **and** removed from `svcMap`; a live context is kept; `nil` is a no-op.
- `TestDropCancelledServiceContextAllowsLeaseRecreation` — demonstrates the consequence: with the stale context dropped, the caller takes the `svcCtx == nil` branch and a lease exists again, so `StartServicesLeaderElection` no longer fails with `no existing lease found`.

```
go build ./...                                  ok
go vet   ./pkg/services/                        ok
go test  ./pkg/services/... ./pkg/lease/...     ok  (both packages)
go test  ./pkg/services/ -run TestDropCancelledServiceContext -v
    --- PASS: TestDropCancelledServiceContext
    --- PASS: TestDropCancelledServiceContext/cancelled_context_is_dropped_and_removed_from_svcMap
    --- PASS: TestDropCancelledServiceContext/live_context_is_kept
    --- PASS: TestDropCancelledServiceContext/nil_context_is_a_no-op
    --- PASS: TestDropCancelledServiceContextAllowsLeaseRecreation
```

## Notes

Opened as a draft: the fix and tests are complete and validated locally, but I would welcome maintainer input on whether you would rather see the guard inline in `AddOrModify`, and on whether any of the other `svcCtx.Cancel()` call sites should instead be purging `svcMap` themselves.

### DIFF
--- a/pkg/services/processor.go
+++ b/pkg/services/processor.go
@@ -165,6 +165,7 @@ func (p *Processor) AddOrModify(ctx context.Context, event watch.Event, serviceF
 		metrics.ServiceReconcileErrorsTotal.WithLabelValues(svc.Namespace, svc.Name, "service_context").Inc()
 		return fmt.Errorf("failed to get service context: %w", err)
 	}
+	svcCtx = p.dropCancelledServiceContext(svc.UID, svcCtx)
 
 	// The modified event should only be triggered if the service has been modified (i.e. moved somewhere else)
 	if event.Type == watch.Modified {
@@ -363,6 +364,28 @@ func (p *Processor) getServiceContext(uid types.UID) (*servicecontext.Context, e
 		return nil, fmt.Errorf("failed to cast service context pointer - UID: %s", uid)
 	}
 	return ctx, nil
+}
+
+// dropCancelledServiceContext discards a service context whose context has already been
+// cancelled, removing it from svcMap and returning nil so that callers create a fresh one.
+//
+// This matters because the in-memory lease and the service context are removed independently.
+// The cleanup goroutine started by StartServicesLeaderElection calls leaseMgr.Delete once
+// svcCtx.Ctx is done, and Manager.Delete drops the lease entirely when its last object goes
+// away. Several paths cancel the service context without also removing it from svcMap - for
+// example the deferred close(stopChan) in watchEndpoint, and the utils.PanicError branch in
+// AddOrModify.
+//
+// If such a cancelled context were reused, AddOrModify would skip its `if svcCtx == nil`
+// branch and therefore never call leaseMgr.Add again, so StartServicesLeaderElection would
+// fail with "no existing lease found" on every subsequent event and the VIP would never be
+// advertised again.
+func (p *Processor) dropCancelledServiceContext(uid types.UID, svcCtx *servicecontext.Context) *servicecontext.Context {
+	if svcCtx == nil || svcCtx.Ctx.Err() == nil {
+		return svcCtx
+	}
+	p.svcMap.Delete(uid)
+	return nil
 }
 
 func serviceChanged(i *instance.Instance, svc *v1.Service) bool {
--- a/pkg/services/processor_lease_test.go
+++ b/pkg/services/processor_lease_test.go
@@ -1,0 +1,119 @@
+package services
+
+import (
+	"context"
+	"testing"
+
+	"github.com/kube-vip/kube-vip/pkg/kubevip"
+	"github.com/kube-vip/kube-vip/pkg/lease"
+	"github.com/kube-vip/kube-vip/pkg/servicecontext"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+// TestDropCancelledServiceContext is a regression test for the lease/svcMap desync that
+// permanently stops a LoadBalancer VIP from being advertised.
+//
+// AddOrModify only calls leaseMgr.Add inside its `if svcCtx == nil` branch, while the
+// in-memory lease is removed independently by the cleanup goroutine in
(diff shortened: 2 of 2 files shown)


p-strusiewiczsurmacki-mobica (approved):
LGTM

PR #1666, opened 2026-08-05
services: start the leader-election restart loop at most once per service

### LINKED ISSUE #1665: Duplicate leader-election restart-loop goroutines under EnableServicesElection cause a service Lease to permanently lose its holder
## Summary

`pkg/endpoints/endpoints.go`'s `startServiceHandlingIfNeeded` spawns a long-lived, self-restarting `startLeaderElection` goroutine on **every** `AddOrModify` call that sees non-zero endpoints for a `EnableServicesElection` (per-service leader election) LoadBalancer service with no guard against spawning a duplicate for a service that already has one running. The ARP-only branch immediately below it *does* have this guard (`if !svcCtx.Signalled.Load()`); the services-election branch does not:

https://github.com/kube-vip/kube-vip/blob/v1.2.2/pkg/endpoints/endpoints.go#L247-L262

```go
func (p *Processor) startServiceHandlingIfNeeded(svcCtx *servicecontext.Context, service *v1.Service,
	serviceFunc func(*servicecontext.Context, *v1.Service, *sync.WaitGroup, bool) error, wg *sync.WaitGroup, les *atomic.Int64) error {
	if p.config.EnableServicesElection {
		wg.Go(func() {
			les.Add(1)
			p.startLeaderElection(svcCtx, service, serviceFunc, wg)
		})
		return nil
	}
	if p.config.EnableARP || (p.config.EnableRoutingTable && p.config.EnableLeaderElection) {
		if !svcCtx.Signalled.Load() {
			// ... only started once ...
		}
	}
	return nil
}
```

`startLeaderElection` (`pkg/endpoints/endpoints.go`) runs an unconditional `for {}` loop until `svcCtx.Ctx` is cancelled, re-attempting election on its own whenever the service's `Lease.Elected` flag is false:

```go
func (p *Processor) startLeaderElection(svcCtx *servicecontext.Context, service *v1.Service, serviceFunc ..., wg *sync.WaitGroup) {
	for {
		select {
		case <-svcCtx.Ctx.Done():
			return
		default:
			l := p.leaseMgr.Get(id)
			l.Lock()
			if !l.Elected.Load() {
				l.Unlock()
				err := serviceFunc(svcCtx, service, wg, true) // StartServicesLeaderElection
				...
			} else {
				l.Unlock()
				time.Sleep(time.Millisecond * 200)
			}
		}
	}
}
```

Because this goroutine already self-restarts forever, it only ever needs to be spawned **once** per service's lifetime. But `AddOrModify` is invoked on every EndpointSlice add/modify/resync event for the service: including a zero-endpoints teardown immediately followed by a healthy endpoint reappearing and each such event with non-zero endpoints spawns another copy. Under endpoint churn (a backend pod cycling in and out, e.g. during a StatefulSet/AG rolling upgrade), a single service can accumulate many concurrent, duplicate restart-loop goroutines, all contending for the same `svcLease.Lock()`/`Elected` state in `pkg/services/leader.go`.

## Impact observed

On a real workload cluster, a `LoadBalancer` Service selecting a SQL Server Availability Group's current primary pod (`role: primary`) legitimately saw a brief zero-endpoint window during an AG primary failover that overlapped with a rolling control-plane node replacement. kube-vip correctly tore down the VIP and released leadership:

```
WARN existing endpoint has been removed, no remaining endpoints for leaderElection  provider=... endpoint=...
INFO leadership lost  service=matst-z1euia uid=... leader=...
INFO stopping leader election  service=matst-z1euia uid=...
[ARP] removing instance
[LOADBALANCER] Stopping load balancers
[VIP] Deleting VIP
```

The backend endpoint came back healthy shortly after, but the Service's `Lease` object never reacquired a holder:

```
$ kubectl get lease kubevip-matst-z1euia -n dsm-system -o yaml
spec:
  acquireTime: "2026-08-04T19:04:10.003444Z"
  holderIdentity: ""
  leaseDurationSeconds: 1
  leaseTransitions: 4
  renewTime: "2026-08-04T19:04:10.003444Z"
```

`acquireTime == renewTime`, frozen at the exact moment the VIP was deleted for over 11 minutes with zero renewals, meaning no candidate anywhere ever called `TryAcquireOrRenew` again for this lease. Meanwhile the Service's advertised VIP kept accepting TCP connections (ARP for the address was still answered by *something*) but returned EOF on any read. The client-visible symptom was `sqlcmd`/any TCP client connecting successfully and then getting an immediate EOF, and higher-level automation saw repeated `context deadline exceeded` on every DB round-trip during this window.

No `"no existing lease found for service..."` error (the signature of the unrelated, already-fixed #1664/`dropCancelledServiceContext` bug) appeared in the logs. This is a distinct failure mode: the lease and service context were both intact, but every live restart-loop goroutine for this service was starved out because of the accumulated duplicates racing on `svcLease.Lock()`/`Elected`, none of them ever winning back the actual `election.RunOrDie` re-acquisition path.

## Root cause

Missing "already started" guard in `pkg/endpoints/endpoints.go`'s `startServiceHandlingIfNeeded`, `EnableServicesElection` branch. This allows unbounded duplicate spawns of the per-service leader-election restart-loop goroutine, one per endpoint-churn event, instead of exactly one for the service's whole lifetime.

## Reproduction sketch

1. Enable `svc_election`/`EnableServicesElection` for a LoadBalancer service.
2. Cause the service's backing endpoint to flip zero → non-zero → zero → non-zero repeatedly in quick succession (e.g. rolling-restart the single backend pod a selector targets several times).
3. Observe (via `pprof`/goroutine dump, or by instrumenting `startServiceHandlingIfNeeded`) that N `startLeaderElection` goroutines are now running concurrently for the same service, where N grows with the number of endpoint transitions.
4. Eventually force a teardown (endpoints → zero) at a moment when several of these goroutines are mid-transition; observe the associated `Lease` object's `renewTime` stop advancing even after endpoints become healthy again.

## Suggested fix

Add a one-time-per-lifetime flag to `servicecontext.Context` (e.g. `LeaderElectionStarted bool`), set immediately before spawning the goroutine and checked before spawning again, mirroring the existing `IsWatched` field's usage pattern for the outer per-service watch goroutine, and analogous to the `Signalled`-guard already present on the ARP-only branch right below it. A PR implementing this is linked below.

## Environment

- kube-vip v1.2.2 (also reachable on v1.2.1, the affected code path is unchanged across both)
- `svc_election` / `EnableServicesElection: true`, `ExternalTrafficPolicy: Local`, selector-based Service fronting a StatefulSet whose "primary" label moves between pods during an Availability Group failover

erimerdal:
## What

`startServiceHandlingIfNeeded`'s `EnableServicesElection` branch spawned a new `startLeaderElection` goroutine on every `AddOrModify` call that saw non-zero endpoints, unlike the ARP-only branch right below it, which guards the equivalent spawn with `svcCtx.Signalled`. `startLeaderElection` already loops forever internally (until `svcCtx.Ctx` is cancelled), re-attempting election on its own whenever the service's `Lease` becomes free, so it only needs to be started once per service lifetime, not once per endpoint event.

Adds `LeaderElectionStarted` to `servicecontext.Context`: a one-time flag set right before the restart-loop goroutine is spawned and checked before spawning again, mirroring the existing `IsWatched` field's once-per-lifetime usage pattern elsewhere in this same struct.

## Why

`AddOrModify` runs on every EndpointSlice add/modify/resync event for a service, including transitions back from zero endpoints (e.g. a backend pod's readiness or role flipping during a rolling upgrade). Without this guard, each such event spawns another duplicate, permanently-running restart-loop goroutine for the same service, all contending on the same shared `Lease`.

We hit this in production: after a legitimate zero-endpoint teardown (VIP correctly deleted, leadership correctly released via `OnStoppedLeading`) followed by the endpoint becoming healthy again, the service's `Lease` never reacquired a holder: `kubectl get lease` showed `holderIdentity: ""` with `acquireTime == renewTime` frozen at the teardown timestamp for 11+ minutes, while the VIP kept accepting TCP connections but returned EOF on read (no active leader serving traffic). Full logs, the `Lease` YAML, and the RCA are in the linked issue.

Distinguished from the superficially similar #1664 (`dropCancelledServiceContext`, stale-`svcCtx` desync): that bug's distinctive symptom, `"no existing lease found for service..."`, was never observed in our logs. This is a different root cause in the same subsystem.

## Testing

Adds `TestAddOrModify_ServicesElectionStartsLeaderElectionOnce` in `pkg/endpoints/endpoints_test.go`, which drives `AddOrModify` three times in a row with non-zero endpoints and asserts the leader-election attempt only starts once. Verified (via a Linux container, since this package needs `GOOS=linux`) that this test:
- **fails** against the pre-fix code, a second restart-loop goroutine starts and is observed
- **passes** with the fix applied, including under `-race`

Built an image with the fix, the EOF issue is no longer observed.

Fixes #1665

### DIFF
--- a/pkg/servicecontext/servicecontext.go
+++ b/pkg/servicecontext/servicecontext.go
@@ -15,6 +15,12 @@ type Context struct {
 	epReady            sync.Once
 	Signalled          atomic.Bool
 	LeaderCancel       context.CancelFunc
+	// LeaderElectionStarted tracks whether the per-service leader-election
+	// restart-loop goroutine has already been spawned for this context's
+	// lifetime. That goroutine runs until Ctx is cancelled, re-attempting
+	// election forever on its own, so it must be started at most once per
+	// service - not once per AddOrModify call with non-zero endpoints.
+	LeaderElectionStarted bool
 }
 
 func New(ctx context.Context) *Context {
--- a/pkg/endpoints/endpoints.go
+++ b/pkg/endpoints/endpoints.go
@@ -247,10 +247,21 @@ func (p *Processor) updateAnnotations(service *v1.Service, lastKnownGoodEndpoint
 func (p *Processor) startServiceHandlingIfNeeded(svcCtx *servicecontext.Context, service *v1.Service,
 	serviceFunc func(*servicecontext.Context, *v1.Service, *sync.WaitGroup, bool) error, wg *sync.WaitGroup, les *atomic.Int64) error {
 	if p.config.EnableServicesElection {
-		wg.Go(func() {
-			les.Add(1)
-			p.startLeaderElection(svcCtx, service, serviceFunc, wg)
-		})
+		// startLeaderElection runs a forever loop (until svcCtx.Ctx is
+		// cancelled) that re-attempts election on its own whenever the lease
+		// becomes free, so it must only ever be spawned once per service.
+		// AddOrModify is invoked repeatedly for the same service (every
+		// endpoint-slice add/modify/resync event, including transitions back
+		// from zero endpoints), so without this guard a duplicate restart-loop
+		// goroutine gets spawned on every such event, piling up concurrent,
+		// redundant loops racing on the same shared Lease.
+		if !svcCtx.LeaderElectionStarted {
+			svcCtx.LeaderElectionStarted = true
+			wg.Go(func() {
+				les.Add(1)
+				p.startLeaderElection(svcCtx, service, serviceFunc, wg)
+			})
+		}
 		return nil
 	}
 
--- a/pkg/endpoints/endpoints_test.go
+++ b/pkg/endpoints/endpoints_test.go
@@ -3,10 +3,13 @@ package endpoints
 import (
 	"context"
 	"sync"
+	"sync/atomic"
 	"testing"
+	"time"
 
 	"github.com/kube-vip/kube-vip/pkg/endpoints/providers"
 	"github.com/kube-vip/kube-vip/pkg/kubevip"
+	"github.com/kube-vip/kube-vip/pkg/lease"
 	"github.com/kube-vip/kube-vip/pkg/servicecontext"
 	v1 "k8s.io/api/core/v1"
 	discoveryv1 "k8s.io/api/discovery/v1"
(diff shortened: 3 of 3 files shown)


pkg/servicecontext/servicecontext.go
```diff
@@ -15,6 +15,12 @@ type Context struct {
 	epReady            sync.Once
 	Signalled          atomic.Bool
 	LeaderCancel       context.CancelFunc
+	// LeaderElectionStarted tracks whether the per-service leader-election
+	// restart-loop goroutine has already been spawned for this context's
+	// lifetime. That goroutine runs until Ctx is cancelled, re-attempting
+	// election forever on its own, so it must be started at most once per
+	// service - not once per AddOrModify call with non-zero endpoints.
+	LeaderElectionStarted bool
```
p-strusiewiczsurmacki-mobica, line 23: Can `Elected` variable from `lease` be used for this, maybe? If new variable is introduced, we need to be sure it is cleared in proper places.

erimerdal: The guard we need is on service not on the lease, so unfortunately no.
This is a new variable but only in the service's context, lifetime is bound to context. 

Is this clearance documentation related or in the code?

Cellebyte: @erimerdal any objections on #1668?

erimerdal: No its great. Thank you!

Cellebyte:
Closed through #1668

PR #1667, opened 2026-08-05
Fixed IP updater for DNS

### LINKED ISSUE #1655: v1.2.2: DNS-mode control-plane VIP vanishes ~60s after leader election; startup goroutine deadlock leaves the address lifetime unrefreshed
## Describe the bug

In control-plane ARP mode with the VIP configured as a DNS name
(`address` set to a hostname rather than an IP), v1.2.2 can deadlock
during post-election startup on slower hardware. The observable symptom:
after winning leader election the VIP comes up, works for about a
minute, then disappears and never returns. The process stays alive and
keeps renewing the leader lease, so no failover happens either - the
worst combination: the node holds leadership but cannot serve the VIP.

Not reproducible on v1.2.1 with identical configuration. On v1.2.2 it
reproduces consistently on arm64 (Raspberry Pi class) control planes,
while the identical configuration on faster amd64 nodes starts up
cleanly - consistent with a timing-dependent race in the startup path.

Observed with the explicit `k8s_config_file`/`kubernetes_addr`
workaround from #1654 applied, so it is independent of that issue.

## Why the VIP disappears after ~60 seconds

DNS-mode addresses are added with `ValidLft = defaultValidLft` (60s)
(`pkg/vip/address.go`), and the DNS updater loop re-asserts the address
every 3 seconds (`pkg/vip/dns.go`, `AddIP(true, false,
defaultInterval)`), so the kernel expires the VIP one lifetime after the
updater stops running. That is exactly what happens: the updater's first
iteration(s) add the address, the process then deadlocks, and the kernel
removes the address 60 seconds later. Nothing re-adds it.

## Where it deadlocks

Log of the elected leader freezes at startup and never advances:

```
INFO Using Kubernetes configuration from explicit file path=/etc/kubernetes/admin.conf address=kubernetes:6443
INFO starting Kube-vip Manager mode=ARP
INFO Start ARP/NDP advertisement Global
INFO [ARP manager] starting ARP/NDP advertisement
INFO cluster membership namespace=kube-system lock=plndr-cp-lock id=<node>
I... leaderelection.go:272] "Successfully acquired lease" lock="kube-system/plndr-cp-lock"
INFO New leader leader=<node>
INFO starting the DNS updater address=<vip-dns-name>
<nothing further, ever>
```

On a healthy (amd64) leader the very next lines are the layer-2
broadcaster start and `[ARP manager] inserting ARP/NDP instance`. Those
come from the `layer2Update` goroutine, which `StartCluster`
(`pkg/cluster/service.go`) only spawns after the synchronous
`network.AddIP(false, false)` call has returned. On the deadlocked
node neither line ever appears and no error is logged, so the main flow
is blocked inside `AddIP` - which starts by taking the shared
`link.Lock`. The DNS updater goroutine (which funnels through the same
lock via its own `AddIP`, after having successfully added the address at
least once) stops producing output at the same moment: no `cannot
lookup` warnings, no errors, for the entire process lifetime.

Process state confirms a userspace deadlock rather than a stuck kernel
operation:

- Every thread of the process is parked in `futex_wait` (kernel stacks
  read from `/proc/<pid>/task/*/stack`); no thread is in a netlink,
  socket, or wait syscall.
- No child processes.
- Leader-election goroutines keep running (lease renewTime advances),
  so the process is only partially wedged.

Environmental causes were ruled out on the affected host: DNS resolution
of the VIP name is instant (~40ms via `getent`), and nothing on the LAN
answers ARP for the VIP (no duplicate-address conflict).

v1.2.1 has the same DNS/ValidLft refresh design but does not exhibit
this, which points at the extensively rewritten `pkg/vip/address.go` /
`pkg/arp/arp.go` (and the new interface-watching added for
leadership-loss-on-interface-down) introducing a lock ordering that can
deadlock when the DNS updater's first iterations interleave with the
main flow's `AddIP` during startup. I could not capture a goroutine dump
to pin the exact cycle: the distroless image exposes no pprof endpoint
on the Prometheus port and SIGQUIT would have killed the instance.

## To Reproduce

1. kubeadm HA control plane, kube-vip v1.2.2 static pods, ARP mode,
   `cp_enable=true`, leader election enabled, `address` set to a DNS
   name for the VIP.
2. Slow control-plane hardware (reproduces consistently on Raspberry Pi
   class arm64 nodes; fast amd64 nodes win the race and start cleanly).
3. Restart the kube-vip static pods and let a node win the election.
4. The VIP appears, then vanishes about 60 seconds later; the leader's
   log is frozen as above; the lease keeps renewing so no other member
   takes over.

## Expected behavior

The elected leader either serves the VIP (DNS updater keeps refreshing
the 60s lifetime) or crashes/loses leadership so another member can take
over. It must not keep the lease while permanently unable to hold the
VIP.

## Suggested angles

- Audit lock ordering between `StartCluster`'s synchronous `AddIP`, the
  DNS updater goroutine, and the ARP manager / interface-watch code
  added in the v1.2.2 rewrite of `pkg/vip/address.go` and
  `pkg/arp/arp.go`.
- Consider exposing pprof on the metrics listener to make field
  diagnosis of this class of bug possible.
- Consider tying lease renewal to VIP health: a leader whose VIP has
  expired should surrender the lease.

## Environment

- kube-vip v1.2.2 (not reproducible on v1.2.1), image
  ghcr.io/kube-vip/kube-vip, ARP mode, `cp_enable=true`, `svc_enable`
  unset, leader election enabled, DNS-name VIP
- Kubernetes v1.36.2, kubeadm, containerd 2.3.3
- Reproduces on arm64 control planes; does not reproduce on faster
  amd64 with identical configuration

p-strusiewiczsurmacki-mobica:
This PR reworks the `ipUpdater` used with DNS, which, according to the reports fixes #1655

The change is to move two of the locking functions outside of the `for` loop, and using a ticker instead of a sleep.

### DIFF
--- a/pkg/vip/dns.go
+++ b/pkg/vip/dns.go
@@ -31,21 +31,25 @@ func NewIPUpdater(vip Network) IPUpdater {
 
 // Run runs the IP updater
 func (d *ipUpdater) Run(ctx context.Context) {
+	mode := utils.IPv4Family
+	if utils.IsIPv6(d.vip.IP()) {
+		mode = utils.IPv6Family
+	}
+
+	dnsName := d.vip.DNSName()
+
+	t := time.NewTicker(defaultInterval * time.Second)
+	defer t.Stop()
 
 	for {
 		select {
 		case <-ctx.Done():
 			log.Info("stop ipUpdater")
 			return
-		default:
-			mode := "ipv4"
-			if utils.IsIPv6(d.vip.IP()) {
-				mode = "ipv6"
-			}
-
-			ip, err := utils.LookupHost(d.vip.DNSName(), mode, true)
+		case <-t.C:
+			ip, err := utils.LookupHost(dnsName, mode, true)
 			if err != nil {
-				log.Warn("cannot lookup", "name", d.vip.DNSName(), "err", err)
+				log.Warn("cannot lookup", "name", dnsName, "err", err)
 				// fallback to renewing the existing IP
 				ip = []string{d.vip.IP()}
 			}
@@ -58,8 +62,6 @@ func (d *ipUpdater) Run(ctx context.Context) {
 			if _, err := d.vip.AddIP(true, false, defaultInterval); err != nil {
 				log.Error("error adding virtual IP", "err", err)
 			}
-
 		}
-		time.Sleep(defaultInterval * time.Second)
 	}
 }
--- a/pkg/loadbalancer/ipvs.go
+++ b/pkg/loadbalancer/ipvs.go
@@ -58,8 +58,9 @@ type IPVSLoadBalancer struct {
 
 func NewIPVSLB(ctx context.Context, network vip.Network, port uint16, forwardingMethod string, backendHealthCheckInterval int,
 	nftables bool, killFunc func(), wg *sync.WaitGroup) (*IPVSLoadBalancer, error) {
+	log.Info("Starting IPVS LoadBalancer", "network", network)
+
 	address := network.IP()
-	log.Info("Starting IPVS LoadBalancer", "address", network)
 
 	// Create IPVS client
 	c, err := ipvs.New()
--- a/pkg/utils/dns.go
+++ b/pkg/utils/dns.go
@@ -25,7 +25,7 @@ func LookupHost(dnsName, dnsMode string, requireDualStack bool) ([]string, error
 		return nil, errors.Errorf("empty address for %s", dnsName)
 	}
 	addrs := []string{}
-	switch dnsMode {
+	switch strings.ToLower(dnsMode) {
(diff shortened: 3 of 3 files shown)


pkg/vip/dns.go
```diff
@@ -31,21 +31,25 @@ func NewIPUpdater(vip Network) IPUpdater {
 
 // Run runs the IP updater
 func (d *ipUpdater) Run(ctx context.Context) {
+	mode := "ipv4"
```
Cellebyte: Do we haben constants for these somewhere?

p-strusiewiczsurmacki-mobica: Changed that :)

Cellebyte (approved):
LGTM

PR #1668, opened 2026-08-05
fix(services): start the leader-election loop once per service

### LINKED ISSUE #1665: Duplicate leader-election restart-loop goroutines under EnableServicesElection cause a service Lease to permanently lose its holder

MaxRink:
> [!NOTE]
> **Stacked on #1669.** This branch contains #1669's single commit as its base, so the diff here shows both until that merges. Review #1669 first; this PR's own changes start at `fix(services): start the leader-election loop once per service`.
>
> #1669 is a prerequisite: it fixes a lease cleanup bug on the service teardown path that this PR's fault tests exercise.

Fixes #1665. Alternative to #1666.

## Problem

`startLeaderElection` loops until the service context is cancelled, re-attempting election whenever the lease is free. It only needs to be started once per service lifetime. `startServiceHandlingIfNeeded` started it on every `AddOrModify` call, and `AddOrModify` runs on every EndpointSlice event, so endpoint churn piled up duplicate permanent loops for the same service, all contending on the same lease.

The ARP branch right below already guards its spawn with `svcCtx.Signalled`.

## Change

Guard the spawn with a `sync.Once` on the service context. #1666 review asked whether a new flag would be cleared in the right places: a `Once` has nothing to clear, and the loop is bound to the service context, which is replaced whenever the service is recreated.

## Making the bug observable

The duplicate loops were not observable from outside the process, so there was nothing an e2e test could assert on. `startLeaderElection` carried a `les *atomic.Int64` that was incremented and never read. That is replaced with real instrumentation:

- `kube_vip_service_election_loops{namespace,name}` gauge, tracked for the lifetime of the loop goroutine. More than 1 per service means loops leaked.
- `kube_vip_service_election_attempts_total{namespace,name}`, so a wedged restart loop shows up as a counter that stops advancing.
- `kube_vip_service_election_errors_total{namespace,name,reason}`, with `reason="no_lease"` for the svcMap/leaseMgr desync.

## Fault tests

`-electionFaults` in the service e2e suite injects a fault per failure mode in this subsystem and asserts a signal that is actually broken when that bug is present. Every fault is then followed by the feature contract: at most one election loop, a stable error counter, a held lease, and a VIP that serves traffic.

| fault | assertion | targets |
|---|---|---|
| backend scaled 1->0->1 five times | election loops settle at <= 1 per node | #1665 |
| zero endpoints under a local policy | the address stops answering rather than black-holing | |
| EndpointSlice deleted, then Service annotated | `reason="no_lease"` counter stops growing | #1663 / #1664 |
| lease deleted, then holderIdentity blanked | convergence | |
| apiserver blocked from the leader past the lease duration | election attempts advance afterwards | #1650 |
| 15 rapid Service annotation patches | spawn-once driven from the service watch | #1665 |
| externalTrafficPolicy flipped both ways | loops settle at <= 1, held lease, served VIP | #1669 |
| one of two services sharing a lease deleted | the lease stays held for the one that stays | #1669 |
| apiserver blocked from a follower | lease does not move, traffic continues, no loop leaked | |

### Verified both directions

Every assertion was run against the fixed branch (must pass) and against a branch with only the spawn guard reverted and the metrics kept (must fail).

Unfixed:
```
a node still runs 11 leader election loops for service "kube-vip-service" after endpoint
flapping, want at most 1, loops leaked (map[services-worker:11 services-worker2:0])
Testing Complete [10] passed / [1] failed
```

Fixed: `Testing Complete [11] passed / [0] failed`, loops holding at 1 through every fault and `election attempts advanced 5 -> 6` after the leader partition.

`TestAddOrModify_ServicesElectionStartsOnce` covers the same invariant at unit level and asserts the gauge alongside the call count so both layers agree. It reports `started 3 times, want 1` against unpatched main.

### Assertions that needed correcting

Three first attempts were wrong and each was caught by running the fixed branch, not the unfixed one:

- asserting zero election errors: a single `no_lease` happens benignly during service setup, so this failed on correct code. The #1664 signature is a counter that keeps climbing, so it asserts delta-stability now.
- asserting election progress after deleting the lease object: the election client just recreates the lease on the next renew without losing leadership, so no new attempt follows. Moved to the apiserver partition, which is the fault that actually drives `OnStoppedLeading`.
- asserting the loop count from a single scrape: a fault that rebuilds the service context legitimately has the old and new loop alive at the same instant. It polls until the count settles; a leaked loop only exits with its service context, so it never settles.

## Related

The `externalTrafficPolicy` teardown path had two lease-lifecycle defects, both fixed in #1669 underneath this PR and asserted here as one of the faults.

Also worth flagging from reading the shared-lease path (`kube-vip.io/leaseName`), untested and not touched here: the secondary service's goroutine parks on `svcLease.Ctx.Done()` and leaks if only the secondary is deleted, `svcLease.Started` is reassigned without holding the mutex, and `leaseMgr.Get` can return nil at `endpoints.go` between the context check and `l.Lock()`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -150,7 +150,7 @@ e2e-tests: e2e-tests-arp e2e-tests-rt e2e-tests-bgp
 
 service-tests:
 	$(MAKE) -C testing/e2e/e2e dockerLocal
-	E2E_IMAGE_PATH=$(REPOSITORY)/$(TARGET):$(DOCKERTAG) go run ./testing/services -Services -simple -deployments -leaderActive -leaderFailover -localDeploy -endpointFlap -egress -egressIPv6 -dualStack -egressInternal
+	E2E_IMAGE_PATH=$(REPOSITORY)/$(TARGET):$(DOCKERTAG) go run ./testing/services -Services -simple -deployments -leaderActive -leaderFailover -localDeploy -electionFaults -egress -egressIPv6 -dualStack -egressInternal
 
 trivy: dockerx86ActionIPTables
 	docker run -v /var/run/docker.sock:/var/run/docker.sock aquasec/trivy:0.47.0 \
--- a/Makefile
+++ b/Makefile
@@ -150,7 +150,7 @@ e2e-tests: e2e-tests-arp e2e-tests-rt e2e-tests-bgp
 
 service-tests:
 	$(MAKE) -C testing/e2e/e2e dockerLocal
-	E2E_IMAGE_PATH=$(REPOSITORY)/$(TARGET):$(DOCKERTAG) go run ./testing/services -Services -simple -deployments -leaderActive -leaderFailover -localDeploy -egress -egressIPv6 -dualStack -egressInternal
+	E2E_IMAGE_PATH=$(REPOSITORY)/$(TARGET):$(DOCKERTAG) go run ./testing/services -Services -simple -deployments -leaderActive -leaderFailover -localDeploy -endpointFlap -egress -egressIPv6 -dualStack -egressInternal
 
 trivy: dockerx86ActionIPTables
 	docker run -v /var/run/docker.sock:/var/run/docker.sock aquasec/trivy:0.47.0 \
--- a/go.mod
+++ b/go.mod
@@ -87,6 +87,7 @@ require (
 	github.com/josharian/native v1.1.0 // indirect
 	github.com/json-iterator/go v1.1.12 // indirect
 	github.com/k-sone/critbitgo v1.4.0 // indirect
+	github.com/kylelemons/godebug v1.1.0 // indirect
 	github.com/mailru/easyjson v0.9.0 // indirect
 	github.com/mattn/go-isatty v0.0.20 // indirect
 	github.com/mdlayher/genetlink v1.3.2 // indirect
--- a/pkg/cluster/clusterLeaderElection.go
+++ b/pkg/cluster/clusterLeaderElection.go
@@ -37,7 +37,7 @@ func (cluster *Cluster) StartCluster(ctx context.Context, c *kubevip.Config,
 	// Without this, RunOrDie would continue running until leadership is naturally lost.
 	wg.Go(func() {
 		<-objLease.Ctx.Done()
-		leaseMgr.Delete(leaseID, objectName)
+		leaseMgr.Delete(leaseID, objectName, objLease)
 	})
 
 	if !isNew {
--- a/pkg/endpoints/endpoints.go
+++ b/pkg/endpoints/endpoints.go
@@ -6,7 +6,6 @@ import (
 	"net"
 	"strings"
 	"sync"
-	"sync/atomic"
 	"time"
 
 	log "log/slog"
@@ -74,8 +73,6 @@ func (p *Processor) AddOrModify(svcCtx *servicecontext.Context, event watch.Even
 	// If we have a local endpoint then begin the leader Election, unless it's already running
 	//
 
-	les := atomic.Int64{}
-
 	// Check that we have local endpoints
 	if len(endpoints) != 0 {
 		// Ignore IPv4
@@ -85,7 +82,7 @@ func (p *Processor) AddOrModify(svcCtx *servicecontext.Context, event watch.Even
 
 		p.updateLastKnownGoodEndpoint(lastKnownGoodEndpoint, endpoints, service)
 
-		if err := p.startServiceHandlingIfNeeded(svcCtx, service, serviceFunc, wg, &les); err != nil {
+		if err := p.startServiceHandlingIfNeeded(svcCtx, service, serviceFunc, wg); err != nil {
 			return true, err
 		}
 
@@ -103,7 +100,7 @@ func (p *Processor) AddOrModify(svcCtx *servicecontext.Context, event watch.Even
 	} else {
 		if allowReconcileWithoutEndpoints {
 			// Explicit opt-in for controllers that create LoadBalancer services without endpoints
-			if err := p.startServiceHandlingIfNeeded(svcCtx, service, serviceFunc, wg, &les); err != nil {
+			if err := p.startServiceHandlingIfNeeded(svcCtx, service, serviceFunc, wg); err != nil {
 				return true, err
 			}
 			svcCtx.SignalReadiness()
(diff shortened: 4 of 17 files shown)


erimerdal (approved):
thanks! covers https://github.com/kube-vip/kube-vip/issues/1665

PR #1669, opened 2026-08-06
fix(lease): do not let a stale cleanup cancel a recreated lease

MaxRink:
Two defects on the same path: a service that is torn down and immediately rebuilt ends up with a lease nobody holds, and is never handled again.

Reachable from an ordinary service update. Flipping `externalTrafficPolicy` makes `serviceChanged` cancel the service context and rebuild it, which reproduced both.

## 1. A late cleanup cancels the replacement

Every object that starts leader election also starts a goroutine that calls `Manager.Delete` once its context is cancelled. `Delete` looked the lease up by name only, so it acted on whatever held that name **when it ran**, not the instance the caller was given. By then the lease may already have been replaced, and the cleanup cancels the live replacement.

Fixed by passing the lease the caller owns and ignoring a stale caller. All four call sites already hold it. Passing `nil` keeps the old behaviour of acting on whichever lease currently holds the name, which is what the existing tests assert.

## 2. The replacement *is* the doomed instance

The instance guard above cannot help when both are the same object. Teardown cancels the service context and leaves the lease registered, because the cleanup that removes it is deferred. The rebuild calls `Add`, finds that lease still in the map, and gets it straight back — so the new service context is parented to a lease the pending cleanup is about to cancel. Observed as a service that cycles acquire / lose / re-acquire every few seconds and never settles:

```
new leader election    service=kube-vip-service lock_name=kubevip-kube-vip-service
leadership lost        service=kube-vip-service
stopping leader election
new leader election    ... repeating, 11 times in 2 minutes
```

Fixed by adding `Manager.Retire` and calling it from the `serviceChanged` teardown next to the `svcMap` purge, so the lease is out of the map before the replacement context is built. `Add` also refuses to hand out a lease whose context is already cancelled, closing the same hazard for any other path that cancels a lease directly.

## Review follow-up

Patryk spotted that the first cut cancelled the lease context outright, so with a common lease a change to one service would tear down every sibling sharing it. `Retire` was only needed to get the lease out of the map before the rebuild, and `Delete` already does that once the last object leaves — so `Retire` is gone, the teardown calls `Delete` with its own object name, and the manager API stays `Add`/`Delete`/`Get`.

## Tests

Three unit tests, each failing without its corresponding fix:

- `TestManager_Delete_DoesNotCancelRecreatedLease` — `cleanup for the torn down lease cancelled the recreated lease`
- `TestManager_Add_AfterCancelWithoutDelete_ReusesDoomedLease` — `replacement service context would be parented to the doomed lease`
- `TestManager_LeaseLifetimeInvariant` — pins the lifetime rule for the whole surface rather than one scenario: a lease stays usable for exactly as long as at least one object holds it, and a rebuild afterwards gets a fresh one. Table driven over 1, 2 and 4 objects; the 2 and 4 cases fail against the reviewed behaviour with `lease was cancelled with N object(s) still holding it`.

#1668 adds the cluster-level counterpart in its `-electionFaults` suite: two services share a lease, one is deleted, and the lease has to stay held for the one that stays.

`go build`, `go vet`, `gofmt`, `golangci-lint v2.12` and `go test -race ./pkg/... ./cmd/...` all clean.

Found while adding fault-injection coverage for the per-service leader election in #1668, which stacks on top of this.

### DIFF
--- a/pkg/services/processor.go
+++ b/pkg/services/processor.go
@@ -196,6 +196,10 @@ func (p *Processor) AddOrModify(ctx context.Context, event watch.Event, serviceF
 				}
 				// in theory this should never fail
 				p.svcMap.Delete(svc.UID)
+				// Retire the lease as well, so the replacement context below is not
+				// parented to a lease the pending cleanup is about to cancel.
+				ns, name := lease.ServiceName(svc)
+				p.leaseMgr.Retire(lease.NewID(p.config.LeaderElectionType, ns, name), nil)
 				// Reset the the svcCtx when it was garbage collected
 				// As the next function will create a new context when nil
 				svcCtx = nil
--- a/pkg/services/processor.go
+++ b/pkg/services/processor.go
@@ -196,10 +196,12 @@ func (p *Processor) AddOrModify(ctx context.Context, event watch.Event, serviceF
 				}
 				// in theory this should never fail
 				p.svcMap.Delete(svc.UID)
-				// Retire the lease as well, so the replacement context below is not
-				// parented to a lease the pending cleanup is about to cancel.
+				// Drop this service from its lease now, so the replacement context
+				// below is not parented to a lease the pending cleanup is about to
+				// cancel. A lease shared with other services stays alive for them.
 				ns, name := lease.ServiceName(svc)
-				p.leaseMgr.Retire(lease.NewID(p.config.LeaderElectionType, ns, name), nil)
+				leaseID := lease.NewID(p.config.LeaderElectionType, ns, name)
+				p.leaseMgr.Delete(leaseID, lease.ServiceNamespacedName(svc), nil)
 				// Reset the the svcCtx when it was garbage collected
 				// As the next function will create a new context when nil
 				svcCtx = nil
--- a/pkg/cluster/clusterLeaderElection.go
+++ b/pkg/cluster/clusterLeaderElection.go
@@ -37,7 +37,7 @@ func (cluster *Cluster) StartCluster(ctx context.Context, c *kubevip.Config,
 	// Without this, RunOrDie would continue running until leadership is naturally lost.
 	wg.Go(func() {
 		<-objLease.Ctx.Done()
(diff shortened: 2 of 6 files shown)


pkg/services/processor.go
```diff
@@ -196,6 +196,10 @@ func (p *Processor) AddOrModify(ctx context.Context, event watch.Event, serviceF
 				}
 				// in theory this should never fail
 				p.svcMap.Delete(svc.UID)
+				// Retire the lease as well, so the replacement context below is not
+				// parented to a lease the pending cleanup is about to cancel.
+				ns, name := lease.ServiceName(svc)
+				p.leaseMgr.Retire(lease.NewID(p.config.LeaderElectionType, ns, name), nil)
```
p-strusiewiczsurmacki-mobica: `leaseMgr.Retire()` internally calls `leaseMgr.retire()`, which calls `cancel()` on lease context. In case of common lease, this means that modification of one service, will also cancel all the others that share the same lease.

MaxRink: Correct, that was wrong.

`Retire` was only ever needed to get the lease out of the map before the rebuild, and `Delete` already does that once the last object leaves. So `Retire` is gone and the teardown now calls `Delete` with its own object name: siblings on a common lease keep it alive, and the manager API stays `Add`/`Delete`/`Get`.

To keep this from coming back I replaced the single-scenario test with `TestManager_LeaseLifetimeInvariant`, which pins the rule for the whole surface (a lease stays usable for exactly as long as at least one object holds it, and a rebuild afterwards gets a fresh one), table driven over 1, 2 and 4 objects. The 2 and 4 cases fail against the code you reviewed with `lease was cancelled with N object(s) still holding it`.

Also added a cluster level fault in #1668's `-electionFaults` suite: two services share a lease, one is deleted, and the lease has to stay held for the one that stays.

p-strusiewiczsurmacki-mobica (approved):
LGTM

PR #1671, opened 2026-08-06
routing table mode: honor the configured Kubernetes address for backend health checks

### LINKED ISSUE #1670: RT mode: health check dials Node addresses, fails permanently when serving certs don't cover them
# What happens

In the RT mode with cp_enable, the health check resolves the Node address (via `getNodeIPs`) and calls `https://node-ip:6443`.

In our deployment via static pod on OpenShift, apiserver does not cover node IP but only the IPv4 loopback listener -- so the check fails TLS verification and the VIP route is never installed

```
ERROR discover k8s version err="Get \"https://[fd2e:...::14]:6443/version?timeout=10s\":
tls: failed to verify certificate: x509: certificate is valid for 172.30.0.1, fd02::1, not fd2e:...::14"
```

It is kinda race situation because it works on cluster bootstrap (no Node object so we fallback to 127.0.0.1) but when we restart kube-vip on a healthy cluster, we go into the loop of failure posted above.

# What I expect

`kubernetes_addr` / `KubernetesAddr` extended via #1627 should also apply for the backend health check.

# Reproducer

RT-mode + cp_enable instance started while the Node object exists, on a cluster whose apiserver cert does not contain node IPs (e.g. OpenShift with static pod deployment). Observe the TLS loop and no route in the routing table.

# Fix proposal

If set, `KubernetesAddr` should be used as the backend health-check endpoint for every VIP family instead of the node addresses.

The proposal was tested on the PoC OpenShift cluster and will arrive via a PR.

mkowalski:
In the RT-mode with cp_enable, health check resolves the node address via getNodeIPs and calls https://node-ip:6443.

In deployments where apiserver does not have a certificate for the node IP this fails with "failed to verify certificate: x509: [...]"

When an explicit k8s address is configured via KubernetesAddr, we want to use it for the backend health check instead of the node addresses.

Fixes: #1670

### DIFF
--- a/pkg/backend/backend.go
+++ b/pkg/backend/backend.go
@@ -31,10 +31,10 @@ func (e *Entry) Check() bool {
 	// homeConfigPath := filepath.Join(os.Getenv("HOME"), ".kube", "config")
 
 	var k8sAddr string
-	if utils.IsIPv4(e.Addr) {
-		k8sAddr = fmt.Sprintf("%s:%v", e.Addr, e.Port)
-	} else {
+	if utils.IsIPv6(e.Addr) {
 		k8sAddr = fmt.Sprintf("[%s]:%v", e.Addr, e.Port)
+	} else {
+		k8sAddr = fmt.Sprintf("%s:%v", e.Addr, e.Port)
 	}
 
 	switch {
--- a/pkg/cluster/service.go
+++ b/pkg/cluster/service.go
@@ -6,6 +6,8 @@ import (
 	"fmt"
 	"io/fs"
 	"net/http"
+	"net/url"
+	"strconv"
 	"strings"
 	"sync"
 	"syscall"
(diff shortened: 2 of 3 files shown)


pkg/backend/backend.go
```diff
@@ -47,19 +31,13 @@ func (e *Entry) Check() bool {
 	// homeConfigPath := filepath.Join(os.Getenv("HOME"), ".kube", "config")
 
 	var k8sAddr string
-	if utils.IsIPv4(e.Addr) {
-		k8sAddr = fmt.Sprintf("%s:%v", e.Addr, e.Port)
-	} else {
+	if utils.IsIPv6(e.Addr) {
 		k8sAddr = fmt.Sprintf("[%s]:%v", e.Addr, e.Port)
+	} else {
+		k8sAddr = fmt.Sprintf("%s:%v", e.Addr, e.Port)
```
mkowalski, line 37: This change is "smuggled" here to handle hostnames too. Without it we would render hostname like IPv6 address, i.e. inside brackets, and that does not work.

p-strusiewiczsurmacki-mobica (approved):
LGTM

PR #1675, opened 2026-08-07
optional DAD skip for managed VIP addresses

mkowalski:
This PR adds an explicit configuration option `vip_skipdad`. This allows for a proper behaviour in IPv6 networks with RT mode and ECMP. Without this option only one node can ever hold the same IPv6 VIP what defeats the purpose of ECMP.

While the ARP mode currently has a recovery mode for failed DAD, RT mode misses it. It is also more feasible to use config option, as recovery mode in RT mode would always introduce a ~5s blackholing (we know there will be a conflict, so first attempt to add a VIP will fail and only the next one will force IFA_F_NODAD).

For simplicitly we ignore IPv4/IPv6 checks, as kernel transparently strips IFA_F_NODAD from v4 addresses without throwing errors.

For systems with IPv4 duplicate detection (called ACD and used e.g. in RHEL10's NetworkManager) this code has no effect as it's only userspace action and addresses we add via netlink bypass any check.

First discovered on a dual-stack cluster with two routing-table-mode advertisers: the ingress VIP answered or timed out depending on which ECMP path the flow hashed to; with NODAD both advertisers accept traffic.


pkg/vip/address_test.go
```diff
@@ -0,0 +1,28 @@
...
+
+func TestShouldSkipDAD(t *testing.T) {
+	cases := []struct {
+		name     string
+		dadSkip  bool
+		override bool
+		want     bool
+	}{
+		{name: "default: DAD runs", dadSkip: false, override: false, want: false},
+		{name: "network policy skips DAD", dadSkip: true, override: false, want: true},
```
p-strusiewiczsurmacki-mobica: I feel like using `network policy` in the name of this test is a bit misleading, as `Network Policy` is a separate k8s resource we do not deal with. Maybe `DAD skip enabled` or something similar?

pkg/vip/address.go
```diff
@@ -447,10 +462,13 @@ func (configurator *network) AddIP(precheck bool, skipDAD bool, minLifetime ...i
 	// For IPv6 addresses, optionally set NODAD flag to skip Duplicate Address Detection (DAD)
 	// This prevents DADFAILED loops when recovering from a previous DADFAILED state
 	// The flag tells the kernel to skip DAD, which is safe when we're re-adding
-	// an address that we know should be ours (e.g., after DADFAILED recovery)
-	if skipDAD && utils.IsIPv6(configurator.address.IP.String()) {
+	// an address that we know should be ours (e.g., after DADFAILED recovery).
+	// We also allow to globally configure NODAD in case user knows they are running in an
+	// environment where multiple nodes may advertise the same VIP (e.g., ECMP routing).
+	// As kernel strips IFA_F_NODAD from IPv4 addresses, this is only effective for IPv6 addresses.
+	if configurator.shouldSkipDAD(skipDAD) {
```
p-strusiewiczsurmacki-mobica: Without docs on this I'd still rather keep IPv6 check.

mkowalski: No problem, it's cheap, I'll fix

pkg/vip/address_test.go
mkowalski: No problem, naming is always the biggest issue

mkowalski: Done

pkg/vip/address.go
mkowalski: Done

pkg/vip/address_test.go
p-strusiewiczsurmacki-mobica: > naming is always the biggest issue

That's true :sweat_smile:

p-strusiewiczsurmacki-mobica (approved):
LGTM

PR #1681, opened 2026-08-11
Fix/dns lookup

### LINKED ISSUE #1680: v1.2.3: ipUpdater DNS lookups always fail with "no addresses found" (case-mismatched IP family comparison)
**Describe the bug**

After upgrading control-plane kube-vip from v1.2.1 to v1.2.3 (ARP mode, VIP specified as a DNS name, `dns_mode: first`), the elected leader logs `WARN cannot lookup ... err="no addresses found"` every 3 seconds, indefinitely, even though the name resolves fine from the same node.

The periodic lookup in the reworked ipUpdater (#1667) can never succeed because of a case mismatch in the IP family filter:

- [`pkg/vip/dns.go` L34-L37](https://github.com/kube-vip/kube-vip/blob/v1.2.3/pkg/vip/dns.go#L34-L37) sets the lookup family to `utils.IPv4Family` / `utils.IPv6Family`, whose values are the mixed-case strings `"IPv4"` / `"IPv6"`, and [L50](https://github.com/kube-vip/kube-vip/blob/v1.2.3/pkg/vip/dns.go#L50) passes that to `utils.LookupHost(dnsName, mode, true)`.
- [`pkg/utils/dns.go` L28-L30](https://github.com/kube-vip/kube-vip/blob/v1.2.3/pkg/utils/dns.go#L28-L30): `LookupHost` lowercases the value to match the `switch` case, but forwards the original mixed-case string to `getIPbyFamily`.
- [`pkg/utils/dns.go` L45](https://github.com/kube-vip/kube-vip/blob/v1.2.3/pkg/utils/dns.go#L45) and [L49](https://github.com/kube-vip/kube-vip/blob/v1.2.3/pkg/utils/dns.go#L49): `getIPbyFamily` compares `family == strings.ToLower(IPv4Family)`, i.e. `"IPv4" == "ipv4"`, which is always false (same for IPv6). No family checker is registered, the address list stays empty, and [L67-L68](https://github.com/kube-vip/kube-vip/blob/v1.2.3/pkg/utils/dns.go#L67-L68) returns `no addresses found`.

So every ticker iteration fails unconditionally, regardless of what DNS returns. The initial address fetch at startup/election still works because it goes through [`pkg/vip/address.go` L187](https://github.com/kube-vip/kube-vip/blob/v1.2.3/pkg/vip/address.go#L187) with the configured `dns_mode` (`first` in my case), which takes `LookupHost`'s `default` branch and skips the family filter entirely; that is why the VIP still comes up with the correct address.

Impact is limited thanks to the fallback introduced in #1667 (`ip = []string{d.vip.IP()}`): the existing VIP keeps being renewed, so the address does not expire (that part of the #1655 fix works, confirmed on the cluster that used to hit #1655). What is broken: runtime re-resolution of the VIP DNS name (a changed record is never picked up while a leader holds the VIP), and the leader logs the WARN every 3 seconds forever.

**To Reproduce**

1. Run kube-vip v1.2.3 control-plane with the VIP specified as a DNS name. The failure is unconditional, not timing- or resolver-dependent.
2. Let a leader be elected.
3. Observe `cannot lookup ... err="no addresses found"` every 3s in the leader log, even though the name resolves from the same node.

Or reproduce it directly in a test: `utils.LookupHost(name, utils.IPv4Family, true)` (the exact call `ipUpdater.Run` makes) returns `no addresses found` for any resolvable name.

**Expected behavior**

The periodic lookup succeeds and tracks DNS changes of the VIP name, as in v1.2.1.

Leader log after the v1.2.3 upgrade:

```
2026/08/10 18:05:21 INFO starting the DNS updater address=<redacted>
2026/08/10 18:05:21 INFO layer 2 broadcaster starting IP=<redacted> device=eth0
2026/08/10 18:05:21 INFO [ARP manager] inserting ARP/NDP instance name=<redacted>/32-eth0
2026/08/10 18:05:24 WARN cannot lookup name=<redacted> err="no addresses found"
2026/08/10 18:05:27 WARN cannot lookup name=<redacted> err="no addresses found"
2026/08/10 18:05:30 WARN cannot lookup name=<redacted> err="no addresses found"
(repeats every 3s indefinitely)
```

- OS/Distro: Ubuntu 24.04.4 LTS (Raspberry Pi, arm64)
- Kubernetes Version: v1.36.3
- Kube-vip Version: v1.2.3

**`Kube-vip.yaml`:**

Same manifest as I posted in https://github.com/kube-vip/kube-vip/issues/1655#issuecomment-5182615031 (image bumped to v1.2.3, debug env removed).

**Additional context**

Found while validating the v1.2.3 release on the same cluster that reliably hit #1655; the deadlock itself is gone, thanks for the quick fixes of #1654 and #1655. Suggested fix: compare case-insensitively in `getIPbyFamily` (`strings.EqualFold`), or normalize the family string once at the top of `LookupHost`, plus a small regression test asserting that `LookupHost(name, utils.IPv4Family, true)` returns an address for a resolvable name.

Cellebyte:
As mentioned in #1680, 
we introduced again a regression with version v1.2.3.

This should fix the mentioned issue of passing the dnsMode as it was to the underlying function.

closes #1680


p-strusiewiczsurmacki-mobica (approved):
LGTM

thebsdbox (approved):
LGTM

PR #1683, opened 2026-08-11
Fallback to regular BGP if MP-BGP cannot be configured

### LINKED ISSUE #1682: BGP mode crashes on IPv4-only clusters when MP-BGP is enabled (fails to find non-link-local IPv6 address)
**Describe the bug**

When kube-vip is configured with BGP mode and `mpbgp_nexthop=auto_sourceip` on a **single-stack IPv4 cluster** (no IPv6 addresses assigned to nodes), the BGP server fails to start entirely. kube-vip exits immediately with:

```
ERROR start manager err="failed to configure BGP mode: starting BGP server: failed to get MP-BGP addresses: failed to get non link-local IPv6 address: failed to find non-local IP on interface: eth0"
```

This means the **control plane VIP is never advertised**, making the cluster unreachable via BGP. The failure is fatal — kube-vip does not fall back to IPv4-only BGP.

**To Reproduce**

1. Deploy kube-vip `v1.2.1` in BGP mode on a single-stack IPv4 cluster (no IPv6 addresses on nodes).
2. Set the following env vars in the DaemonSet:
   ```yaml
   - name: bgp_enable
     value: "true"
   - name: mpbgp_ipv4
     valueFrom:
       fieldRef:
         fieldPath: status.hostIP
   - name: mpbgp_nexthop
     value: fixed
   - name: bgp_peers
     value: "<peer-ip>:<peer-asn>::false/mpbgp_nexthop=auto_sourceip"
   ```
3. Observe that kube-vip starts, logs `invalid IP` for all BGP peer fields, then crashes immediately.

**Expected behavior**

When `mpbgp_ipv4` is set but the node has no non-link-local IPv6 address, kube-vip should either:
- Skip MP-BGP IPv6 address resolution and operate in IPv4-only MP-BGP mode, **or**
- Log a clear warning that IPv6 MP-BGP is unavailable and continue with IPv4-only BGP.

It must **not** crash the entire BGP stack because IPv6 is absent on an IPv4-only node.

**Environment**

- OS/Distro: Linux (Debian-based)
- Kubernetes Version: v1.32.x
- Kube-vip Version: v1.2.1 (build `4708b07343dd9209f9b56cfa52dc53a85d8d3166`)
- Cluster networking: IPv4 single-stack (no IPv6 addresses on the primary interface)

**`Kube-vip` relevant config (sanitized)**

```yaml
env:
  - name: bgp_enable
    value: "true"
  - name: bgp_hold_time
    value: "3"
  - name: bgp_keepalive_interval
    value: "1"
  - name: bgp_sourceip
    valueFrom:
      fieldRef:
        fieldPath: status.hostIP
  - name: bgp_routerid
    valueFrom:
      fieldRef:
        fieldPath: status.hostIP
  - name: bgp_as
    value: "65200"
  - name: mpbgp_ipv4
    valueFrom:
      fieldRef:
        fieldPath: status.hostIP
  - name: mpbgp_nexthop
    value: fixed
  - name: bgp_peers
    value: "<peer-ip>:65200::false/mpbgp_nexthop=auto_sourceip"
```

**Relevant logs**

```
INFO  Starting the BGP server to advertise VIP routes to BGP peers
INFO  [BGP] peer="&{... NeighborAddress:invalid IP ...}"
ERROR start manager err="failed to configure BGP mode: starting BGP server: failed to get MP-BGP addresses: failed to get non link-local IPv6 address: failed to find non-local IP on interface: eth0"
INFO  Shutting down Kube-Vip
```

**Additional context**

The root cause is that setting `mpbgp_ipv4` unconditionally triggers an IPv6 address lookup (`failed to get non link-local IPv6 address`) even when the node has no IPv6 address. The fix should guard the IPv6 MP-BGP path behind a check for actual IPv6 address availability on the interface, rather than treating its absence as a fatal error whenever MP-BGP is configured.

p-strusiewiczsurmacki-mobica:
This PR fixes #1682 

Instead of returning error if the nexthop address for MP-BGP cannot be found, we will now just log an error, but proceed with using just a regular BGP. Added E2E tests for such scenario.

Additionally, while working on this,  I've found an issue with BGP config parameters which could lead to an issue with configuration of BGP for control plane. That was fixed as well - the global variables will be used by default, and will be overridden by per-peer config if needed.

### DIFF
--- a/pkg/bgp/peers.go
+++ b/pkg/bgp/peers.go
@@ -3,7 +3,7 @@ package bgp
 import (
 	"context"
 	"fmt"
-	"log/slog"
+	log "log/slog"
 	"net"
 	"net/netip"
 	"strconv"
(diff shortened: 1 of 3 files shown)


thebsdbox (approved):
LGTM

PR #1684, opened 2026-08-11
fix(dns): compare IP families case-insensitively in getIPbyFamily

jpflouret:
From my comment in https://github.com/kube-vip/kube-vip/issues/1680, this is a fix in `getIPbyFamily` directly.


Cellebyte (approved):
LGTM

PR #1686, opened 2026-08-12
docs: fix broken star history chart in README

Dessalines39394:
The star history chart embedded in the README is currently broken because the provider it used was affected by GitHub's stargazer API restrictions. This change updates the chart image and its surrounding link to use an alternative provider that serves the same data without requiring an API token, restoring a working star history chart for kube-vip.

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,7 +130,7 @@ Additionally it is now relatively easy and quick to develop with [skaffold](http
 
 ## Star History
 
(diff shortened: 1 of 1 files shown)


Cellebyte:
can you please rebase onto main using `--signoff` :)

thebsdbox:
Test failures are irellevant to the changes

PR #1688, opened 2026-08-13
Restart kube-vip on RetryWatcher error

### LINKED ISSUE #1685: Endpoint watcher (watchEndpoint) silently exits on watch.Error (e.g. transient 403), tearing down VIP without recovery
**Describe the bug**

When kube-vip encounters a transient API server connection reset or authorization error during active event streaming (such as during a cluster reboot, control-plane failover, or API server restart), client-go's RetryWatcher sends a watch.Error event to its result channel and terminates its internal loop.

Instead of catching this terminal state, setting an error, or canceling the service context to trigger a retry, watchEndpoint() merely logs the error event and allows the for event := range ch loop to exit naturally when the channel closes.

The function then returns nil, signaling a clean, successful shutdown to the caller. Consequently, kube-vip cleans up the local endpoint watches and removes the ARP/NDP VIP bindings from the host interface (mgmt-br), leaving the cluster VIP unavailable indefinitely without attempting to recover.

logs:
```
2026/08/10 16:10:05 WARN watch auth error, retrying err="services is forbidden: User \"system:serviceaccount:harvester-system:kube-vip\" cannot watch resource \"services\" in API group \"\" at the cluster scope - error from a previous attempt: read tcp 192.168.122.153:39772->10.53.0.1:443: read: connection reset by peer"
E0810 16:10:05.063680       1 retrywatcher.go:144] "Watch failed: ensure the client has valid credentials and watch permissions on the resource" err="endpointslices.discovery.k8s.io is forbidden: User \"system:serviceaccount:harvester-system:kube-vip\" cannot watch resource \"endpointslices\" in API group \"discovery.k8s.io\" in the namespace \"kube-system\""
2026/08/10 16:10:05 ERROR watch error provider=endpointslices err="endpointslices.discovery.k8s.io is forbidden: User \"system:serviceaccount:harvester-system:kube-vip\" cannot watch resource \"endpointslices\" in API group \"discovery.k8s.io\" in the namespace \"kube-system\""
2026/08/10 16:10:05 INFO [endpoint watcher] stopping watching provider=endpointslices "service name"=rke2-traefik namespace=kube-system
2026/08/10 16:10:05 INFO [ARP manager] removing ARP/NDP instance name=192.168.122.131/32-mgmt-br
```

Full diagnostic log available at: [harvester/harvester#11374](https://github.com/harvester/harvester/issues/11374)

RCA:

`RetryWatcher` sends `watch.Error` and terminates:
When client-go receives an HTTP 403 Forbidden or 401 Unauthorized response during a mid-stream reconnection attempt (e.g., when the API server or etcd is temporarily warming up during failover), retrywatcher.go sends a watch.Error status event into ResultChan() and returns true, 0, permanently exiting its internal loop and closing the channel.

watchEndpoint() ignores the terminal error state:

In [pkg/services/watch_endpoints.go](https://github.com/kube-vip/kube-vip/blob/5b2a62a10b01526f03d72bf4adbac4d57def342c/pkg/services/watch_endpoints.go#L90-L97):

```
case watch.Error:
    errObject := apierrors.FromObject(event.Object)
    statusErr, _ := errObject.(*apierrors.StatusError)
    log.Error("watch error", "provider", provider.GetLabel(), "err", statusErr)
```
The error is logged, but no error variable is recorded, svcCtx.Cancel() is not invoked, and execution continues.

Loop terminates and returns nil:
Because `RetryWatcher` closed the channel, the `for event := range ch` loop terminates. Execution reaches the end of watchEndpoint():

Go

```
	log.Info("[endpoint watcher] stopping watching", "provider", provider.GetLabel(), "service name", service.Name, "namespace", service.Namespace)
	return nil //nolint:govet
```
https://github.com/kube-vip/kube-vip/blob/5b2a62a10b01526f03d72bf4adbac4d57def342c/pkg/services/watch_endpoints.go#L96-L97

VIP binding is removed:
The caller interprets nil as a graceful, expected watch termination (such as an intentional deletion), triggering endpoint cleanup and tearing down the ARP/NDP instance from mgmt-br.

(Note: ServicesWatcher in [pkg/services/watch_services.go](https://github.com/kube-vip/kube-vip/blob/5b2a62a10b01526f03d72bf4adbac4d57def342c/pkg/services/watch_services.go#L138-L139) shares a similar structure by returning nil when its event channel closes).

**To Reproduce**

Deploy kube-vip on a multi-node Kubernetes / Harvester cluster managing a LoadBalancer service (e.g., rke2-traefik).

Reboot or restart control-plane nodes / API servers simultaneously to induce transient network resets (connection reset by peer) or temporary API server RBAC authorization errors during reconnects.

Observe kube-vip logs: watchEndpoint encounters the forbidden watch error, logs stopping watching, and unbinds the VIP without restarting or retrying.

**Expected behavior**

When the watch channel closes unexpectedly while the parent service context (svcCtx.Ctx) is still active, watchEndpoint() should detect the abnormal closure (or catch watch.Error), cancel its context, and return an explicit error. This ensures the parent supervisor loop catches the failure and re-establishes the watcher stream using WatchWithAuthRetry.

**Screenshots**

 - OS/Distro: [e.g. Ubuntu 1804]
 - Kubernetes Version: [e.g. v.1.18]   1.30+
 - Kube-vip Version: [e.g. 0.2.3]  v1.21

**`Kube-vip.yaml`:**
If Possible add in your kube-vip manifest (please remove anything that is confidential)

**Additional context**

proposed fix:
```
....
var watchErr error
	for event := range ch {
		switch event.Type {
		// ...
		case watch.Error:
			errObject := apierrors.FromObject(event.Object)
			if statusErr, ok := errObject.(*apierrors.StatusError); ok {
				watchErr = statusErr
			} else {
				watchErr = fmt.Errorf("unknown watch error event: %#+v", event.Object)
			}
			log.Error("watch error", "provider", provider.GetLabel(), "err", watchErr)
			svcCtx.Cancel()
		}
	}

	// If the event channel closed while the service context was NOT cancelled, treat as stream failure
	if svcCtx.Ctx.Err() == nil {
		if watchErr != nil {
			return fmt.Errorf("[%s] endpoint watch stream terminated with error: %w", provider.GetLabel(), watchErr)
		}
		return fmt.Errorf("[%s] endpoint watch channel closed unexpectedly", provider.GetLabel())
	}

	log.Info("[endpoint watcher] stopping watching", "provider", provider.GetLabel(), "service name", service.Name, "namespace", service.Namespace)
	return nil
...
```

NOTE: The issue RCA is partially assisted by AI.

Other related issues/PRs:

https://github.com/kube-vip/kube-vip/issues/1464
https://github.com/kube-vip/kube-vip/issues/1510

https://github.com/kube-vip/kube-vip/pull/1465

p-strusiewiczsurmacki-mobica:
This PR should fix #1685

This adds a call to `killFunc` when `RetryWatcher` error, which should trigger kube-vip's restart.

This is just a quick fix, that was not tested. If no one will not provide better solution (especially one that would not trigger full kube-vip's restart), I'll look into this once again after 31st August while working on #1236 as this is kind of a similar topic.

### DIFF
--- a/pkg/services/processor.go
+++ b/pkg/services/processor.go
@@ -61,6 +61,8 @@ type Processor struct {
 	TunnelMgr *wireguard.TunnelManager
 
 	routeMgr *route.Manager
+
+	killFunc func()
 }
 
 // labelManager is the interface for the node label manager to add/remove labels
@@ -68,7 +70,8 @@ type Processor struct {
 func NewServicesProcessor(config *kubevip.Config, bgpServer *bgp.Server,
 	clientSet *kubernetes.Clientset, rwClientSet *kubernetes.Clientset,
 	intfMgr *networkinterface.Manager, arpMgr *arp.Manager, nodeLabelManager node.Labeler,
-	electionMgr *election.Manager, leaseMgr *lease.Manager, routeMgr *route.Manager) *Processor {
+	electionMgr *election.Manager, leaseMgr *lease.Manager, routeMgr *route.Manager,
+	killFunc func()) *Processor {
 	lbClassFilterFunc := lbClassFilter
 	if config.LoadBalancerClassLegacyHandling {
 		lbClassFilterFunc = lbClassFilterLegacy
@@ -88,6 +91,7 @@ func NewServicesProcessor(config *kubevip.Config, bgpServer *bgp.Server,
 		electionMgr:      electionMgr,
 		TunnelMgr:        wireguard.NewTunnelManager(),
 		routeMgr:         routeMgr,
+		killFunc:         killFunc,
 	}
 }
 
@@ -258,6 +262,8 @@ func (p *Processor) AddOrModify(ctx context.Context, event watch.Event, serviceF
 				}
 				if err = p.watchEndpoint(svcCtx, p.config.NodeName, svc, provider); err != nil {
 					log.Error(err.Error())
+					// if watch endpoint returns error, kill kube-vip
+					p.killFunc()
 				}
 			})
 
--- a/pkg/manager/manager.go
+++ b/pkg/manager/manager.go
@@ -256,10 +256,7 @@ func New(ctx context.Context, configMap string, config *kubevip.Config) (*Manage
 	leaseMgr := lease.NewManager()
 	routeMgr := route.NewManager()
 
-	svcProcessor := services.NewServicesProcessor(config, bgpServer, clientset, rwClientSet,
-		intfMgr, arpMgr, nodeLabelManager, electionMgr, leaseMgr, routeMgr)
-
-	return &Manager{
+	m := &Manager{
 		clientSet:   clientset,
 		rwClientSet: rwClientSet,
 		configMap:   configMap,
@@ -271,7 +268,6 @@ func New(ctx context.Context, configMap string, config *kubevip.Config) (*Manage
 			Help:      "Count all events fired by the service watcher categorised by event type",
 		}, []string{"type"}),
 		signalChan:       signalChan,
-		svcProcessor:     svcProcessor,
 		intfMgr:          intfMgr,
 		arpMgr:           arpMgr,
 		bgpServer:        bgpServer,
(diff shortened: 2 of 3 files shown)


pkg/services/processor.go
```diff
@@ -258,6 +262,8 @@ func (p *Processor) AddOrModify(ctx context.Context, event watch.Event, serviceF
 				}
 				if err = p.watchEndpoint(svcCtx, p.config.NodeName, svc, provider); err != nil {
 					log.Error(err.Error())
+					// if watch endpoint returns error, kill kube-vip
+					p.killFunc()
```
w13915984028, line 266: could you also add another log before call kill?

Kill self to recover is at least a practicical solution to recover from this rare case, thanks.

Cellebyte:
this should be fixed by #1697

PR #1695, opened 2026-08-18
Fix 1685

w13915984028:
On-top-of PR https://github.com/kube-vip/kube-vip/pull/1688 and minor fix

the test log:

https://github.com/kube-vip/kube-vip/issues/1685#issuecomment-5326880505

kube-vip issue https://github.com/kube-vip/kube-vip/issues/1685
original issue https://github.com/harvester/harvester/issues/11374

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -55,7 +55,7 @@ dockerx86Dev:
 
 dockerx86Iptables:
 	@-rm ./kube-vip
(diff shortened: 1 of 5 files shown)


Cellebyte:
fixed by #1697. If not we need to create another MR.

PR #1698, opened 2026-08-20
Refactor/testing to speedup CI in Pull Requests

Cellebyte:


thebsdbox:
awesome, I hope we see some good speedups 😄

Cellebyte:
Yeah still working on this.

Cellebyte:
@thebsdbox it is now revieable :)

pkg/services/processor.go
```diff
@@ -105,9 +105,9 @@ func (p *Processor) AddOrModify(ctx context.Context, event watch.Event, serviceF
 		return nil
 	}
 
-	// We only care about LoadBalancer services
+	// A tracked LoadBalancer must be torn down when its type changes.
 	if svc.Spec.Type != v1.ServiceTypeLoadBalancer {
```
Cellebyte, line 109: this should fix #1700

thebsdbox:
Amazing stuff !

MaxRink (commented):

The PR parallelizes e2e CI (ginkgo --procs=4 per mode, per-process IP offsets and cluster names, per-namespace service-test phases with an errgroup-driven harness), replaces fixed sleeps with polling, fixes the broken 'matrix.mode==' conditions, and adds concurrency groups and timeouts. It also makes two production changes: tearing down tracked VIPs when a Service leaves type LoadBalancer (with a good unit test) and discovering cluster CIDRs from the ServiceCIDR/Node APIs with matching RBAC. The structure is sound and most CI coverage is preserved, but there are real defects: the per-namespace metrics-isolation mechanism is a no-op because the DaemonSet sets an env var kube-vip never reads, the egress CIDR discovery drops the controller-manager pod-CIDR fallback and silently narrows SNAT exclusion from the cluster CIDR to a per-node CIDR, the failover e2e assertions were weakened so an unreachable VIP after failover now passes, and the new ginkgo JSON reports are written to a path the artifact upload never sees.

Overall: significant issues that should be addressed.

### Findings without an inline anchor

**nit** (`testing/services/pkg/deployment/services.go`): StartServiceTest (services.go:32) is dead code now that main.go inlines the phases. The E2E_KEEP_LOGS-conditional removal of TempDirPath and the 'Testing Complete [passed]/[failed]' summary lived only in that function and were silently dropped; run() also collapses multiple Phase A failures into the errgroup's first error, so the exit code no longer counts failures. Delete StartServiceTest and reinstate the summary/cleanup in run().

testing/services/pkg/deployment/kubernetes.go
```diff
@@ -31,113 +34,162 @@ type Service struct {
...
+		{Name: "enable_endpoints", Value: "false"},
+		{Name: "svc_election", Value: "true"},
+		{Name: "EGRESS_CLEAN", Value: "true"},
+		{Name: "vip_loglevel", Value: "-4"},
+		{Name: "egress_withnftables", Value: "true"},
+		// instance_name scopes the nftables egress table to this namespace so
+		// parallel DaemonSets on the same host network don't overwrite each other.
+		{Name: "instance_name", Value: ns},
+		// suppress the default :2112 binding when empty
+		{Name: "prometheusHTTPServer", Value: metricsAddr},
```
MaxRink, line 91: **important**: The env var name 'prometheusHTTPServer' is not read by kube-vip. Environment parsing reads 'prometheus_server' (pkg/kubevip/config_envvar.go:236), and 'prometheusHTTPServer' is only the CLI flag name with default ':2112' (cmd/kube-vip.go:160). On top of that, ParseEnvironment ignores empty values (config_environment.go:704-707), so 'pass empty string to disable prometheus' cannot work through any env var. Result: every parallel DaemonSet still tries to bind :2112 on the shared host network (servePrometheusHTTPServer only logs the bind error), the metricsAddr parameter threaded through EnsureNamespaceWithMetrics is a no-op, and Phase C only works because its ':2112' happens to match the flag default. The phase isolation currently rests entirely on the WaitForNamespaceGone barriers. Either set the value via container args ('--prometheusHTTPServer=...') or use the real 'prometheus_server' env var, and note that disabling requires args, not env.

pkg/services/egress.go
```diff
@@ -392,31 +394,88 @@ func (p *Processor) prepareEgressNftablesTable(serviceUID string, ipv6 bool) err
...
+	serviceCIDR, podCIDR = p.discoverCIDRsFromAPI(ctx)
+	if serviceCIDR == "" {
+		legacyServiceCIDR, legacyErr := p.discoverServiceCIDRFromControllerManager(ctx)
+		if legacyErr != nil {
+			return serviceCIDR, podCIDR, legacyErr
+		}
+		serviceCIDR = legacyServiceCIDR
+	}
+	if podCIDR == "" {
+		return serviceCIDR, podCIDR, fmt.Errorf("unable to determine CNI Pod CIDRs from Node objects; configure egress_podcidr explicitly")
```
MaxRink, line 406: **important**: AutoDiscoverCIDRs no longer falls back to the kube-controller-manager '--cluster-cidr' flag for the pod CIDR; only Node .spec.podCIDR(s) is used and discovery hard-fails otherwise. Clusters whose CNI does its own IPAM without node PodCIDR allocation previously worked via the flag and now get an error. Separately, the semantics change even when node PodCIDRs exist: podCIDRsFromNodes returns per-node CIDRs (e.g. 10.244.0.0/24,10.244.1.0/24,...), and getSameFamilyCidr picks the single CIDR containing the pod IP, so the SNAT ignoreCIDRs in configureEgress now exclude only the local node's pod CIDR instead of the whole cluster CIDR. Egress traffic from the pod to pods on other nodes will now be SNAT'd to the VIP, a behavior change for auto-discovery users. The e2e DaemonSet pins egress_podcidr to the full cluster CIDR, so this narrowing is not exercised by the tests.

testing/services/pkg/deployment/services.go
```diff
@@ -249,9 +261,13 @@ func leaderFailover(ctx context.Context, name, leaderNode *string, clientset *ku
 					slog.Infof("🔍 updated with address [%s] on node [%s]", svc.Status.LoadBalancer.Ingress[0].IP, svc.Annotations[kubevip.VipHost])
 					err = httpTest(svc.Status.LoadBalancer.Ingress[0].IP)
 					if err != nil {
-						return err
+						// Transient failure during annotation updates; keep watching.
+						slog.Warnf("🔍 service [%s] not yet reachable after update: %v", *name, err)
```
MaxRink, line 265: **important**: This weakens the failover assertion twice over. Before: every Modified event with an LB ingress had to pass httpTest or leaderFailover returned an error, and the watch ran the full 30 s window across all five leader kills. Now: httpTest failures only warn and continue, and the function returns nil on the first success, cancelling the killer after typically a single kill. If the VIP never becomes reachable after failover, the watch drains at the 30 s rw.Stop() and the function still returns nil, so Failover passes with only the duplicate-address check. The same pattern is at line 364 in podFailover, where the old code called slog.Fatal on an unreachable VIP. A failover regression that leaves the VIP dead now passes CI. At minimum, track whether any post-kill update ever passed httpTest and return an error from the drain path when none did.

.github/workflows/ci-pull-request.yaml
```diff
@@ -25,21 +40,22 @@ jobs:
         run: make dockerx86Local
       - name: Run Manifest generation tests
         run: make manifest-test
+        if: matrix.mode == 'arp'
       - name: Run ARP mode tests
-        run: DOCKER_API_VERSION=1.48 E2E_KEEP_LOGS=true make e2e-tests-arp
-        if: matrix.mode== 'arp'
+        run: DOCKER_API_VERSION=1.48 E2E_KEEP_LOGS=true GINKGO_ARGS="--json-report=/tmp/kube-vip-test-report-arp.json" make e2e-tests-arp
```
MaxRink, line 45: **minor**: The JSON reports are never uploaded. With no --output-dir, ginkgo v2 computes the report path as filepath.Join(suiteAbsPath, reportName) (ginkgo/internal/profiles_and_reports.go, AbsPathForGeneratedAsset), so '--json-report=/tmp/kube-vip-test-report-arp.json' is written to <repo>/testing/e2e/tmp/kube-vip-test-report-arp.json, which the artifact glob '/tmp/kube-vip-test*' does not match. Use GINKGO_ARGS="--output-dir=/tmp --json-report=kube-vip-test-report-arp.json" (same for rt at line 48 and bgp at line 54).

pkg/services/processor.go
```diff
@@ -311,23 +311,16 @@ func (p *Processor) Delete(event watch.Event, forcedOnly bool) error {
 		return nil
 	}
 
+	return p.deleteTrackedService(svc)
+}
+
+func (p *Processor) deleteTrackedService(svc *v1.Service) error {
```
MaxRink, line 317: **minor**: Every Add/Modified event for every non-LoadBalancer service in the watched scope now flows through deleteTrackedService, which unconditionally logs '(svcs) deleted' at Info level (line 349) even when svcCtx is nil and nothing was tracked. In a cluster with many ClusterIP services this produces a misleading 'deleted' log line for every watch event. Move the log statement inside the 'if svcCtx != nil' block.

testing/e2e/e2e_bgp_healthcheck_test.go
```diff
@@ -23,10 +23,7 @@ import (
 )
 
 var _ = Describe("kube-vip BGP ControlPlane health-check", Ordered, func() {
-	if Mode != ModeBGP {
-		return
-	}
-
+	if Mode == ModeBGP {
```
MaxRink, line 26: **minor**: The body of the new 'if Mode == ModeBGP {' block was not re-indented, so the file is no longer gofmt-clean ('gofmt -l' flags it; the file on main was clean). CI does not catch this because 'make check' only formats '*.go pkg cmd' and only warns. Run gofmt on the file, or restructure the guard the way the sibling files do.

testing/services/pkg/deployment/tests.go
```diff
@@ -1073,26 +1184,38 @@ func checkNoDuplicateElectionLoops(namespace, name, fault string) error {
 func checkLeaseErrorsStable(namespace, name, fault string) error {
 	const metric = "kube_vip_service_election_errors_total"
 
-	first, err := scrapeServiceGauge(metric, namespace, name)
+	// Poll until two consecutive readings are identical; the counter only grows on
+	// the #1664 bug path so we stop as soon as it has been stable for one interval.
+	deadline := time.Now().Add(time.Second * 12)
```
MaxRink, line 1189: **minor**: checkLeaseErrorsStable now declares the counter stable after a single 2 s interval without growth, where the old code required a full 10 s quiet window. The #1664 bug path it guards is a failing election retry loop; retry periods of a few seconds can easily produce two identical consecutive 2 s samples and a false pass. Require two or three consecutive stable intervals before returning success.

.github/workflows/ci.yaml
```diff
@@ -1,11 +1,18 @@
 name: For each commit
 on:
   push:
+    branches: [main]
   pull_request:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
```
MaxRink, line 9: **minor**: cancel-in-progress: true applies to push runs on main too (group resolves to '<workflow>-refs/heads/main'), so back-to-back merges cancel in-progress main builds and intermediate commits end with cancelled runs. Scope it with cancel-in-progress: ${{ github.event_name == 'pull_request' }}. Also note the new 'branches: [main]' filter removes the tag-push trigger the bare 'push' event previously included; confirm the release workflow covers lint/tests for tags.

pkg/services/services.go
```diff
@@ -541,6 +550,19 @@ func (p *Processor) updateEgressConfiguration(ctx context.Context, svc *v1.Servi
 		return fmt.Errorf("service instance not found for %s/%s", svc.Namespace, svc.Name)
 	}
 
+	// The svc snapshot may have been captured before the LB IP was assigned.
+	// Refresh from the API so FetchServiceAddresses sees the current ingress.
```
MaxRink, line 554: **nit**: updateEgressConfiguration issues a Services Get before the unchanged-endpoint short-circuit at line 571. With the new ActionNone hook this Get now runs on every no-op watch event for every egress service. Compare the ActiveEndpoint annotations first and only refresh from the API when they differ.

PR #1699, opened 2026-08-20
refactor/endpoints for different modes

Cellebyte:
* I was finding some small nits around endpoint handling.

### DIFF
--- a/pkg/endpoints/endpoints.go
+++ b/pkg/endpoints/endpoints.go
@@ -90,11 +90,7 @@ func (p *Processor) AddOrModify(svcCtx *servicecontext.Context, event watch.Even
 
 		svcCtx.SignalReadiness()
 
-		// There are local endpoints available on the node
-		// Process immediately if:
-		// - No services/leader election is enabled, OR
-		// - WireGuard is enabled (it always needs immediate DNAT rule updates)
-		if (!p.config.EnableServicesElection && !p.config.EnableLeaderElection) || p.config.EnableWireguard {
+		if p.shouldProcessInstance() {
 			if err := p.worker.processInstance(svcCtx, service); err != nil {
 				return false, fmt.Errorf("failed to process non-empty instance: %w", err)
 			}
@@ -107,21 +103,13 @@ func (p *Processor) AddOrModify(svcCtx *servicecontext.Context, event watch.Even
 			}
 			svcCtx.SignalReadiness()
 
-			if (!p.config.EnableServicesElection && !p.config.EnableLeaderElection) || p.config.EnableWireguard {
+			if p.shouldProcessInstance() {
 				if err := p.worker.processInstance(svcCtx, service); err != nil {
 					return false, fmt.Errorf("failed to process endpointless instance: %w", err)
 				}
 			}
 		} else if svcCtx.Signalled.Load() {
-			// There are no local endpoints
-			svcCtx.ResetReadiness()
-			p.worker.clear(svcCtx, lastKnownGoodEndpoint, service)
-			if p.config.EnableARP && !p.config.EnableServicesElection {
-				i := instance.FindServiceInstance(service, *p.instances)
-				for _, c := range i.Clusters {
-					c.Stop()
-				}
-			}
+			p.handleNoEndpoints(svcCtx, service, lastKnownGoodEndpoint)
 		}
 	}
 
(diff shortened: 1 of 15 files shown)


pkg/endpoints/endpoints.go
```diff
@@ -133,9 +134,23 @@ func (p *Processor) AddOrModify(svcCtx *servicecontext.Context, event watch.Even
...
+
+	endpoints, err := p.worker.getEndpoints(service, id)
+	if err != nil {
+		return err
+	}
+	if err := p.worker.setInstanceEndpointsStatus(svcCtx.Ctx, service, endpoints); err != nil {
+		log.Error("updating instance", "err", err)
+	}
+
+	if len(endpoints) == 0 && svcCtx.Signalled.Load() && !shouldAllowReconcileWithoutEndpoints(service) {
```
MaxRink: **important**: Delete's zero-endpoints branch mirrors AddOrModify's (ResetReadiness + clear) but omits the ARP cluster stop that AddOrModify performs at lines 119-124 (EnableARP && !EnableServicesElection: stop all instance clusters). Deleting the last EndpointSlice in ARP global-leader mode leaves the VIP clusters running and the address advertised with no backends, while the equivalent Modified-to-zero event stops them. Add the same cluster-stop block here or extract a shared handleNoEndpoints helper used by both paths.

pkg/endpoints/endpoints.go
```diff
@@ -133,9 +134,23 @@ func (p *Processor) AddOrModify(svcCtx *servicecontext.Context, event watch.Even
...
+func (p *Processor) Delete(svcCtx *servicecontext.Context, service *v1.Service, id string,
+	object runtime.Object, lastKnownGoodEndpoint *string) error {
+	if err := p.provider.DeleteObject(object); err != nil {
+		return fmt.Errorf("[%s] error deleting object: %w", p.provider.GetLabel(), err)
+	}
+
+	endpoints, err := p.worker.getEndpoints(service, id)
+	if err != nil {
+		return err
+	}
```
MaxRink: **important**: When a slice deletion leaves endpoints > 0, Delete only calls setInstanceEndpointsStatus. Unlike AddOrModify it skips updateLastKnownGoodEndpoint, so if lastKnownGoodEndpoint was in the deleted slice it stays stale: egress rules are not torn down or moved (removeEgress never runs), and the active-endpoint annotations are not updated until an unrelated Add/Modified event arrives. The EndpointSlice controller can delete a whole slice during consolidation without touching the surviving slice, so no follow-up event is guaranteed. Run the same lastKnownGoodEndpoint/annotation reconciliation as AddOrModify when endpoints remain.

pkg/endpoints/endpoints.go
```diff
@@ -133,9 +134,23 @@ func (p *Processor) AddOrModify(svcCtx *servicecontext.Context, event watch.Even
...
-func (p *Processor) Delete(ctx context.Context, service *v1.Service, id string) error {
-	if err := p.worker.delete(ctx, service, id); err != nil {
-		return fmt.Errorf("[%s] error deleting service: %w", p.provider.GetLabel(), err)
+func (p *Processor) Delete(svcCtx *servicecontext.Context, service *v1.Service, id string,
+	object runtime.Object, lastKnownGoodEndpoint *string) error {
+	if err := p.provider.DeleteObject(object); err != nil {
+		return fmt.Errorf("[%s] error deleting object: %w", p.provider.GetLabel(), err)
+	}
+
+	endpoints, err := p.worker.getEndpoints(service, id)
```
MaxRink: **important**: In WireGuard mode, Delete never calls worker.processInstance, but wireguardWorker.processInstance is what rebuilds the DNAT target set from current endpoints (pkg/endpoints/endpoints_wireguard.go:45). A slice deletion that leaves other endpoints intact keeps DNAT rules pointing at the removed addresses until the next Add/Modified event, so traffic is NATed to dead backends. AddOrModify handles this via the EnableWireguard branch at line 97; Delete needs the equivalent for the endpoints > 0 case.

pkg/endpoints/providers/endpointslices.go
```diff
@@ -59,56 +58,62 @@ func (ep *Endpointslices) LoadObject(endpoints runtime.Object, cancel context.Ca
...
-				}
-				localEndpoints = append(localEndpoints, address)
+	for _, eps := range ep.slices {
+		for _, endpoint := range eps.Endpoints {
+			// Per the EndpointConditions.Serving godoc: if nil, defer to Ready.
+			serving := endpoint.Conditions.Serving
+			if serving == nil {
+				serving = endpoint.Conditions.Ready
+			}
+			if serving == nil || !*serving {
```
MaxRink: **important**: The fallback from nil Serving to Ready matches the Serving godoc, but the follow-through is still wrong: per the EndpointConditions.Ready godoc, a nil Ready means unknown and 'consumers should interpret this unknown state as ready' (kube-proxy does). Here an endpoint with both conditions nil, which is common in manually managed EndpointSlices for external workloads, is skipped and never selected as a local endpoint. After deferring to Ready, treat nil as true instead of continuing.

pkg/manager/worker/bgp.go
```diff
@@ -113,6 +113,8 @@ func (b *BGP) StartServices(ctx context.Context) error {
 		if err := b.PerServiceLeader(ctx, false); err != nil {
 			return err
 		}
+	} else if b.config.EnableLeaderElection {
```
MaxRink, line 116: **important**: This silently changes the datapath for existing deployments. EnableLeaderElection also selects control-plane election in BGP.StartControlPlane, so a BGP user who enabled it for the control plane previously got ServicesNoLeader: every node with the service advertised the VIP (ECMP/multipath). After this change the same config runs a single global services leader and only one node advertises. The alignment with ARP/RoutingTable is defensible, but this needs at least a release note, and ideally a config validation warning, since nothing in the config changed but service HA semantics did.

pkg/endpoints/providers/endpointslices_test.go
```diff
@@ -0,0 +1,110 @@
...
+	provider := NewEndpointslices().(*Endpointslices)
+	serving := true
+	nodeName := "node-1"
+
+	slice1 := &discoveryv1.EndpointSlice{
+		ObjectMeta:  metav1.ObjectMeta{Name: "slice-1"},
+		AddressType: discoveryv1.AddressTypeIPv4,
+		Endpoints: []discoveryv1.Endpoint{{
+			Addresses:  []string{"10.0.0.1"},
+			Conditions: discoveryv1.EndpointConditions{Serving: &serving},
```
MaxRink, line 22: **minor**: The new tests always set Conditions.Serving explicitly, so the dedicated 'nil Serving defers to Ready' fix commit has no coverage: no case with Serving nil + Ready true, Serving nil + Ready nil, or Serving false + Ready true. Processor.Delete's new recompute-and-clear logic is also untested (endpoints_test.go only exercises AddOrModify via fakeWorker). Add condition-permutation cases here and a Delete case to the processor test.

MaxRink (commented):

The PR replaces the per-mode delete()/deleteAction() hooks with a provider-level DeleteObject plus a single recompute-then-withdraw path in Processor.Delete, and it fixes a real base bug: EndpointSlice deletions previously acted on stale pre-deletion provider state (or were no-ops), and two slices of the same address family overwrote each other in the Endpointslices provider. The new per-slice map, the reference-counted route cleanup in deleteService (verified against route.Manager.Delete, which is refcounted and no-op safe), and the removal of WireGuard's duplicate services watcher are all sound, and the interface shrinks rather than grows. However, the Delete path only mirrors part of AddOrModify's zero-endpoint behavior (ARP cluster stop and the remaining-endpoints bookkeeping are missing), the Serving/Ready nil handling still contradicts the k8s convention that a nil Ready means ready, and routing BGP's EnableLeaderElection to a global services leader silently changes the datapath for existing BGP deployments that had that flag set for the control plane. Cross-compile (GOOS=linux go build ./...) and vet pass; new provider tests cover slice tracking but not the ready-condition fallback or Processor.Delete itself.

Overall: minor issues, nothing blocking.

pkg/manager/worker/wireguard.go
```diff
@@ -194,6 +191,17 @@ func (w *WireGuard) watchKubernetesEndpoints(ctx context.Context, tunnelConfig *
 			if err := w.updateControlPlaneDNAT(tunnelConfig, endpoints); err != nil {
 				log.Error("failed to update control plane DNAT rules", "err", err)
 			}
+		case watch.Deleted:
```
MaxRink: **minor**: The watch.Deleted case body duplicates the Added/Modified body except for the LoadObject/DeleteObject call. Collapse to a single case that picks the mutation based on event.Type, then shares the GetAllEndpoints + updateControlPlaneDNAT tail; that also keeps the log message ('kubernetes endpoints changed') from being maintained twice.

pkg/endpoints/endpoints.go
```diff
@@ -133,9 +134,23 @@ func (p *Processor) AddOrModify(svcCtx *servicecontext.Context, event watch.Even
...
+	if err != nil {
+		return err
```
MaxRink: **nit**: The getEndpoints error is returned unwrapped while every other error in this function and in AddOrModify carries the '[%s]' provider label prefix. Wrap it for consistent log correlation: fmt.Errorf("[%s] error getting endpoints: %w", p.provider.GetLabel(), err).

pkg/endpoints/providers/endpoints.go
```diff
@@ -61,6 +61,16 @@ func (ep *Endpoints) LoadObject(endpoints runtime.Object, cancel context.CancelF
 	return nil
 }
 
+func (ep *Endpoints) DeleteObject(endpoints runtime.Object) error {
```
MaxRink, line 66: **nit**: DeleteObject type-asserts the incoming object only to discard it and unconditionally resets ep.endpoints to an empty struct. That is correct today because exactly one v1.Endpoints object exists per service, but the asymmetry with the slice provider (which deletes by name) is non-obvious; a one-line comment stating the single-object invariant would prevent someone 'fixing' it later.

Cellebyte: Added a comment as suggested

pkg/endpoints/providers/endpointslices_test.go
Cellebyte: addressed.

pkg/manager/worker/bgp.go
Cellebyte: Thanks for pointing that out.

PR #1701, opened 2026-08-21
fix(egress): prevent stale active-endpoint overwrite

### LINKED ISSUE #1673: Egress active-endpoint annotation is overwritten with a stale empty value during service reconciliation
### Describe the bug

For an egress-enabled LoadBalancer Service, kube-vip can write the correct endpoint IP to `kube-vip.io/active-endpoint` and then overwrite it with an empty or older value during the same service reconciliation.

The endpoint is discovered successfully, the Service update succeeds, and egress reconfiguration is invoked with the correct Pod IP. Immediately afterwards, however, the persisted Service annotation is empty. Kubernetes `managedFields` identifies kube-vip as the writer, and manually setting the annotation succeeds, which rules out RBAC as the primary cause.

A representative log ordering is:

```text
DEBUG updated active endpoint annotation service=egress-service namespace=default endpoint=172.30.2.40
INFO  [service] updating egress configuration service=egress-service namespace=default old_ipv4="" new_ipv4=172.30.2.40 old_ipv6="" new_ipv6=""
INFO  [service] egress configuration updated successfully service=egress-service namespace=default
```

The Service then contains:

```yaml
kube-vip.io/active-endpoint: ""
```

The same behavior is reproducible after an endpoint failover that temporarily leaves the Service with zero endpoints.

### Environment and configuration

The issue was reproduced with:

- kube-vip v1.2.0
- Kubernetes v1.34.4
- three kube-vip replicas
- Routing Table mode
- `svc_enable=true`
- `svc_election=true`
- `vip_leaderelection=true`
- `enable_endpoints=true`
- `externalTrafficPolicy: Local`
- IPv4 LoadBalancer Service with IPv4 egress enabled
- dual-stack Pods where the legacy Endpoints object contains the correct IPv4 address

The conflicting code is also present in v1.2.1, v1.2.2, and current `main`. The same logical race applies to both the Endpoints and EndpointSlice providers.

### Root cause

There are two concurrent writers for the same annotation.

The endpoint watcher selects the current endpoint and persists it through `UpdateServiceAnnotation`:

https://github.com/kube-vip/kube-vip/blob/cfd86de936441e86864a6cd99ae53f367d0c853e/pkg/endpoints/endpoints.go#L177-L240

At service startup, the service handler and endpoint watcher run concurrently:

https://github.com/kube-vip/kube-vip/blob/cfd86de936441e86864a6cd99ae53f367d0c853e/pkg/services/processor.go#L234-L262

After configuring egress, `configureService` performs another annotation update using the `svc` object captured when service handling started:

https://github.com/kube-vip/kube-vip/blob/cfd86de936441e86864a6cd99ae53f367d0c853e/pkg/services/services.go#L364-L380

That captured Service can still contain an empty or previous active endpoint. The provider fetches the latest Service inside `RetryOnConflict`, but then assigns the endpoint value supplied by the caller:

https://github.com/kube-vip/kube-vip/blob/cfd86de936441e86864a6cd99ae53f367d0c853e/pkg/endpoints/providers/endpoints.go#L100-L116

Therefore conflict retries protect the Kubernetes `resourceVersion`, but do not prevent this semantic stale write.

The observed interleaving is:

1. The Service watcher captures a Service whose active endpoint is empty.
2. The endpoint watcher selects `172.30.2.40` and persists it successfully.
3. The endpoint callback reconfigures egress with `172.30.2.40`.
4. `configureService` resumes with the older Service object.
5. Its second `UpdateServiceAnnotation` call passes an empty endpoint.
6. The provider fetches the latest Service and successfully writes the caller-supplied empty value.

No Kubernetes conflict or update error is expected in this ordering.

### Steps to reproduce

1. Run multiple kube-vip replicas with per-Service election and egress enabled.
2. Create an egress-enabled LoadBalancer Service with no initial `kube-vip.io/active-endpoint` value.
3. Publish one matching local endpoint.
4. Observe the successful `updated active endpoint annotation` log.
5. Read the Service annotation and observe that it becomes empty.
6. Set the annotation manually to the current endpoint.
7. Delete the active workload Pod so the endpoint temporarily becomes empty and is then repopulated.
8. Observe the annotation becoming empty again despite a valid endpoint and recovered Lease holder.

### Expected behavior

The endpoint watcher should be the authoritative owner of:

- `kube-vip.io/active-endpoint`
- `kube-vip.io/active-endpoint-ipv6`

Once it persists the selected endpoint, service initialization must not overwrite it using an older Service snapshot.

### Suggested fix

Remove the duplicate `UpdateServiceAnnotation` call from `configureService`. The endpoint watcher already persists these annotations and directly invokes egress reconfiguration.

Regression coverage should include:

- an endpoint writer updating the Service while `configureService` still holds an older empty snapshot;
- initial IPv4 endpoint selection;
- endpoint to zero endpoints to repopulated endpoint;
- preservation of the other address-family annotation;
- both Endpoints and EndpointSlice modes;
- ensuring egress reconfiguration is invoked once after a successful annotation update.

### Related work

- #1428
- #1433
- #1481

PR #1433 introduced direct persistence from the endpoint watcher and the egress update callback, but the older persistence call in `configureService` remained. The recent endpoint-election fixes #1664 and #1668 do not remove this duplicate writer.

A separate, broader concern remains: with multiple kube-vip replicas and `externalTrafficPolicy: Local`, endpoint watchers are not gated on per-Service leadership and can still contend on the cluster-wide annotation. That requires separate ownership/leader-transition design and integration coverage; it is not needed to explain the deterministic empty overwrite described here.

Cellebyte:
Make the endpoint watcher the sole active-endpoint annotation writer so a Service snapshot captured during startup cannot replace a newer endpoint. Preserve intentionally empty snapshots across endpoint-to-zero-to-same transitions, select the cleared annotation from the configured egress family, reject malformed or wrong-family endpoints, and preserve the other family in dual-stack state.

Successor of #1674 
Fixes #1673.

### DIFF
--- a/pkg/endpoints/endpoints.go
+++ b/pkg/endpoints/endpoints.go
@@ -179,28 +179,41 @@ func (p *Processor) updateAnnotations(service *v1.Service, lastKnownGoodEndpoint
 	egressUpdateFunc func(context.Context, *v1.Service) error) {
 	// Set the service accordingly
 	if service.Annotations[kubevip.Egress] == "true" {
-		ip := net.ParseIP(*lastKnownGoodEndpoint)
+		if *lastKnownGoodEndpoint != "" {
+			ip := net.ParseIP(*lastKnownGoodEndpoint)
+			expectIPv6 := service.Annotations[kubevip.EgressIPv6] == "true"
+			if ip == nil || (ip.To4() == nil) != expectIPv6 {
+				log.Warn("ignoring active endpoint with unexpected address family",
+					"service", service.Name,
+					"namespace", service.Namespace,
+					"endpoint", *lastKnownGoodEndpoint,
+					"expected_ipv6", expectIPv6)
+				return
+			}
+		}
 
 		// Store old values from ServiceSnapshot to detect if annotation actually changed
 		// We use the ServiceSnapshot instead of the service parameter because the service parameter
 		// may have stale annotations if the last update failed
 		var oldEndpoint, oldEndpointIPv6 string
+		snapshotFound := false
 		if p.instances != nil {
 			serviceInstance := instance.FindServiceInstance(service, *p.instances)
-			if serviceInstance != nil {
+			if serviceInstance != nil && serviceInstance.ServiceSnapshot != nil {
+				snapshotFound = true
 				oldEndpoint = serviceInstance.ServiceSnapshot.Annotations[kubevip.ActiveEndpoint]
 				oldEndpointIPv6 = serviceInstance.ServiceSnapshot.Annotations[kubevip.ActiveEndpointIPv6]
 			}
 		}
-		// Fall back to service annotations if we couldn't find the instance
-		if oldEndpoint == "" && oldEndpointIPv6 == "" {
+		// Empty annotations in an existing snapshot are meaningful after a zero-endpoint transition.
+		if !snapshotFound {
 			oldEndpoint = service.Annotations[kubevip.ActiveEndpoint]
 			oldEndpointIPv6 = service.Annotations[kubevip.ActiveEndpointIPv6]
 		}
 
 		// Determine which annotation to update based on IP version
 		var endpoint, endpointIPv6 string
-		if ip.To4() == nil && !p.config.EnableEndpoints {
+		if service.Annotations[kubevip.EgressIPv6] == "true" && !p.config.EnableEndpoints {
 			// IPv6
 			endpointIPv6 = *lastKnownGoodEndpoint
 			endpoint = oldEndpoint // Preserve existing IPv4 if any
--- a/pkg/endpoints/endpoints_test.go
+++ b/pkg/endpoints/endpoints_test.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"github.com/kube-vip/kube-vip/pkg/endpoints/providers"
+	"github.com/kube-vip/kube-vip/pkg/instance"
 	"github.com/kube-vip/kube-vip/pkg/kubevip"
 	"github.com/kube-vip/kube-vip/pkg/lease"
 	"github.com/kube-vip/kube-vip/pkg/metrics"
@@ -17,6 +18,7 @@ import (
 	discoveryv1 "k8s.io/api/discovery/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/watch"
+	"k8s.io/client-go/kubernetes"
 )
 
 func TestShouldAllowReconcileWithoutEndpoints(t *testing.T) {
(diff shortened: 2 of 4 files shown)


MaxRink (commented):

The production fix is correct and verified: removing the UpdateServiceAnnotation call from configureService makes the endpoint watcher (pkg/endpoints/endpoints.go updateAnnotations) the only annotation writer, the snapshotFound flag correctly preserves intentionally empty snapshots after endpoint-to-zero transitions, and keying the family branch on the egress-ipv6 annotation instead of parsing the (possibly empty) endpoint fixes the cleared annotation landing in the wrong family slot. The three new endpoints tests are good: compiled against origin/main they fail there and pass on the PR head. The problem is the headline regression test: TestConfigureServiceDoesNotOverwriteActiveEndpoint also passes against the unfixed base code, because client-go v0.36 sends the Service Update as protobuf, the fake API server's JSON decoder rejects it with 400, and updateRequests never increments, so the stale overwrite is attempted on base but invisible to the assertions. There is also a residual staleness gap: configureService still programs egress SNAT rules from the captured startup service annotations, which can install rules for a stale endpoint that no teardown path removes. No draft leftovers (TODOs, debug logging) were found in the diff.

Overall: significant issues that should be addressed.

### Findings without an inline anchor

**important** (`pkg/services/services.go`): configureService still programs egress rules from the captured startup service object (svc.Annotations[kubevip.ActiveEndpoint]/[ActiveEndpointIPv6] around lines 315-354), which is the same stale snapshot the removed annotation write used. In the race this PR fixes (endpoint watcher updates the annotation and calls updateEgressConfiguration before the elected node runs addService/configureService), configureService can still install SNAT rules for the old endpoint. Nothing tears those down: updateEgressConfiguration's teardown and deleteService's egress.Teardown both use the instance ServiceSnapshot, which by then holds the newer endpoint, so the stale iptables rules leak until reboot. Consider reading the egress section's endpoint from inst.ServiceSnapshot (post endpoint-watcher update) instead of the svc parameter.

pkg/services/services_test.go
```diff
@@ -1,14 +1,90 @@
...
+		defer mutex.Unlock()
+		writer.Header().Set("Content-Type", "application/json")
+		switch request.Method {
+		case http.MethodGet:
+			if err := json.NewEncoder(writer).Encode(currentService); err != nil {
+				t.Errorf("encode Service response: %v", err)
+			}
+		case http.MethodPut:
+			updatedService := &v1.Service{}
+			if err := json.NewDecoder(request.Body).Decode(updatedService); err != nil {
```
MaxRink, line 47: **blocking**: TestConfigureServiceDoesNotOverwriteActiveEndpoint does not guard the regression: compiled against origin/main's services.go it still passes. The base code does attempt the stale write (log shows 'error updating Service Spec ... put services test-service'), but client-go v0.36 sends the Update body as application/vnd.kubernetes.protobuf (verified empirically with rest.Config{Host: server.URL}), so json.NewDecoder(request.Body).Decode fails, the handler returns 400, updateRequests stays 0, and both assertions pass without the fix. Set ContentType: "application/json" in the rest.Config (rest.Config{Host: server.URL, ContentConfig: rest.ContentConfig{ContentType: "application/json"}}), and move updateRequests++ before the decode so a rejected PUT still counts as an attempted overwrite.

pkg/endpoints/endpoints.go
```diff
@@ -179,28 +179,41 @@ func (p *Processor) updateAnnotations(service *v1.Service, lastKnownGoodEndpoint
...
+		if *lastKnownGoodEndpoint != "" {
+			ip := net.ParseIP(*lastKnownGoodEndpoint)
+			expectIPv6 := service.Annotations[kubevip.EgressIPv6] == "true"
+			if ip == nil || (ip.To4() == nil) != expectIPv6 {
+				log.Warn("ignoring active endpoint with unexpected address family",
+					"service", service.Name,
+					"namespace", service.Namespace,
+					"endpoint", *lastKnownGoodEndpoint,
+					"expected_ipv6", expectIPv6)
+				return
```
MaxRink, line 191: **minor**: The wrong-family early return can wedge a live service when kube-vip.io/egress-ipv6 is toggled: updateLastKnownGoodEndpoint only replaces lastKnownGoodEndpoint when it disappears from the endpoint list, so a still-existing old-family endpoint hits this branch on every event, logging the warning forever while the annotation and egress rules are never cleared or updated until endpoint churn. Consider clearing lastKnownGoodEndpoint (or falling through to a clear) instead of returning.

pkg/endpoints/endpoints.go
```diff
@@ -179,28 +179,41 @@ func (p *Processor) updateAnnotations(service *v1.Service, lastKnownGoodEndpoint
...
+			}
+		}
 
 		// Store old values from ServiceSnapshot to detect if annotation actually changed
 		// We use the ServiceSnapshot instead of the service parameter because the service parameter
 		// may have stale annotations if the last update failed
 		var oldEndpoint, oldEndpointIPv6 string
+		snapshotFound := false
 		if p.instances != nil {
 			serviceInstance := instance.FindServiceInstance(service, *p.instances)
```
MaxRink, line 201: **minor**: FindServiceInstance and the ServiceSnapshot read run on the endpoint-watcher goroutine without the services.Processor mutex, while services.Processor.AddOrModify appends to ServiceInstances and deleteService replaces the slice (under p.mutex, which this reader does not take). The race is pre-existing, but this PR makes the snapshot the sole source for change detection, so a torn read now means a missed or duplicated annotation write plus egress reconfigure. Worth passing an accessor that takes the processor mutex instead of the bare *[]*instance.Instance.

pkg/endpoints/endpoints_test.go
```diff
@@ -47,6 +49,191 @@ type fakeWorker struct {
...
+						repopulated = annotationUpdate{endpoint: family.other, endpointIPv6: family.endpoint}
+					} else {
+						cleared = annotationUpdate{endpointIPv6: family.other}
+						repopulated = annotationUpdate{endpoint: family.endpoint, endpointIPv6: family.other}
+					}
+				}
+				want := []annotationUpdate{cleared, repopulated}
+				if len(recorder.updates) != len(want) {
+					t.Fatalf("annotation updates = %+v, want %+v", recorder.updates, want)
+				}
```
MaxRink, line 138: **nit**: The updateSnapshot fake unconditionally commits the snapshot, so the suite only covers the path where egress reconfiguration succeeds. A case where egressUpdateFunc returns an error would pin the retry behavior (snapshot not advanced, next event repeats both the annotation write and reconfigure), which is the failure-path contract updateAnnotations now relies on.

PR #1704, opened 2026-08-23
fix(e2e): guard etcd suite teardown against failed cluster creation

MaxRink:
## Purpose

Preserve the original etcd E2E setup failure instead of masking it with a teardown panic.

## Key changes

- Guards suite cleanup when cluster creation returned nil.
- Makes `Cluster.Delete` safe for a nil receiver.

## Validation

The etcd E2E suite compiles with `--tags=e2e`; formatting, vet, and current CI checks pass.

## Dependency

### DIFF
--- a/testing/e2e/etcd/cleanup_test.go
+++ b/testing/e2e/etcd/cleanup_test.go
@@ -1,0 +1,23 @@
+//go:build e2e
+// +build e2e
+
+package etcd_test
+
+import (
+	"path/filepath"
+	"testing"
+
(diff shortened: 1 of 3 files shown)


README.md
p-strusiewiczsurmacki-mobica, line 135: I believe this was already added and should not be a part of this PR.

MaxRink: Removed. The current diff no longer changes README.md.

p-strusiewiczsurmacki-mobica (approved):

PR #1705, opened 2026-08-23
fix: merge all config file fields dropped by the field-by-field merge

MaxRink:
## Purpose

Make runtime configuration precedence presence-aware across defaults, config files, environment variables, and flags.

## Key changes

- Tracks whether values were explicitly supplied instead of inferring source from zero values.
- Applies config-file values without dropping valid fields, then lets explicitly set environment variables and flags override them.
- Aligns BGP configuration handling and documents the supported precedence behavior.
- Adds source-precedence coverage across command and kubevip configuration paths.

## Validation

## Dependency

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Fixed
+- Make `manager` and `service` runtime configuration presence-aware, with explicit CLI flags taking precedence over environment variables, configuration files, and defaults. Strict YAML/JSON loading now preserves explicit false, zero, and empty values and normalizes standalone BGP peer configuration for runtime use.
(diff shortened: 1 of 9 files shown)


MaxRink:
The failing bgp e2e leg is the pre-existing testing/e2e/bgp/server.go:227 MP-BGP path-propagation timeout flake, unrelated to this config-file-merge change. Fixed by #1706. Merge order: #1706 first.

pkg/kubevip/config_merge_reflection_test.go
```diff
@@ -0,0 +1,396 @@
...
+	}
+	sort.Strings(paths)
+	t.Logf("tested %d reflected merge fields", len(fields))
+	t.Logf("known unmerged fields (%d):", len(paths))
+	for _, path := range paths {
+		t.Logf("  %s", path)
+	}
+}
+
+func collectMergeReflectionFields(t *testing.T, typ reflect.Type, prefix string) []mergeReflectionField {
```
p-strusiewiczsurmacki-mobica: Is there any reason to add `prefix` parameter if it's not used in the calls?

PR #1709, opened 2026-08-23
fix(route): insert route tracker entry only after AddRoute succeeds

MaxRink:
## Bug
`Manager.Add` inserted `m.tracker[key]` before calling `r.AddRoute()`. When AddRoute returned a non-EEXIST error, Add returned the error but left the tracker entry in place (empty objects). The next Add saw `exists==true`, skipped the whole add block, and returned nil, so the route was never installed while callers set `ConfiguredNetworks=true` believing it was. Permanent silent failure until restart.

## Fix
Insert into `m.tracker` only after AddRoute (and the EEXIST/update branch) succeeds. Adds `TestManagerRetriesRouteAfterInitialAddFailure`.

## Rebase / CI note

### DIFF
--- a/pkg/endpoints/endpoints_wireguard_test.go
+++ b/pkg/endpoints/endpoints_wireguard_test.go
@@ -6,7 +6,7 @@ import (
 	v1 "k8s.io/api/core/v1"
 )
 
-func TestWireguardDeleteDoesNotDereferenceNilServiceContext(t *testing.T) {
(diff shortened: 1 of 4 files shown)


p-strusiewiczsurmacki-mobica:
LGTM

p-strusiewiczsurmacki-mobica (approved):

p-strusiewiczsurmacki-mobica (approved):

PR #1710, opened 2026-08-23
fix(cluster): make Cluster.Stop concurrency-safe and lock Processor.Stop

MaxRink:
## Bug
`Cluster.Stop()` did `close(cluster.stop); cluster.stop = make(chan bool)` with no synchronization. During shutdown, `Processor.Stop()` iterates `ServiceInstances` calling `cluster.Stop()` without holding `p.mutex`, concurrently with `deleteService()` (from election `onStoppedLeading`) which also calls `cluster.Stop()`. Concurrent Stop on the same cluster double-closes the same channel (panic) and races the `ServiceInstances` slice.

## Fix
Guard `Cluster.Stop()` with a mutex so close+recreate is atomic and idempotent under concurrency; take `p.mutex` in `Processor.Stop()`. Adds `TestStopConcurrentDoesNotRaceOrPanic` (validated under `-race`: DATA RACE + panic without the fix, clean with it).

## Rebase / CI note

### DIFF
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"net/http"
 	"os"
+	"sync"
 	"time"
 
 	log "log/slog"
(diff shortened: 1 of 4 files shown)


p-strusiewiczsurmacki-mobica (approved):
LGTM

PR #1711, opened 2026-08-23
fix(bgp): honour configured peer port and keep source interface on MP-BGP fallback

MaxRink:
## Bugs (both in pkg/bgp/peers.go)
1. `AddPeer` built the peer Transport with `RemotePort: defaultBGPPort` (179) unconditionally and never read `peer.Port`, so a configured/parsed per-peer port was silently dropped.
2. When `MpbgpNexthop` is set but `FindMpbgpAddresses` fails (the #1683 graceful degrade to regular BGP), AddPeer took the error branch that skips the block applying `SourceIP`/`SourceIF`, dropping the configured source interface (breaks IPv6 link-local / source-routed sessions).

## Fix
Use `peer.Port` when non-zero; apply `SourceIP`/`SourceIF` on the MP-BGP fallback path too.

## Generalized verification
The two narrow tests are consolidated into the table-driven `TestAddPeerConfiguresTransportOptions`, covering both configured remote-port handling and configured source-interface preservation after MP-BGP resolution failure.

## Verification
- `go test ./pkg/bgp`
- `go test -race ./pkg/bgp`

Both commands passed in the Debian Bookworm ARM64 OrbStack Linux VM.

## Rebase note

## Rebase / CI note

### DIFF
--- a/pkg/bgp/peers.go
+++ b/pkg/bgp/peers.go
@@ -25,6 +25,11 @@ const defaultBGPPort uint32 = 179
 
 // AddPeer will add peers to the BGP configuration
 func (b *Server) AddPeer(ctx context.Context, peer kubevip.BGPPeer) (err error) {
+	remotePort := defaultBGPPort
+	if peer.Port != 0 {
+		remotePort = uint32(peer.Port)
+	}
+
 	p := &api.Peer{
 		Conf: &api.PeerConf{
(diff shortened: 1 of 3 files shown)


p-strusiewiczsurmacki-mobica (commented):
LGTM

p-strusiewiczsurmacki-mobica (approved):

PR #1712, opened 2026-08-23
fix(manager): handle empty node list in annotations watcher

MaxRink:
## Bug
The annotations watcher indexed `nodeList.Items[0]` after listing nodes filtered by `kubernetes.io/hostname=<NodeName>`, with no length check. An empty list (misconfigured NodeName, or the hostname label not yet present) panics with index out of range at startup.

## Fix
Return a descriptive error when the filtered node list is empty instead of indexing. Adds `TestAnnotationsWatcherHandlesEmptyNodeList` (panics without the fix).

## Rebase / CI note

### DIFF
--- a/pkg/endpoints/endpoints_wireguard_test.go
+++ b/pkg/endpoints/endpoints_wireguard_test.go
@@ -6,7 +6,7 @@ import (
 	v1 "k8s.io/api/core/v1"
 )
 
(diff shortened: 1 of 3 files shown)


p-strusiewiczsurmacki-mobica (approved):
LGTM

PR #1713, opened 2026-08-23
fix(bgp): validate MP-BGP fixed-address family

MaxRink:
## Bug
The fixed MP-BGP addresses were validated with `net.ParseIP(...) == nil`, which accepts either family, so an IPv6 literal in `MpbgpIPv4` (or vice versa) passed despite the error strings promising family validation. The mis-family address was then used as the family-specific MP-BGP nexthop.

## Fix
Validate with the existing family-specific helpers (`utils.IsIPv4`/`utils.IsIPv6`). Adds `TestFindMpbgpAddressesRejectsFixedAddressFamilyMismatches`.

## Rebase / CI note

### DIFF
--- a/pkg/endpoints/endpoints_wireguard_test.go
+++ b/pkg/endpoints/endpoints_wireguard_test.go
@@ -6,7 +6,7 @@ import (
 	v1 "k8s.io/api/core/v1"
 )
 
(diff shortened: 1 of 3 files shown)


p-strusiewiczsurmacki-mobica (commented):
LGTM

p-strusiewiczsurmacki-mobica (approved):

PR #1714, opened 2026-08-23
fix(endpoints): only use Hostname for locality when NodeName is absent

MaxRink:
## Bug
Both endpoint providers fell through to a Hostname match even when `NodeName` was present but mismatched, so an endpoint with `NodeName=node-b, Hostname=node-a` was wrongly classified local to node-a. The code comments state Hostname is only useful when NodeName is unavailable; the implementation contradicted that, misfiring ExternalTrafficPolicy=Local leader election.

## Fix
Only evaluate the Hostname branch when `NodeName` is nil, in both the EndpointSlices and legacy Endpoints providers.

## Generalized verification
The narrow regression fixture is replaced by `pkg/endpoints/providers/providers_test.go`, which exercises both providers through the same conformance-style cases and includes a generic NodeName-over-Hostname precedence test.

## Verification
- `go test ./pkg/endpoints/providers`
- `go test -race ./pkg/endpoints/providers`

Both commands passed in the Debian Bookworm ARM64 OrbStack Linux VM.

## Rebase note

## Rebase / CI note


p-strusiewiczsurmacki-mobica (approved):
LGTM

PR #1715, opened 2026-08-23
fix(bgp): keep host tracker in sync with GoBGP to prevent silent route loss

MaxRink:
## Purpose

Keep the BGP host tracker synchronized with GoBGP across operation failures and daemon recovery.

## Key changes

- Always issues idempotent `AddPath` calls instead of trusting stale tracker presence.
- Updates tracker refcounts only after successful daemon operations.
- Allows advertisements to recover after failed add/delete operations or GoBGP restart.

## Validation

Adds retry and recovery regression tests. Package race tests and current CI checks pass.

## Dependency

### DIFF
--- a/pkg/bgp/hosts.go
+++ b/pkg/bgp/hosts.go
@@ -16,29 +16,32 @@ func (b *Server) AddHost(ctx context.Context, addr string, object string) error 
 
 	objects, exists := b.tracker[addr]
 
+	ip, _, err := net.ParseCIDR(addr)
+	if err != nil {
+		return err
+	}
+
+	p := b.getPath(ip)
+	if p == nil {
+		return fmt.Errorf("failed to get path for %v", ip)
+	}
+
+	// The tracker can outlive the GoBGP server or retain a route after a
+	// failed withdrawal. Re-adding the path keeps the daemon and tracker in
+	// sync; GoBGP treats an identical path as an idempotent operation.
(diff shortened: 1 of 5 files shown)


pkg/bgp/hosts.go
```diff
@@ -16,29 +16,32 @@ func (b *Server) AddHost(ctx context.Context, addr string, object string) error
...
+		return fmt.Errorf("failed to get path for %v", ip)
+	}
 
-		p := b.getPath(ip)
-		if p == nil {
-			return fmt.Errorf("failed to get path for %v", ip)
-		}
+	// The tracker can outlive the GoBGP server or retain a route after a
+	// failed withdrawal. Re-adding the path keeps the daemon and tracker in
+	// sync; GoBGP treats an identical path as an idempotent operation.
```
p-strusiewiczsurmacki-mobica: > GoBGP treats an identical path as an idempotent operation.

Can we get more info about this? Won't GoBGP send an update to the neighbors every time the route is added, even if it's the same route?

MaxRink: Tested with real GoBGP v4.8.0 peers. Ordinary sessions suppress an identical outbound UPDATE; negotiated Add-Path sessions do not. kube-vip does not enable Add-Path. Commit `22ca393` adds coverage for both cases and recovery.

PR #1716, opened 2026-08-23
fix(debouncer): prevent deadlock when forwarding to a terminating object

MaxRink:
## Bug
`debouncer.Start` forwarded events with an unguarded blocking send `eventObject.input <- tmp`. If the per-object worker self-terminated (sent its Deleted event and returned) but had not yet been removed from the map, the main loop could find the stale, receiver-less object and block forever, so `Start` never returned and the output channel stalled. Context cancellation could not rescue it.

## Fix
Guard the send with a select against the debouncer context, and remove the object from the map synchronously at termination so the main loop cannot observe a receiver-less object. Adds regression tests for both the cancellation and the non-cancellation orphan-window cases (run under `-timeout` and `-race`).

Found during the kube-vip test-coverage/bug-bash effort; adversarially confirmed and bidirectionally validated (repro tests pass with the fix, fail when reverted), full package green under -race. No unrelated changes.

## Rebase / CI note

### DIFF
--- a/pkg/debouncer/debouncer.go
+++ b/pkg/debouncer/debouncer.go
@@ -52,9 +52,8 @@ func (n *ns) add(name string, output chan<- watch.Event) *object {
 	return i
 }
 
-func (n *ns) del(name string) {
-	if _, exists := n.Load(name); exists {
-		n.Delete(name)
+func (n *ns) del(name string, object *object) {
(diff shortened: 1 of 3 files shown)


p-strusiewiczsurmacki-mobica (approved):
LGTM

PR #1717, opened 2026-08-23
fix(endpoints): aggregate endpoints across multiple EndpointSlices

MaxRink:
## Purpose

Keep complete endpoint state when a Service spans multiple EndpointSlices.

## Key changes

- Stores endpoints by slice name and aggregates them by address family.
- Removes only the deleted slice contribution and preserves dual-stack and named-port behavior.
- Adds a shared provider conformance suite for legacy Endpoints and EndpointSlices, including replacement and cleanup.

## Validation

`go test ./pkg/endpoints/providers ./pkg/endpoints ./pkg/services` and the corresponding `-race` run passed in Debian Bookworm ARM64 Linux. Current CI checks pass.

## Dependency

### DIFF
--- a/pkg/endpoints/providers/endpointslices.go
+++ b/pkg/endpoints/providers/endpointslices.go
@@ -88,12 +88,19 @@ func isServing(conditions discoveryv1.EndpointConditions) bool {
 
 func (ep *Endpointslices) GetAllEndpoints() ([]string, error) {
 	result := []string{}
+	seen := map[string]struct{}{}
 	for _, eps := range ep.slices {
 		for _, e := range eps.Endpoints {
 			if !isServing(e.Conditions) {
 				continue
 			}
-			result = append(result, e.Addresses...)
+			for _, address := range e.Addresses {
+				if _, ok := seen[address]; ok {
+					continue
+				}
+				seen[address] = struct{}{}
+				result = append(result, address)
+			}
 		}
 	}
 	return result, nil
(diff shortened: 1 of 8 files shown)


pkg/endpoints/providers/interface.go
```diff
@@ -24,6 +24,12 @@ type Provider interface {
 	ResolvePort(servicePort v1.ServicePort) int32
 }
 
+// ObjectDeleter is implemented by providers that retain state for individual
+// watched objects and need to remove that state on a watch.Deleted event.
+type ObjectDeleter interface {
```
Cellebyte: this can be part of the general interface.

pkg/endpoints/providers/endpointslices.go
```diff
@@ -153,11 +172,64 @@ func (ep *Endpointslices) GetLabel() string {
...
+		for _, p := range ep.allPorts() {
 			if p.Name != nil && *p.Name == name && p.Port != nil {
 				return *p.Port
 			}
 		}
 		return 0
 	})
 }
+
+func (ep *Endpointslices) ensureStorage() {
```
Cellebyte: And do the same here again.

I am not sure if this is a bug as an endpointslice is handled per UID

pkg/endpoints/providers/interface.go
MaxRink: Addressed. `DeleteObject(runtime.Object) error` is now a required `Provider` method, and the service reconciler plus WireGuard endpoint watcher call it directly. The shared provider conformance suite covers replacement, per-object deletion, and cleanup across legacy Endpoints and EndpointSlices. The affected normal and `-race` suites pass in the OrbStack Linux VM.

p-strusiewiczsurmacki-mobica:
I believe that something went wrong during iteration as this PR only contains the test code.

MaxRink:
The production EndpointSlice aggregation and deletion handling is included at the current head; the PR is no longer test-only.

PR #1718, opened 2026-08-23
fix(servicecontext): synchronize access to shared context fields

MaxRink:
## Bug
Three fields on `servicecontext.Context` were accessed from sibling goroutines (the election goroutine and the endpoint-watcher goroutine, both launched from `processor.AddOrModify`) with no synchronization, so `go test -race` (which CI runs) reported data races:
- `LeaderCancel` written in `leader.go`, read/invoked in the `endpoints_*.go` cleanup paths.
- `EndpointsReady` (channel) reassigned in `ResetReadiness` while received on in `leader.go`/`services.go`.
- `IsWatched` written and read in `processor.go`.

## Fix
Add a mutex to `servicecontext.Context` and route these accesses through accessors (`SetLeaderCancel`/`CallLeaderCancel`, `GetEndpointsReady`, `SetWatched`/`IsWatchedLocked`, plus locked `SignalReadiness`/`ResetReadiness`). `CallLeaderCancel` and `GetEndpointsReady` copy under the lock and act on the copy outside it, so the lock is never held across the cancel call or a channel receive. Call sites in `endpoints_*.go`, `leader.go`, `processor.go` and `services.go` updated to use the accessors.

## Verification
`pkg/servicecontext`, `pkg/services` and `pkg/endpoints` all pass `go test -race` with no DATA RACE and no deadlock. The confirmation stage reproduced the original races under `-race`; the added tests guard the accessor synchronization. gofmt/vet clean.

Found during the kube-vip test-coverage/bug-bash effort. No behavior change.

## Rebase note

## Rebase / CI note

### DIFF
--- a/pkg/endpoints/endpoints_bgp.go
+++ b/pkg/endpoints/endpoints_bgp.go
@@ -66,9 +66,7 @@ func (b *BGP) clear(svcCtx *servicecontext.Context, lastKnownGoodEndpoint *strin
 
 	b.clearEgress(lastKnownGoodEndpoint, service)
 
-	if svcCtx.LeaderCancel != nil {
-		svcCtx.LeaderCancel()
-	}
+	svcCtx.CallLeaderCancel()
 }
 
 func (b *BGP) getEndpoints(service *v1.Service, id string) ([]string, error) {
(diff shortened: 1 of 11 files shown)


p-strusiewiczsurmacki-mobica (approved):
LGTM

p-strusiewiczsurmacki-mobica (approved):

PR #1719, opened 2026-08-23
fix(kubevip): write LoseLeadership timeout under its dedicated env var

MaxRink:
## Purpose

Fix generated LoseLeadership configuration and eliminate a service endpoint teardown race.

## Key changes

- Writes the timeout under `vip_loseleadership_timeout_seconds` instead of duplicating the boolean `vip_loseleadership` variable.
- Serializes endpoint teardown with service processing so a stale endpoint event cannot race deletion and recreate state for a removed Service.
- Adds focused config-generation and service lifecycle race coverage.

## Validation

## Dependency

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Fixed
+- Generate the lose-leadership timeout under its dedicated `vip_leadershiplosstimeout` environment variable.
 - Propagate `bgp_attach_ip_to_interface` into per-service config so it attaches BGP-mode Service VIPs to the interface as configured.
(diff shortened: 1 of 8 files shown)


pkg/kubevip/config_generator_loseleadership_test.go
```diff
@@ -0,0 +1,25 @@
+package kubevip
+
+import "testing"
+
+func TestGeneratePodSpecLoseLeadershipTimeoutUsesDedicatedEnvironmentVariable(t *testing.T) {
```
p-strusiewiczsurmacki-mobica: Can we put this test in `config_generator_test.go` instead of creating a separate file for just this?

p-strusiewiczsurmacki-mobica (changes requested):

pkg/kubevip/config_generator_loseleadership_test.go
MaxRink: Addressed. The lose-leadership environment-variable test is now in `pkg/kubevip/config_generator_test.go`; the separate test file was removed.

PR #1720, opened 2026-08-23
fix(vip): apply per-call DAD skip instead of leaking it into persistent state

MaxRink:
## Bug
`AddIP` only ever OR-ed `IFA_F_NODAD` onto the long-lived `configurator.address.Flags` when a per-call DAD skip was requested, and never cleared it. A transient DADFAILED-recovery call (`AddIP(_, true)`) therefore permanently disabled DAD for that VIP; a later normal `AddIP(_, false)` kept the stale NODAD flag, silently skipping DAD against the operator's `dadSkip=false` configuration.

## Fix
Set or clear `IFA_F_NODAD` to reflect the current call rather than only OR-ing it. Adds `TestAddIPPerCallDADSkipDoesNotPersist`.

Found during the kube-vip test-coverage/bug-bash effort; adversarially confirmed and bidirectionally validated (repro test passes with the fix, fails when reverted). No unrelated changes.

## Rebase / CI note

### DIFF
--- a/pkg/endpoints/endpoints_wireguard_test.go
+++ b/pkg/endpoints/endpoints_wireguard_test.go
@@ -6,7 +6,7 @@ import (
 	v1 "k8s.io/api/core/v1"
 )
 
-func TestWireguardDeleteDoesNotDereferenceNilServiceContext(t *testing.T) {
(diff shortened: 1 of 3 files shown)


p-strusiewiczsurmacki-mobica (approved):
LGTM

PR #1721, opened 2026-08-23
fix(cluster): close BGP health-check bodies and withdraw CP route on cancel

MaxRink:
## Purpose

Prevent BGP health-check connection leaks and stale control-plane advertisements after leadership loss.

## Key changes

- Closes each health-check response body within its polling iteration.
- Withdraws statically announced control-plane CIDRs when the service context is canceled.
- Covers per-poll body closure and static-route withdrawal.

## Validation

Regression tests pass, together with unit, integration, lint, CodeQL, and E2E checks.

## Dependency

### DIFF
--- a/pkg/cluster/service_bodyleak_internal_test.go
+++ b/pkg/cluster/service_bodyleak_internal_test.go
@@ -1,0 +1,132 @@
+package cluster
+
+import (
+	"bytes"
+	"context"
+	"io"
+	"net"
+	"net/http"
+	"net/http/httptest"
+	"sync/atomic"
+	"testing"
+	"time"
+
+	"github.com/kube-vip/kube-vip/pkg/kubevip"
+)
+
+type testRoundTripper func(*http.Request) (*http.Response, error)
+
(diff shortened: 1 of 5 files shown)


pkg/cluster/service_bodyleak_internal_test.go
```diff
@@ -0,0 +1,102 @@
...
+	if closes.Load() != 1 {
+		t.Fatalf("health-check response body was not closed before the next poll: got %d closes", closes.Load())
+	}
+
+	cancel()
+	select {
+	case <-done:
+	case <-time.After(time.Second):
+		t.Fatal("health check loop did not stop after context cancellation")
+	}
```
p-strusiewiczsurmacki-mobica: I believe this is already executed by `t.Cleanup`.

p-strusiewiczsurmacki-mobica (changes requested):

pkg/cluster/service_bodyleak_internal_test.go
MaxRink: Removed the redundant completion result; cleanup remains handled by t.Cleanup.

PR #1722, opened 2026-08-23
fix(nftables): bounds-check and handle unknown expr types in ruleEqual

MaxRink:
## Purpose

Make nftables rule comparison safe and correct for differing or previously unhandled expressions.

## Key changes

- Rejects rules with different expression counts before indexing.
- Uses deep equality for expression types not covered by specialized comparisons.
- Prevents panics and incorrect equality in unique-rule and existence checks.

## Validation

Adds `TestRuleEqualRejectsRulesThatDiffer`. Unit, integration, lint, CodeQL, and E2E checks pass.

## Dependency

### DIFF
--- a/pkg/nftables/client.go
+++ b/pkg/nftables/client.go
@@ -320,6 +320,9 @@ func ruleEqual(a, b *nftables.Rule) bool {
 	if !bytes.Equal(a.UserData, b.UserData) {
 		return false
 	}
+	if len(a.Exprs) != len(b.Exprs) {
+		return false
+	}
 
 	for i := range a.Exprs {
 		switch a.Exprs[i].(type) {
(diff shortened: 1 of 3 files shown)


pkg/nftables/client.go
```diff
@@ -320,6 +320,9 @@ func ruleEqual(a, b *nftables.Rule) bool {
 	if !bytes.Equal(a.UserData, b.UserData) {
 		return false
 	}
+	if len(a.Exprs) != len(b.Exprs) {
```
p-strusiewiczsurmacki-mobica, line 323: I am not entirely sure this is correct. I think it is possible that rules are optimized on kernel level, and therefore the same rule in kernel might have different length of expressions than what we pass, but I am not a 100% sure about that.

MaxRink: Tested with a privileged Linux kernel round trip using nftables v0.3.0 and current main. kube-vip expression counts and order stayed unchanged. Only counter state and lookup SetID changed, and both are normalized. Unsupported decoder expressions can be omitted, where strict length inequality is the safe result.

PR #1723, opened 2026-08-23
fix(vip): harden DHCP client lifecycle and reply parsing

MaxRink:
## Bugs (pkg/vip/dhcpv4.go, dhcpv6.go)
1. **DHCPv4 backoff deadlock**: `requestWithBackoff` called `Stop()` from the pre-loop `Start` goroutine, which blocks forever on `releasedChan` (only closed by the select loop this goroutine hasn't reached). Triggered by the default 3-attempt backoff against an unreachable server.
2. **DHCPv6 Stop refcount/socket leak**: the shared client manager is registered under the parent interface name but `Stop` decremented by the child interface name, so the refcount never dropped and sockets leaked on every DHCPv6 stop.
3. **DHCPv6 getAddress panic**: a duplicated guard left `Addresses()[0]` unchecked, panicking on a server reply whose IANA carries no IAADDR (pool exhausted / malformed) - remote-triggerable.

## Fix
Signal backoff failure without self-deadlocking; key the manager decrement by the stored parent name; reject an IANA with no addresses. Adds regression tests for all three (the deadlock test runs under `-timeout`).

Note: the DHCPv4-backoff and getAddress tests are bidirectionally validated; the DHCPv6-stop test exercises the fixed manager-key path.

Found during the kube-vip test-coverage/bug-bash effort; adversarially confirmed and bidirectionally validated. No unrelated changes.

## Rebase / CI note

### DIFF
--- a/pkg/endpoints/endpoints_wireguard_test.go
+++ b/pkg/endpoints/endpoints_wireguard_test.go
@@ -6,7 +6,7 @@ import (
 	v1 "k8s.io/api/core/v1"
 )
 
-func TestWireguardDeleteDoesNotDereferenceNilServiceContext(t *testing.T) {
+func TestWireguardClearDoesNotDereferenceNilServiceContext(t *testing.T) {
 	worker := &wireguardWorker{}
 	service := &v1.Service{}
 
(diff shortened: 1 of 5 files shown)


p-strusiewiczsurmacki-mobica (approved):
LGTM

PR #1724, opened 2026-08-23
fix(kubevip): keep configured values when their env var is unset

MaxRink:
## Purpose

Preserve configured ARP and DHCP values when their environment variables are unset.

## Key changes

- Defaults `ArpBroadcastRate` only when no value is already configured.
- Derives `DHCPMode` from DNS mode only when DHCP mode is still unset.
- Restores the documented `flags > environment > file` behavior for this scope.

## Validation

Adds `TestParseEnvironmentPreservesConfiguredARPAndDHCPValues`. Unit, integration, lint, CodeQL, and E2E checks pass.

## Dependency

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Fixed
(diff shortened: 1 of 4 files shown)


p-strusiewiczsurmacki-mobica (approved):
LGTM

PR #1725, opened 2026-08-23
fix(etcd): handle closed observe channel and honor context in leader election

MaxRink:
## Purpose

Prevent etcd election crashes and delayed shutdown when observation or leadership is lost.

## Key changes

- Handles closure of the etcd Observe channel without dereferencing a nil response.
- Replaces the lease-TTL sleep with context-aware waiting so cancellation stops promptly.
- Preserves the `OnStoppedLeading` cleanup path and avoids starting leadership with a canceled context.

## Validation

Adds focused closed-channel and cancellation tests. Unit, integration, lint, CodeQL, and E2E checks pass.

## Dependency

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Fixed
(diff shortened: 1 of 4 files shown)


p-strusiewiczsurmacki-mobica (approved):
LGTM

PR #1727, opened 2026-08-23
fix(cluster): withdraw control-plane route on leadership loss in routing-table mode

MaxRink:
## Purpose

Make routing-table control-plane leadership install and withdraw routes through the intended election lifecycle.

## Key changes

- Dispatches routing-table control-plane startup to the actual leader-election path instead of bypassing election.
- Withdraws control-plane routes when leadership context is canceled.
- Adds coverage for election dispatch and route cleanup on service stop.

## Validation

## Dependency


p-strusiewiczsurmacki-mobica (approved):
LGTM

PR #1737, opened 2026-08-24
fix(instance): propagate BGPAttachIPToInterface to per-service VIP configs

MarijnRitzen:
The per-service configs built in pkg/instance drop BGPAttachIPToInterface, so shouldAddServiceIP() never saw the flag for service VIPs and the IP was  never attached to the interface. Copy the field from the global config.

 Follow-up to #1644.


MarijnRitzen:
Fixed in https://github.com/kube-vip/kube-vip/pull/1744

PR #1739, opened 2026-08-25
test(endpoints): follow WireGuard clear refactor

MaxRink:
## Summary

- update the WireGuard endpoint unit test to exercise the current `clear` API after the endpoint worker refactor
- keep the nil service-context regression coverage without relying on the removed `delete` helper

## Validation

- `go test -race ./pkg/endpoints ./pkg/endpoints/providers ./pkg/services` in an OrbStack Linux VM


p-strusiewiczsurmacki-mobica:
LGTM

PR #1743, opened 2026-08-26
Crash when default interface changes

Xartos:
According to [this PR](https://github.com/kube-vip/kube-vip/pull/339) the purpose of the monitoring of the default interface was to hard crash if it changed to force the pod to restart.
This behavior seems to have changed during [this change](https://github.com/kube-vip/kube-vip/pull/1043) and now it just seems to print the error and stop monitoring the interface.

This PR aims to revert to this behavior.

### DIFF
--- a/cmd/kube-vip.go
+++ b/cmd/kube-vip.go
@@ -450,7 +450,7 @@ var kubeVipManager = &cobra.Command{
 				wg.Go(func() {
 					if err := vip.MonitorDefaultInterface(ctx, defaultIF); err != nil {
 						log.Error("interface monitor", "err", err)
-						os.Exit(1)
+						cancel()
 					}
 				})
 			}
--- a/cmd/kube-vip.go
+++ b/cmd/kube-vip.go
@@ -449,9 +449,8 @@ var kubeVipManager = &cobra.Command{
 
 				wg.Go(func() {
 					if err := vip.MonitorDefaultInterface(ctx, defaultIF); err != nil {
-
 						log.Error("interface monitor", "err", err)
-						return
+						os.Exit(1)
 					}
 				})
 			}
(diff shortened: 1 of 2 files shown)


cmd/kube-vip.go
```diff
@@ -449,9 +449,8 @@ var kubeVipManager = &cobra.Command{
 
 				wg.Go(func() {
 					if err := vip.MonitorDefaultInterface(ctx, defaultIF); err != nil {
-
 						log.Error("interface monitor", "err", err)
-						return
+						os.Exit(1)
```
p-strusiewiczsurmacki-mobica: I agree with the idea, but, as the manager might be already started, I think it would be better to cancel the context so we know things that should be cleaned up, will be cleaned up.

p-strusiewiczsurmacki-mobica (changes requested):

cmd/kube-vip.go
Xartos: That is very true, not super used with go so I completely missed that.

But this should be enough if I'm not mistaken, right? https://github.com/kube-vip/kube-vip/pull/1743/commits/0c0083fabda4ce68605cc3f7335a8c4fe486636c

Xartos: Did some more testing and found that this was needed as well https://github.com/kube-vip/kube-vip/pull/1743/commits/d496442a9ea034d3aabb468741862ee95e98a9b6

p-strusiewiczsurmacki-mobica (approved):
LGTM
I think this can be merged if the tests will pass.

Xartos:
@thebsdbox Just checking, is it possible to get this approved for the tests? Would be nice for us to get this merged :innocent:

PR #1744, opened 2026-08-29
Fix bgp_attach_ip_to_interface not applied to Service VIPs

jrcichra:
`bgp_attach_ip_to_interface` is meant to bind BGP-mode Service VIPs to the interface, but it currently has no effect. `shouldAddServiceIP()` checks the flag on a per-service config struct, and `NewInstance()` builds that struct fresh for each Service, copying over most fields from the global config, missing this one. As a result the flag is ignored regardless of its value, and Service VIPs are never bound.

Copies the field at both construction sites where it's built, and adds a regression test covering the propagation.

Verified on a live cluster: before the fix, no nodes logged a successful bind; after, all seven did consistently.


Cellebyte:
This should not be needed.
As the Service VIP should not be bound to an interface.
The IP is configured in kube-proxy.


Any reason/use-case we would need to attach it also to an interface?

jrcichra:
This isn't about changing the default, it's just that bgp_attach_ip_to_interface (#1644) already exists and is documented, but it's a no-op because it's not wired up on v1.2.3.

I've had some strange behavior with my k3s + EdgeRouter X setup without the VIP present on lo, and having this actually wired up lets me experiment if it helps.

thebsdbox:
Attaching the vip to lo, was the original behaviour.. I wasn't aware it had been removed. I understand that kube-proxy can handle this with iptables rules but other implementations such as cilium or others may require a physical address to be there. Perhaps we could gate this with a flag if needed?

Cellebyte:
@thebsdbox I think this is already gated on the service using the annotation if I am not misunderstanding.

MarijnRitzen:
Nice, I can close this PR then https://github.com/kube-vip/kube-vip/pull/1737

MarijnRitzen:
@Cellebyte we need this change before we can upgrade our kubevip, could you please cut a new release?🙏

PR #1738, opened 2026-08-25
CORE Locks and Concurrency issues

Cellebyte:
* Let's try to fix some concurrency issues.

PR #1745, opened 2026-08-30
fix(services): isolate shared lease member lifecycle

MaxRink:
## Summary

Follow-up to kube-vip/kube-vip#1738, addressing the production-reachable findings from the repeated concurrency review.

- Move shared Service election ownership from the initiating Service context to a per-lease coordinator.
- Track exact Service/readiness generations so deleted or superseded candidates cannot activate after delayed acquisition.
- Keep endpoint loss and restoration member-local while preserving a healthy sibling's shared campaign.
- Drain member workers before voluntarily releasing the final shared campaign.
- Serialize shared network resources through construction, replacement, and generation cleanup.
- Make cluster worker generations stoppable and joinable, including partial-startup rollback.
- Preserve pre-existing/shared VIPs and roll back only address ownership acquired by the current startup.
- Keep failed route installation out of the ownership tracker so a later retry reaches the kernel.
- Replace the private callback registry with an explicit internal role and make optional callbacks nil-safe.
- Reconcile WireGuard DNAT per address family, include SCTP, distinguish TCP/UDP/SCTP IDs, bound long IDs with a hash suffix, and remove legacy IDs during migration.
- Handle closed interface subscriptions safely while preserving interface generation ownership.
- Track and coalesce SIGUSR1 configuration dumps through shutdown.

The complete lifecycle design is documented in
[`pkg/services/architecture.md`](https://github.com/kube-vip/kube-vip/blob/cfba582984df177ac02c0692f649b8426dcc29e1/pkg/services/architecture.md), with the WireGuard document linking to it from
[`pkg/wireguard/architecture.md`](https://github.com/kube-vip/kube-vip/blob/cfba582984df177ac02c0692f649b8426dcc29e1/pkg/wireguard/architecture.md).

The architecture document covers component ownership, desired-state versions and tombstones, Service/readiness generations, reconciliation and replacement, endpoint processing, election coordination and states, cleanup/retry semantics, lock ordering, worker generations, shared-VIP ownership and rollback boundaries, route claims, WireGuard tunnel/DNAT family and protocol behavior, legacy-ID migration, direct and election-backed deletion, shutdown, observability, and behavioral test contracts.

## Production scenarios covered

- A Service that initiated a shared campaign is deleted before the node acquires the Kubernetes/etcd lease; only live siblings activate afterward.
- Candidate or follower endpoints disappear and later return; only that member is removed/restored, without moving the shared campaign.
- All members become ineligible or leadership is lost; the old generation drains before another local campaign starts.
- Shared and mixed VIPs stop concurrently, including explicit removal racing leadership loss and `PreserveVIPOnLeadershipLoss`.
- Rapid endpoint recovery does not race stale ARP/VIP cleanup.
- VLAN, DHCP, hostname-backed, and recreated Service resources cannot race construction and deletion.
- A WireGuard EndpointSlice refresh does not cancel Service leadership or delete a healthy same-family tunnel rule.
- TCP, UDP, and SCTP Services sharing a port receive distinct WireGuard rule IDs, while legacy rules are removed.
- Failed route installation can be retried rather than being accepted from a stale tracker entry.
- A partially successful address add is rolled back without deleting a pre-existing shared VIP.
- Netlink subscription closure cannot panic or spin the monitor.

## Validation

- Full Linux/arm64 `go test -race -shuffle=on ./...` and focused repeated race/stress runs were completed for the preceding lifecycle patch.
- On final head `cfba582984df177ac02c0692f649b8426dcc29e1`, native tests passed for `./pkg/route ./pkg/servicecontext`.
- Linux/arm64 test binaries for `./pkg/services ./pkg/endpoints ./pkg/cluster ./pkg/wireguard` compile and vet successfully; execution is covered by CI.
- Delta `golangci-lint`: 0 issues.
- `git diff --check`: passed.
- GitHub unit, integration, lint, image scan, ARP/BGP/routing-table/service E2E, DCO, dependency, and security checks: passed on final head `cfba582984df177ac02c0692f649b8426dcc29e1`.
- External License Compliance reports the same opaque `5 issues found` status on the target commit and the final head, with no result URL; this PR changes neither `go.mod` nor `go.sum`.

The branch is based directly on `kube-vip:folowup/locks` at `866058595eec5dd64279d972c7031330a404e2fb` and is intended to merge into that branch before kube-vip/kube-vip#1738.


Cellebyte:
I am thinking of some simplifications.
The mentioned ideas here are valid and could be checked.
The issue with this code addition is that it adds much more hoops to jump through we already have a lot o locks and channels.
I will use the code as input for my plan I will prepare.
#1738 is also not what we need....

MaxRink:
@thebsdbox Could you please request a Copilot code review for this PR? GitHub requires upstream write access to add reviewers, so the fork author cannot trigger `@copilot` through the CLI or review-request API. I will address any actionable findings it posts.

Cellebyte:
let's close this for now

PR #1750, opened 2026-09-04
break: remove preserveVipOnLeadershipLoss

Cellebyte:
* Remove this CLI flag as it introduces a lot of risk on the cleanup routines in kube-vip


Cellebyte:
successor is #1772

PR #1767, opened 2026-09-08
feat(lease): record VIP ownership in lease annotations

Cellebyte:
* First PR to ease Reviews


Cellebyte:
this must ensure the order of the VIPs stays always the same and only changes on updates.

pkg/election/lease_lock.go
```diff
@@ -0,0 +1,90 @@
...
+
+func newAnnotatedLeaseLock(lock resourcelock.Interface, leases coordinationv1client.LeaseInterface,
+	name string, annotations map[string]string) resourcelock.Interface {
+	return &annotatedLeaseLock{Interface: lock, leases: leases, name: name, annotations: annotations}
+}
+
+func (lock *annotatedLeaseLock) Get(ctx context.Context) (*resourcelock.LeaderElectionRecord, []byte, error) {
+	return lock.Interface.Get(ctx)
+}
+
```
p-strusiewiczsurmacki-mobica: Redundant function - `resourcelock.Interface` is embedded and it's `Get()` can be used without this.

pkg/election/lease_lock.go
```diff
@@ -0,0 +1,90 @@
...
+		return nil
+	}
+	changed, err := lock.ensureAnnotations(ctx)
+	if err != nil {
+		return err
+	}
+	if changed {
+		_, _, err = lock.Interface.Get(ctx)
+	}
+	return err
```
p-strusiewiczsurmacki-mobica: Nit: this is the same as the code in `Update()`, could be moved to separate function.

pkg/kubevip/lease_annotations.go
```diff
@@ -0,0 +1,102 @@
...
+	IFAProto     int        `json:"ifa_proto"`
+	VIPs         []LeaseVIP `json:"vips"`
+}
+
+type LeaseVIP struct {
+	Index int    `json:"index"`
+	Value string `json:"value"`
+}
+
+func WithLeaseVIPs(annotations map[string]string, instanceName string, ifaProto int, vips []string) (map[string]string, error) {
```
p-strusiewiczsurmacki-mobica, line 34: In this PR `vips []string` is passed as an empty slice, so it's hard to tell, but `parseLeaseVIP` can only support IP addresses, and therefor if any FQDN will be passed here, `normalizeLeaseVIPs(vip)`  will discard it. I am flagging this for when I'll check other PRs.

pkg/election/lease_lock.go
```diff
@@ -0,0 +1,90 @@
...
+	if record.HolderIdentity != lock.Identity() {
+		return nil
+	}
+	changed, err := lock.ensureAnnotations(ctx)
+	if err != nil {
+		return err
+	}
+	if changed {
+		_, _, err = lock.Interface.Get(ctx)
+	}
```
p-strusiewiczsurmacki-mobica, line 61: I am not sure if wrong annotations or getting the lock after successful update should return error, as this will stop the election.

pkg/kubevip/lease_annotations.go
Cellebyte: I think going back to string and using the alphabetical sort should also work here.

Cellebyte:
@p-strusiewiczsurmacki-mobica you can take another look, I addressed your feedback :)

p-strusiewiczsurmacki-mobica (approved):
LGTM

PR #1768, opened 2026-09-08
fix(vip): synchronize shared datapath state

Cellebyte:
Serialize the interface link cache, route tracker and address configuration so concurrent Services cannot corrupt shared state or deadlock on nested address locks.


p-strusiewiczsurmacki-mobica (approved):
I see nothing wrong with this one.

PR #1769, opened 2026-09-08
fix: wireguard + routing-protocol sync

Cellebyte:


pkg/wireguard/service_id.go
```diff
@@ -0,0 +1,39 @@
...
+)
+
+const maxServicePortIDLength = 50
+
+// ServicePortIDs returns the protocol-qualified nftables identifier and the
+// prior port-only identifier that must be removed during migration.
+//
+// Sanitisation maps '-' onto the '_' separator, so "a-b/c" and "a/b-c" would
+// otherwise share a chain; a hash of the raw name keeps them distinct.
+func ServicePortIDs(namespace, name string, port v1.ServicePort) (string, string) {
```
p-strusiewiczsurmacki-mobica, line 19: This looks like a good change, but seems to be unused. If this is to be used later, then that's fine.

pkg/instance/instance.go
```diff
@@ -531,35 +589,105 @@ func (i *Instance) addVLAN(parentInterface string, tag int) error {
...
+	instance.VLANInterface = interfaceName
+	instance.IsVLAN = true
+
+	return nil
+}
+
+// CleanupLinkAttachments stops this instance's DHCP clients and removes only
+// VLAN or macvlan links created by this instance that are not used by a
+// remaining Service instance.
+func (instance *Instance) CleanupLinkAttachments(remaining ...*Instance) error {
```
p-strusiewiczsurmacki-mobica, line 601: I like the owner-passing thing, but if we add it here, shouldn't we also use it in other places that remove vlan/dhcp (macvlan) interfaces (at least in: `pkg/services/services.go`, lines 491, 511)?

pkg/wireguard/service_id.go
Cellebyte: should be addressed :)

pkg/instance/instance.go
Cellebyte: this one as well :)

p-strusiewiczsurmacki-mobica (approved):
LGTM

PR #1784, opened 2026-09-15
Refactored and fixed metrics code

p-strusiewiczsurmacki-mobica:
This PR refactors metrics endpoint related code a bit, as I think it would be better to put int in separate package than to keep it in `cmd/kube-vip.go`.

### Changes

- Metrics endpoint creation code moved from `cmd/kube-vip.go` to `pkg/metrics/server.go`
- Added a `sync.Once` sefeguard for `RegisterPrometheusMetrics`, so metrics won't be ever registered twice (unlikely, but that would cause a panic).
- Added unit tests for metrics server.
- Removed unused `countServiceWatchEvent` from the `Manager`.


Cellebyte (approved):
LGTM

PR #1785, opened 2026-09-15
Added support for pprof endpoint 

p-strusiewiczsurmacki-mobica:
This PR extends #1784 with support for serving profiling data with [pprof](https://pkg.go.dev/net/http/pprof).

### Main changes:

- added separate HTTP server with pprof endpoints
- disabled by default, can be enabled with `--enablePprof` flag or `enable_pprof` env
- by default listens on `127.0.0.1:6060`. This can be configured with `--pprofServer` or `pprof_server`.
- Healthcheck server in `pkg/manger/manager.go` updated to use a separate HTTP mux, as `pprof` uses the default mux (and this cannot be changed), and therefore, `pprof` endpoints would be also announced by healthcheck server if it would use the default one.
- added some tests for those


Cellebyte:
I think this one needs rebase?

